PR #117, opened 2024-07-07
feat: Initial support for non-linux

lu-zero:
The fallback is thread unsafe, but it is what other shells are using.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,9 +23,19 @@ permissions:
   contents: read
 
 jobs:
-  build_x86_64:
-    name: "Build (x86_64/Linux)"
-    runs-on: ubuntu-latest
+  build:
+    strategy:
+      matrix:
+        include:
+          - host: "ubuntu-latest"
+            os: "linux"
+            arch: "x86_64"
+          - host: "macos-latest"
+            os: "macos"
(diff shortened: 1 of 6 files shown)


reubeno:
Thanks so much for contributing! I’ll look at the change more carefully this week. 

Out of curiosity, are there specific non-Linux OSes that you’re interested in? (Once the project is building on a given platform, I’d love to see us add some basic PR/CI checks to keep it building.)

lu-zero:
I tried it on macos and since it wasn't compiling I looked into it a bit.

reubeno:
It looks like the checks are all passing, thanks for the adjustments! I'll look into sorting out the remaining policy checks.

reubeno:
Okay, one of two policy issues resolved. For some reason the PR report workflow is silently failing; I’ll look at it next. Thanks for your patience as we work through these growing pains 🙂

reubeno (approved):

reubeno:
@lu-zero -- I've updated the PR report workflow and tested it on both a change of mine and a dependabot PR.

Would you mind syncing your PR branch with latest changes in main? That should be enough to re-trigger the right workflows. One way or the other, we'll get the changes merged. 

Thanks!

lu-zero:
Sure, no problem :)

PR #147, opened 2024-08-04
fix: Posix compliant argument parsing for `-c` mode

39555:
## Before

Brush misses `--` because of clap's approach to argument parsing. See: [clap-rs/clap/clap_builder/src/parser/parser.rs](https://github.com/clap-rs/clap/blob/edcaf8fcae5b747cee7e70f196d70a49ffcbdb84/clap_builder/src/parser/parser.rs#L103) `if arg_os.is_escape() { ...; continue } // means arg_os == '--'` and `--` is skipped.

But Posix can handle `--` as an operand or a command_name just fine.

```sh
cargo run -- -c 'echo $0: $1 $2 \"$@\"' --  -1 -2
#> CommandLineArgs.script_path = Some('-1')
#> CommandLineArgs.script_args = ['-2']
#> -1: -2 "-2"

bash -c 'echo $0: $1 $2 \"$@\"' --  -1 -2
#> --: -1 -2 "-1 -2"
```
```sh
cargo run -- -c 'echo $0: $1 $2 \"$@\"'  -1 -2
#> error: unexpected argument '-1' found

bash -c 'echo $0: $1 $2 \"$@\"'  -1 -2
#> -1: -2 "-2"
```

```sh
# cat script.sh
#> echo "$0: $1 $2 \"$@\""

cargo run -- script.sh --  -1 -2
#> CommandLineArgs.script_path = Some('script.sh')
#> CommandLineArgs.script_args = ['-1', '-2']
#> script.sh: -1 -2 "-1 -2"

bash script.sh -- -1 -2
#> script.sh: -- -1 "-- -1 -2"
```

```sh
cargo run -- script.sh -1 -2
#> error: unexpected argument '-1' found

bash script.sh  -1 -2
#> script.sh: -1 -2 "-1 -2"
```

## After This PR

```sh
cargo run -- -c 'echo $0: $1 $2 \"$@\"' --  -1 -2
#> --: -1 -2 "-1 -2"

bash -c 'echo $0: $1 $2 \"$@\"' --  -1 -2
#> --: -1 -2 "-1 -2"
```
```sh
cargo run -- -c 'echo $0: $1 $2 \"$@\"'  -1 -2
#> -1: -2 "-2"

bash -c 'echo $0: $1 $2 \"$@\"'  -1 -2
#> -1: -2 "-2"
```

```sh
cargo run -- script.sh --  -1 -2
#> script.sh: -- -1 "-- -1 -2"

bash script.sh -- -1 -2
#> script.sh: -- -1 "-- -1 -2"
```

```sh
cargo run -- script.sh -1 -2
#> script.sh: -1 -2 "-1 -2"

bash script.sh  -1 -2
#> script.sh: -1 -2 "-1 -2"
```

### DIFF
--- a/brush-core/src/builtins.rs
+++ b/brush-core/src/builtins.rs
@@ -382,3 +382,75 @@ fn brush_help_styles() -> clap::builder::Styles {
         .literal(styling::AnsiColor::Magenta.on_default() | styling::Effects::BOLD)
         .placeholder(styling::AnsiColor::Cyan.on_default())
 }
+
+/// This function and the [`try_parse_known`] exists to deal with
+/// the Clap's limitation of treating `--` like a regular value
+/// `https://github.com/clap-rs/clap/issues/5055`
+///
+/// # Arguments
+///
+/// * `args` - An Iterator from [`std::env::args`]
+///
+/// # Returns
+///
+/// * a parsed struct T from [`clap::Parser::parse_from`]
+/// * the remain iterator `args` with `--` and the rest arguments if they present
+///   othervise None
+///
+/// # Examples
+/// ```
+///    use clap::{builder::styling, Parser};
+///    #[derive(Parser)]
+///    struct CommandLineArgs {
+///       #[clap(allow_hyphen_values = true, num_args=1..)]
+///       script_args: Vec<String>,
+///    }
+///
+///    let (mut parsed_args, raw_args) =
+///        brush_core::builtins::parse_known::<CommandLineArgs, _>(std::env::args());
+///    if raw_args.is_some() {
+///        parsed_args.script_args = raw_args.unwrap().collect();
+///    }
+/// ```
+pub fn parse_known<T: Parser, S>(
+    args: impl IntoIterator<Item = S>,
+) -> (T, Option<impl Iterator<Item = S>>)
+where
+    S: Into<std::ffi::OsString> + Clone + PartialEq<&'static str>,
+{
+    let mut args = args.into_iter();
+    // the best way to save `--` is to get it out with a side effect while `clap` iterates over the args
+    // this way we can be 100% sure that we have '--' and the remaining args
+    // and we will iterate only once
+    let mut hyphen = None;
+    let args_before_hyphen = args.by_ref().take_while(|a| {
+        let is_hyphen = *a == "--";
+        if is_hyphen {
+            hyphen = Some(a.clone());
+        }
+        !is_hyphen
+    });
+    let parsed_args = T::parse_from(args_before_hyphen);
+    let raw_args = hyphen.map(|hyphen| std::iter::once(hyphen).chain(args));
+    (parsed_args, raw_args)
+}
+
+/// Similar to [`parse_known`] but with [`clap::Parser::try_parse_from`]
+/// This function is used to parse arguments in builtins such as [`crate::builtins::echo::EchoCommand`]
+pub fn try_parse_known<T: Parser>(
+    args: impl IntoIterator<Item = String>,
+) -> Result<(T, Option<impl Iterator<Item = String>>), clap::Error> {
+    let mut args = args.into_iter();
+    let mut hyphen = None;
+    let args_before_hyphen = args.by_ref().take_while(|a| {
+        let is_hyphen = a == "--";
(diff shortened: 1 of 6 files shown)


39555:
I haven't write integration tests yet. It Seems test cases can't compare its output to the bash equivalent yet?

reubeno:
Thanks for contributing! I'd like to find a few minutes to read through your changes to make sure I understand them, but it looks like a good find.

There's a (very) minimal write-up on the integration tests under docs/reference but the full schema of the .yaml case files still needs to be properly documented. As you called out, the test cases are run under bash and brush with exit code, stderr, and stdout compared. There's also some options for specifying command-line arguments to the shell, ignoring stderr differences, etc. The existing test cases are the best examples.

If the test case schema isn't sufficiently expressive to handle a new case, we can always change the test harness to support more options.

39555:
I will write test cases today soon

39555:
Tests for argument handling are ready. I've created a new file `arguments.yaml` for various tests related to argument handling.

While testing I found another bug/difference with bash. I think it is also related to clap
```sh
# bash
echo -- 1
#> -- -1

# brush
echo -- -1
#> -1
```

reubeno:
@39555 thanks for adding test coverage! For each new bug we find, I'd very much like us to follow that pattern to ensure the fix doesn't regress.

As for the other case you mention (`echo`) I'd agree with your assessment. This seems like another instance of what you had to fix with `brush`'s command line, due to clap's special handling of `--`.

Given that we're now seeing this in multiple places, do you think there's a more general fix or at least some common helper function/code we should extract to solve this problem across built-ins and brush's command line?

39555:
I found and open issue in clap https://github.com/clap-rs/clap/issues/5055. Maybe we could resolve it. `uutils`'s `echo` has the same problem https://github.com/uutils/coreutils/blob/main/src/uu/echo/src/echo.rs

reubeno:
Good to know that `uutils` has the same challenge. I like the idea of working to get this supported upstream in clap, thanks for posting there.

In the meantime, I'm supportive of moving forward with a fix for this issue and filing a separate issue on `echo` and the other built-ins affected (for us to resolve separately).

Regarding the actual code change here, would you please factor out the parsing logic that you've updated into a separate helper function? I think that will at least help to contain the workaround. Thanks!

39555:
@reubeno To which module can I place such a function? `brush_core::builtins`?

reubeno:
@39555 Sorry, I wasn't very clear. For now, I was just thinking of extracting a helper function in the same `.rs` file to keep `main()` easier to read. We can always move it somewhere more reusable in a future PR if we'd like to start using it in `echo` and other built-ins.

39555:
I meant that I've already tried it in `echo` and it works perfectly! In the last commit:
- I created two helper functions (for now in `brush_core::builtins`)
- I overrode `Parser::parse_from` in `CommandLineArgs`with the custom `parse_from` and moved the temporary solution out of main, so we can safely remove it when the issue is resolved and do not affect the code base. 
- Similarly in `echo`, I overrode `Command::new` so we can just remove it in the future
- Added another tests and some tests for echo

What do you think?

reubeno:
Ah! Thanks for clarifying and continuing to work on it. I'll look more closely at your latest rev.

reubeno (approved):
Thanks for making these changes! (And apologies for taking so long to get back to you.) Hopefully we'll be able to remove the workaround in the future if clap-rs adds support for this case...

reubeno:
Hi @39555 -- are you still willing and interested to help get these changes in?

39555:
Hi! Sorry! I've rebased onto the main and updated my branch

reubeno:
Looks like the `clippy` check error will need to get fixed, but the test failure seems to be related to my most recent changes in `main`. I will look at disabling the offending test until I can sort out what's going on--thanks for your patience on this.

39555:
All done

reubeno:
Thanks for the quick turnaround! Looks like all checks are passing fine. If you wouldn't mind doing a final merge/rebase, then I'll complete the PR shortly.

I plan to create a new release sometime this week and would like to include this fix in it. Thanks again for your contribution 🙂

PR #185, opened 2024-10-02
chore: adopt #[expect()] wherever possible

reubeno:
`#[expect]` will complain if the indicated warning doesn't apply. This allows us to quickly notice as soon as an allowed warning no longer needs to be allowed.


reubeno:
Turns out this requires a newer MSRV and we're not quite ready for it, so we'll close the PR for now and rehydrate it when we're ready to move the MSRV far enough forward (or find some other way to support the new attribute in a backwards compatible way.

PR #197, opened 2024-10-10
chore: where possible replace `async-trait` with native async trait support in 1.75+

39555:
As of Rust 1.75, the `async_fn_in_trait` feature has been merged (see: https://github.com/rust-lang/rust/pull/115822), meaning the async-trait crate is now only required for dynamic dispatch (dyn Trait). Since brush doesn’t use `dyn Command`, I decided to try removing the dependency. ~~The modules `arithmetic.rs` and `interp.rs` are most affected, as they use async recursion, which requires futures to be boxed. I’ve wrapped all recursive functions in `async move { .. }.boxed()`.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,7 +252,6 @@ version = "0.2.10"
 dependencies = [
  "anyhow",
  "async-recursion",
- "async-trait",
  "brush-parser",
  "cached",
  "cfg-if",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,6 +252,7 @@ version = "0.2.10"
 dependencies = [
  "anyhow",
  "async-recursion",
+ "async-trait",
  "brush-parser",
  "cached",
  "cfg-if",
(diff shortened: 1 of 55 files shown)


39555:
That's what `async_trait` or `async_recursion` do, but `async_recursion` does not work with an empty trait function body. I will revert the recursive functions since `async_trait` is really nice, and I forgot `Pin<..>` anyway. At least other traits without recursion would benefit from the native async though.

reubeno:
That sounds like a better direction. Let's remove unnecessary `#[async_trait::async_trait]` attributes where it should now "just work natively". For other places where we'd have to make the code less readable, let's hold off for now.

39555:
@reubeno done

reubeno:
LGTM.

Thanks also for addressing some warnings along the way. I'll get this merged, but if you find additional warnings like these that _aren't_ being caught by the GitHub PR workflows, then it would be interesting to figure out how we could tweak things to catch them in the future.

PR #198, opened 2024-10-11
refactor: replace `lazy_static` with `std::sync::LazyLock`

39555:
The functionality of `lazy_static` now in the core since 1.80 https://github.com/rust-lang/rust/pull/121377


reubeno:
Nice!

Our current MSRV is 1.75, which was initially released about 10 months ago, and this would force us to move that forward. What would you see as the benefits of doing so now vs. later?

More broadly, we need to establish a general procedure for when we upgrade the MSRV. I'll need to do some research to see what other projects do, but am very open to input here.

39555:
We can do it later. There are no huge benefits

Elsie19:
> More broadly, we need to establish a general procedure for when we upgrade the MSRV. I'll need to do some research to see what other projects do, but am very open to input here.

I found this: https://doc.rust-lang.org/cargo/reference/rust-version.html.

reubeno:
I've cherry-picked this into #553, along with a full upgrade to MSRV=1.85.0!

Closing this PR now that it's done its job. Thanks for your contributions toward this, @39555 :smile:.

PR #200, opened 2024-10-11
feat: implement `command -p`

39555:
Implements the missing `command -p` flag to use a default value of $PATH that is guaranteed to find all of the standard utilities.

- Uses `nix::libc::confstr` with `_CS_PATH` (see [confstr.3](https://man7.org/linux/man-pages/man3/confstr.3.html)) similar to what Bash do.
- If no value is returned falls back to the `STANDARD_UTILS_PATH`, which is taken from Bash.

Doesn't make sense on Windows though.

### DIFF
--- a/brush-core/src/builtins/command.rs
+++ b/brush-core/src/builtins/command.rs
@@ -3,6 +3,11 @@ use std::{fmt::Display, io::Write, path::Path};
 
 use crate::{builtins, commands, error, shell, sys::fs::PathExt, ExecutionResult};
 
+/// The value for PATH when invoking `command -p`. This is only used when
+/// the Posix.2 `confstr()` returns nothing
+/// The value of this variable is taken from the BASH source code.
+const STANDARD_UTILS_PATH: &[&str] = &["/bin", "/usr/bin", "/sbin", "/usr/sbin", "/etc:/usr/etc"];
+
 /// Directly invokes an external command, without going through typical search order.
 #[derive(Parser)]
 pub(crate) struct CommandCommand {
@@ -32,10 +37,6 @@ impl builtins::Command for CommandCommand {
         &self,
         context: commands::ExecutionContext<'_>,
     ) -> Result<builtins::ExitCode, error::Error> {
-        if self.use_default_path {
-            return error::unimp("command -p");
-        }
-
         if self.print_description || self.print_verbose_description {
             if let Some(found_cmd) = self.try_find_command(context.shell) {
                 if self.print_description {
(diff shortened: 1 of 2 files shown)


reubeno:
Thanks for working on this. A few higher-level comments...

We're not quite there today with the existing code, but I'm trying to keep platform-switched logic under `brush-core/src/sys`, as a sort of adaptation layer. There's already existing modules there around process management, signals, terminal configuration, etc. It seems reasonable to add a helper under `fs` there or under a new `paths` module that exposes the "default" search paths for executables on the platform.

For Windows, there's a set of paths reasonable to hard-code as fallbacks (e.g., `%SystemRoot%\system32`, `%SystemRoot%`, etc.) but I think it would be okay leaving that stubbed out for now and sorted out in a separate set of changes.

For Unix-like platforms, I didn't know about `confstr()`. Do you know if, in practice, it ever returns anything other than the usual dir paths (e.g., `/bin`, `/usr/bin`)? If it doesn't, then I'd prefer that we avoid the unsafe code calling into it and keep the fallback set of paths you added anyway. (In general, we're trying to limit the unsafe code in this crate; we acknowledge that `nix` and `libc` have a fair bit of it, but ideally trying to leverage those crates and stay at a higher level.)

Once we sort out the above, would also be great if you could add a compat test to cover `command -p`. Thanks!

39555:
@reubeno https://github.com/NixOS/nixpkgs/pull/586 nixos returns `/run/current-system/sw/bin:/bin:/usr/bin`
Checked with shell command `getconf CS_PATH`

reubeno:
@39555 -- that's helpful context, thank you for digging that up. I'd love to see `confstr` get properly wrapped in a higher-level Unix crate like [`nix`](https://crates.io/crates/nix), which we typically utilize to provide a higher-level API on top of `libc`-exported bindings. (It would be interesting to know how open they'd be to taking the wrapper you've added here, or some version of it.)

With that said, I'm supportive of keeping the unsafe code in `brush` for now (in a reasonable place under `sys`, like described above), but will want to take more time to carefully review the logic and the docs for `confstr()`.

39555:
It is not the most urgent feature, so I think we should wait for https://github.com/rust-lang/libc/pull/3612 which makes `confstr` available on all unix systems. Than I will make PR to the nix library.

reubeno:
> It is not the most urgent feature, so I think we should wait for [rust-lang/libc#3612](https://github.com/rust-lang/libc/pull/3612) which makes `confstr` available on all unix systems. Than I will make PR to the nix library.

That sounds like a great plan; thanks for finding that upstream PR and being willing to put in the effort to PR to `nix`. That will benefit other `nix` consumers.

reubeno:
Thanks for your work on this! Closing this now that the changes have been integrated into #402 and merged into main.

Issue #202, opened 2024-10-13
correct implementation of the `physical -P` and `logical -L` directories in `cd`, `pwd` and `dirs`


39555:
I've started working on correctly handling symlinked paths and have some questions about `cwd` implementation details.

- Why does brush use its own `Shell.working_dir` instead of the system's `std::env::set_current_dir/get_current_dir`? It's because a subshell is not a new process?

reubeno:
> It's because a subshell is not a new process?

Yes, that's the main reason. It's been a source of friction along the way, but in practice, actual interactions with the filesystem have been relatively limited to I/O redirection, the `source` builtin, `cd`/`pushd`/`popd`, and a few other places.

If you look for everywhere the `shell` object can get cloned--and there's more places than would be ideal--you'll see where we leverage this.

PR #219, opened 2024-10-20
feat: implement physical and logical cwd

39555:
After diving into the system's paths rabbit hole for a while, I've implemented cross-platform `cwd` handling for the shell, along with support for `cd -L -P` and `pwd -L -P`.

Closes: #202

With this PR, the shell now maintains the logical current working directory alongside with the physical one. To correctly create an absolute path and expand all `..` and `.`  from the logical path, the functions have been added to the `brush_core::sys::fs` crate:
- `expand_tilde_with_home` - expands the tilde using the given `home` directory
- `make_absolute` - makes the path absolute based on the given `base` path.
- `normalize_lexically` - performs canonicalization without touching the filesystem.

## Posix notes:
There is a special handling for paths with double slashes, such as `//users/home`
> If a pathname begins with two successive <slash> characters, the first component following the leading <slash> characters may be interpreted in an implementation-defined manner, although more than two leading <slash> characters shall be treated as a single <slash> character.

See: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13

## Windows notes
There is a lot of special handling for Windows weird path formats:
- UNC `\\server\share\my_dir`.
- Verbatim paths `\\?\C$\super_path`, this paths must not be normalized.
- Device namespace `\\.\Pipe` and UNC with the device namespace `\\.\UNC\`.
- volume relative `C:iamrelative`.

See: https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
See: The book of windows file path: https://chrisdenton.github.io/omnipath/

I've written a lot of tests, trying to cover all the special cases.

### DIFF
--- a/brush-core/src/builtins/cd.rs
+++ b/brush-core/src/builtins/cd.rs
@@ -9,11 +9,11 @@ use crate::{builtins, commands};
 #[derive(Parser)]
 pub(crate) struct CdCommand {
     /// Force following symlinks.
-    #[arg(short = 'L')]
+    #[arg(short = 'L', overrides_with = "use_physical_dir")]
     force_follow_symlinks: bool,
 
     /// Use physical dir structure without following symlinks.
-    #[arg(short = 'P')]
+    #[arg(short = 'P', overrides_with = "force_follow_symlinks")]
     use_physical_dir: bool,
 
     /// Exit with non zero exit status if current working directory resolution fails.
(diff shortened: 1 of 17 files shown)


reubeno:
Thanks for diving deep into this; I'll need a bit of time to review the changes here. Is there a reasonable way to break this up into multiple, smaller changes?

39555:
I don't think so, maybe `cd` and `pwd`, but they just call the new shell api `Shell.set_current_working_dir_from_logical`.

All changes are based on the new `Cwd` type instead of PathBuf for `Shell.working_dir`. `Cwd` contains both a physical PathBuf and a logical PathBuf. When we set a new cwd with `Shell.set_current_working_dir_from_logical`, the path should be normalized but without using syscalls (for preserving symlinks):

1. expand tilde
2. make it absolute based on our current physical or logical cwd
3. expand dots

For startup cwd Inside `Shell::new`, we try to get the logical path from the PWD environment variable if it is present.

In other changed files, such as dirs or interp, I fixed the usage of `working_dir` to use the physical path for now. This should be addressed in other PRs if completion needs the logical path instead.

reubeno:
My apologies for having put this on hold for so long, @39555 --thanks again for putting so much hard work into this and being patient while we addressed other areas of the project. We've since cleared out a number of the most commonly hit gaps in `brush` and I'd like to be able to get through the backlog of other good changes -- this one included.

I recall you mentioning in another issue that you may not have bandwidth for contributions right now. Is that the case? If so, would you be open to me helping rebase this against latest changes and shepherd it through? I'm also okay waiting if you'd prefer to come back to this yourself at some point in the future.

39555:
Hi @reubeno ! Yes it would be great if you could help me finalize this feature. Big thanks!

reubeno:
Thanks again for all the great work on this! We merged the core work here and are intentionally going to hold off on some of the complex Windows logic (which I think would make great additions to helper crates?). Hopefully we can avoid losing this code for the future...

PR #221, opened 2024-10-21
feat: implement `kill -l`

39555:
Utilize `nix::signal::Signal` to print the all signals.

Note that `0 - EXIT` is missing for now because it is not posix, and I didn't find any documentation for the `EXIT` signal, except for the `trap` builtin https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html

> If a sigspec is 0 or EXIT, arg is executed when the shell exits.

### DIFF
--- a/brush-shell/tests/cases/builtins/kill.yaml
+++ b/brush-shell/tests/cases/builtins/kill.yaml
@@ -4,7 +4,8 @@ cases:
     ignore_stderr: true
     stdin: |
       for i in $(seq 1 31); do kill -l $i; done
-      for i in $(kill -l | sed -e "s/[[:digit:]]*)//g"); do echo $i; done
+      # limit the number of signals to 32. Realtime signals are not implemented yet.
+      for i in $(kill -l | sed -e "s/[[:digit:]]*)//g"); do echo $i; done | head -32
       # invalid option
       kill -l 9999
       kill -l HUP
--- a/brush-shell/tests/cases/builtins/kill.yaml
+++ b/brush-shell/tests/cases/builtins/kill.yaml
@@ -6,7 +6,6 @@ cases:
     # 1) .. 2) .. 3)  .. etc. instead of just a simple list
     stdin: |
       # kill -l
-      # kill -l 0 # TODO: EXIT signal
       kill -l 1
       kill -l 2
       kill -l 3
@@ -36,8 +35,14 @@ cases:
       kill -l 28
       kill -l 29
       kill -l 30
+      kill -l 31
       # invalid option
       kill -l 9999
       kill -l HUP
+      kill -l iNt
+      kill -l int
       kill -l SIGHUP
+      kill -l EXIT
+      kill -l ERR
+      kill -l DEBUG
 
--- a/brush-shell/tests/cases/builtins/kill.yaml
+++ b/brush-shell/tests/cases/builtins/kill.yaml
@@ -2,40 +2,9 @@ name: "Builtins: kill"
 cases:
   - name: "kill -l"
     ignore_stderr: true
-    # note that we dont use 'kill -l' because of the shape of the bash list
-    # 1) .. 2) .. 3)  .. etc. instead of just a simple list
     stdin: |
-      # kill -l
-      kill -l 1
-      kill -l 2
-      kill -l 3
-      kill -l 4
-      kill -l 5
-      kill -l 6
-      kill -l 7
-      kill -l 9
-      kill -l 10
-      kill -l 11
-      kill -l 12
-      kill -l 13
-      kill -l 14
-      kill -l 15
-      kill -l 16
-      kill -l 17
-      kill -l 18
-      kill -l 19
-      kill -l 20
-      kill -l 21
-      kill -l 22
-      kill -l 23
-      kill -l 24
-      kill -l 25
-      kill -l 26
-      kill -l 27
-      kill -l 28
-      kill -l 29
-      kill -l 30
-      kill -l 31
+      for i in $(seq 1 31); do kill -l $i; done
+      for i in $(kill -l | sed -e "s/[[:digit:]]*)//g"); do echo $i; done
       # invalid option
       kill -l 9999
       kill -l HUP
@@ -43,4 +12,5 @@ cases:
       kill -l int
       kill -l SIGHUP
       kill -l EXIT
+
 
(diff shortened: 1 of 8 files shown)


39555:
Cool thanks! I will look at that

reubeno:
> 1. I've noticed that in bash `kill -l`can print selected signals e.g., `kill -l HUP 2 TERM` would print
> _[snipped]_
> 
> while `trap -l` always prints all the signals. Should the `trap -l` also be able to print selected signals?

Yes, I think it would be best for `trap -l` to behave like `bash`'s implementation does. In theory, I suppose that some script *could* use it to decode signal numbers / names. Seems unlikely, but couldn't hurt to be consistent there.

> 2. Maybe we should also print `EXIT`, `DEBUG` and `ERR`? Bash is highly inconsistent in this regard. For example, when you print all with `kill -l`, Bash doesn't show these signals, but if you type `kill -l 0`, it prints `EXIT`. Furthermore, when you type `kill -l 33` it shows nothing but when you type `kill -l ERR` it shows `33`

I see the inconsistencies as well; I'm assuming some of it maybe intentional, and other aspects may come down to the way the shared code happened to be structured.

For `kill`, `EXIT`, `DEBUG`, and `ERR` don't really make sense. They aren't really signals that can be sent to a job or process. They're virtual signals that can be trapped via `trap`. I'd vote for us avoiding duplicating much logic between `trap` and `kill`, and where possible without too much effort, hiding these 3 virtual signals from `kill`.

> 3. `trap` is case insensitive  (e.g `trap int` works), while `kill -l` is not. Should `kill` also be case insensitive?

I see `trap` call out its case insensitivity; did you see `kill -l` being case *sensitive*? From my quick tests on a few versions of `bash`, I saw it behave in a similar case _insensitive_ fashion:

```bash
bash$ kill -l int
2
bash$ kill -l iNt
2
bash$ kill -l INT
2
bash$ kill -l SiGiNT
2
bash$ kill -l SIGINT
2
bash$ kill -l sigint
2
```

> 4. when `kill -l` bash prints full names (e.g `SIGINT`) but when `kill -l 2`, it prints `INT`. Should it be always `SIGINT` or should I mimic the bash behavior?

I think it makes sense to mimic the bash behavior for `kill -l <signal>`, just on the extreme off-chance that some caller would notice the difference. For `kill -l` (listing all of them), I don't think I have a strong preference one way or the other.

39555:
Interesting, the value of ERR and DEBUG is based on `NSIG` macro. On macos Err = 33, DEBUG = 32, on linux Err = 65, Debug = 64. 

Also I found a line in the bash change log for the version 2.01.1

> s.  A fix was made so that you can no longer trap `SIGEXIT' or `SIGDEBUG' --
    only `EXIT' and `DEBUG' are accepted.

39555:
@reubeno Should I try to use `libc::SIGRTMAX` and `libc::SIGN` where possible and fallback to the `32` or leave DEBUG, ERR without numbers (than it would be `kill -l ERR` -> `ERR: invalid signal specification`)?

reubeno:
What would your recommendation be? I think the requirements here are that the numbers for these virtual signals don't conflict with any of the *real* signals. They're not expected to be portable across systems, to my knowledge.

39555:
> What would your recommendation be? I think the requirements here are that the numbers for these virtual signals don't conflict with any of the _real_ signals. They're not expected to be portable across systems, to my knowledge.

Fоr now, I made it without numbers. It is pretty accurate, `kill` just prints a name instead:
```
kill -l INT
2
kill -l DEBUG
DEBUG
```
in `kill -l` and `trap -l` ERR and DEBUG are not listed, similar to Bash.

reubeno:
Sounds good to me. I'll try to review this in the next day or so. (I'd gotten a bit busy looking at an incoming issue report that seemed more severe (syntax parse failures).)

brush-shell/tests/cases/builtins/kill.yaml
```diff
@@ -0,0 +1,46 @@
...
+    stdin: |
+      # kill -l
+      kill -l 1
+      kill -l 2
+      kill -l 3
+      kill -l 4
+      kill -l 5
+      kill -l 6
+      kill -l 7
+      kill -l 9
```
reubeno: Was `kill -l 8` intentionally omitted? For that matter, what about going with a `for` loop for better maintainability/readability?

brush-shell/tests/cases/builtins/kill.yaml
```diff
@@ -0,0 +1,46 @@
+name: "Builtins: kill"
+cases:
+  - name: "kill -l"
+    ignore_stderr: true
+    # note that we dont use 'kill -l' because of the shape of the bash list
```
reubeno: Any other ideas on how to get some test coverage for `kill -l`?

I'm not sure how much we want to assume about the format of `bash`'s output, but something like this would extract a "flat" list from its implementation of `kill -l`:

```bash
siglist=($(kill -l | sed -e "s/[[:digit:]]\+)//g"))
for s in ${siglist[@]}; do echo $s; done
```

reubeno (commented):
Changes generally look good to me; just a few comments related to testing. Thanks for putting in the work to get `trap` and `kill` to agree on things. I think that will pay off over time.

brush-shell/tests/cases/builtins/kill.yaml
39555: @reubeno should we also hide EXIT from the printed list? The test is currently failing because of EXIT

reubeno: I don't have a strong preference on that. If we want to allow it to be different, we can always intentionally have the test `grep -v` / exclude it so it doesn't show up in a delta.

39555:
There should be additional signals for linux that are not listed in nix::Signal...

39555:
There is ongoing pr for realtime signals in nix https://github.com/nix-rust/nix/pull/2451

reubeno:
> There is ongoing pr for realtime signals in nix [nix-rust/nix#2451](https://github.com/nix-rust/nix/pull/2451)

Is that something we could integrate with later (without preventing us from moving forward with this set of changes now)?

39555:
Yeah. For now just realtime signals are missing (the linux failed test shows them)

reubeno:
> Yeah. For now just realtime signals are missing (the linux failed test shows them)

Sounds like something we could track with a TODO follow-up and exclude from the diff (via grep) for now?

reubeno:
Your latest changes are looking good to me. Please let us know when you're ready for final review before merging?

39555:
@reubeno I'm ready

reubeno (approved):

PR #227, opened 2024-10-24
feat: implement `read -t`

39555:
`read -t` reading stdin with timeout. Implemented with polling on unix and with a little bit insane and unsafe{} async implementation on windows. It works but work in progress. Just want to discuss different directions how we can properly implement this. Seems like non of the popular IO libraries support Windows's stdin/file IO operations because it is really a madness

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -276,6 +276,7 @@ dependencies = [
  "tracing",
  "uzers",
  "whoami",
+ "windows 0.56.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


39555:
Just read the crossterm sources, maybe crossterm can help https://github.com/crossterm-rs/crossterm. Brush already have it as indirect dependency. I've noticed that brush has `sys::terminal` with a lot of TODO's and crossterm has that functions as well.

`read` implementation in nushell is very clean https://github.com/nushell/nushell/blob/af9c31152a3926c5c85bb67e1125cdb52abe3ace/crates/nu-command/src/platform/input/input_.rs#L14

reubeno:
I skimmed your changes but didn't read them too closely yet. You weren't kidding when you said the Windows side would be complicated!

I'm not against getting it working on Windows, but I think the complexity required is beyond the scope of what we should do and maintain in `brush`. With some exceptions, we've _tried_ to keep away from lower-level console/terminal I/O, instead delegating those responsibilities to `reedline`, `crossterm`, `nix`, et al.

My vote here would be to:

* Leverage existing abstractions in terminal/console crates to keep the built-in implementation relatively clean and platform-agnostic; where those abstractions don't do enough, abstract away any truly platform-specific details via `sys`.
* Lean on existing standard crates for Unix implementation.
* Leave the Windows side stubbed out / unimplemented for now.
* Consider contributing/integrating your Windows-specific solution with one of the terminal/console crates in the crates ecosystem.

reubeno:
Thanks for all your work on this! We adapted the non-Windows portions of these changes and got them merged :smile:.

Issue #104, opened 2024-06-29
Implement `|&` extension


reubeno:
In bash land, `|&` is an extension that's basically a shorthand for `2>&1 |`. We have support for both `2>&1` and `|` -- we just need to make sure we parse this sequence correctly.

PR #240, opened 2024-10-25
feat: implement `|&` extension

39555:
Closed: #104

### DIFF
--- a/brush-parser/src/parser.rs
+++ b/brush-parser/src/parser.rs
@@ -255,7 +255,7 @@ peg::parser! {
         rule bang() -> bool = specific_word("!") { true }
 
         rule pipe_sequence() -> Vec<ast::Command> =
-            c:command() ++ (specific_operator("|") linebreak()) { c }
+            c:command() ++ ((non_posix_extensions_enabled() specific_operator("|&")) / specific_operator("|") linebreak()) { c }
 
         // N.B. We needed to move the function definition branch up to avoid conflicts with array assignment syntax.
         rule command() -> ast::Command =
@@ -587,11 +587,27 @@ peg::parser! {
                 w:word() {
                     ast::CommandPrefixOrSuffixItem::Word(ast::Word::from(w))
                 }
-            )+ { ast::CommandSuffix(s) }
+            )+ e:input_output_pipe_extension()? {
+                let mut s = s;
+                if let Some(e) = e {
+                    s.push(ast::CommandPrefixOrSuffixItem::IoRedirect(e))
+                }
+                ast::CommandSuffix(s)
+            }
 
         rule redirect_list() -> ast::RedirectList =
-            r:io_redirect()+ { ast::RedirectList(r) } /
+            r:io_redirect()+ e:input_output_pipe_extension()? {
+                let mut r = r;
+                if let Some(e) = e {
+                    r.push(e)
+                }
+                ast::RedirectList(r)
+            } /
             expected!("redirect list")
+
+
+        rule input_output_pipe_extension() -> ast::IoRedirect =
+            non_posix_extensions_enabled() &specific_operator("|&") { ast::IoRedirect::File(Some(2), ast::IoFileRedirectKind::DuplicateOutput, ast::IoFileRedirectTarget::Fd(1)) }
 
         // N.B. here strings are extensions to the POSIX standard.
         rule io_redirect() -> ast::IoRedirect =
@@ -909,6 +925,23 @@ esac\
         assert_eq!(command.cases[0].patterns.len(), 1);
         assert_eq!(command.cases[0].patterns[0].flatten(), "x");
 
+        Ok(())
+    }
+
+    #[test]
+    fn parse_redirection() -> Result<()> {
+        let input = r"echo |& wc";
+
+        let tokens = tokenize_str(input)?;
+        let program = super::token_parser::program(
+            &Tokens {
+                tokens: tokens.as_slice(),
+            },
+            &ParserOptions::default(),
+            &SourceInfo::default(),
+        )?;
+        let snapshot = r#"Program { complete_commands: [CompoundList([CompoundListItem(AndOrList { first: Pipeline { bang: false, seq: [Simple(SimpleCommand { prefix: None, word_or_name: Some(Word { value: "echo" }), suffix: None }), Simple(SimpleCommand { prefix: None, word_or_name: Some(Word { value: "wc" }), suffix: None })] }, additional: [] }, Sequence)])] }"#;
+        assert_eq!(snapshot, format!("{:?}", program));
         Ok(())
     }
 }
--- a/brush-parser/src/parser.rs
+++ b/brush-parser/src/parser.rs
@@ -255,36 +255,21 @@ peg::parser! {
         rule bang() -> bool = specific_word("!") { true }
 
         pub(crate) rule pipe_sequence() -> Vec<ast::Command> =
-            c:(c:command() r:input_output_pipe_extension()? {
+            c:(c:command() r:&pipe_extension_redirection()? {? // check for `|&` without consuming the stream.
                 let mut c = c;
-                if let Some(r) = r {
-                    match &mut c {
-                        ast::Command::Simple(c) => {
-                            let r = ast::CommandPrefixOrSuffixItem::IoRedirect(r);
-                            if let Some(suffix) = &mut c.suffix {
-                                suffix.0.push(r);
-                            } else {
-                                let v = vec![r];
-                                c.suffix = Some(ast::CommandSuffix(v));
-                            }
-                        }
-                        ast::Command::Compound(c, l) => {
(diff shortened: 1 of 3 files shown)


brush-parser/src/parser.rs
```diff
@@ -911,4 +927,21 @@ esac\
...
+
+        let tokens = tokenize_str(input)?;
+        let program = super::token_parser::program(
+            &Tokens {
+                tokens: tokens.as_slice(),
+            },
+            &ParserOptions::default(),
+            &SourceInfo::default(),
+        )?;
+        let snapshot = r#"Program { complete_commands: [CompoundList([CompoundListItem(AndOrList { first: Pipeline { bang: false, seq: [Simple(SimpleCommand { prefix: None, word_or_name: Some(Word { value: "echo" }), suffix: None }), Simple(SimpleCommand { prefix: None, word_or_name: Some(Word { value: "wc" }), suffix: None })] }, additional: [] }, Sequence)])] }"#;
```
reubeno: Thanks for adding a test case here. Having `Debug` implemented for human debugging is helpful, but we should avoid taking a programmatic dependency on its specific textual format.

In some other test cases, we match against the `ast::Program` (or more specific AST nodes). Could you go with that approach here?

brush-parser/src/parser.rs
```diff
@@ -255,7 +255,7 @@ peg::parser! {
         rule bang() -> bool = specific_word("!") { true }
 
         rule pipe_sequence() -> Vec<ast::Command> =
-            c:command() ++ (specific_operator("|") linebreak()) { c }
+            c:command() ++ ((non_posix_extensions_enabled() specific_operator("|&")) / specific_operator("|") linebreak()) { c }
```
reubeno: Maybe there's something I'm missing, but doesn't this rule now accept `x | y` or `x |& y` yet discard the difference between them?

brush-parser/src/parser.rs
```diff
@@ -587,12 +587,28 @@ peg::parser! {
...
+                let mut s = s;
+                if let Some(e) = e {
+                    s.push(ast::CommandPrefixOrSuffixItem::IoRedirect(e))
+                }
+                ast::CommandSuffix(s)
+            }
 
         rule redirect_list() -> ast::RedirectList =
-            r:io_redirect()+ { ast::RedirectList(r) } /
+            r:io_redirect()+ e:input_output_pipe_extension()? {
```
reubeno: I don't recall offhand; did you confirm that `|&` is actually allowed anywhere a redirect list can be present (e.g., not just where `|` can be used)?

brush-parser/src/parser.rs
```diff
@@ -587,12 +587,28 @@ peg::parser! {
                 w:word() {
                     ast::CommandPrefixOrSuffixItem::Word(ast::Word::from(w))
                 }
-            )+ { ast::CommandSuffix(s) }
+            )+ e:input_output_pipe_extension()? {
```
reubeno: Would this mean that `|&` is only allowed at the very end of a suffix (and not intermingled with other suffixes)?

brush-parser/src/parser.rs
39555: This rule acts as a separator between commands. |& actually peeks without consuming inside the command. Is there a better approach maybe?

brush-parser/src/parser.rs
39555: @reubeno yes, only at the end of the command, where `|` is used. The `command` parser checks if the input has `|&` after all suffixes, and if so, appends the redirection but without consuming `|&`. The `pipe_sequence` parser then splits commands by `|&` or '|'.
https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Redirecting-Standard-Output-and-Standard-Error
> If ‘|&’ is used, command1’s standard error, in addition to its standard output, is connected to command2’s standard input through the pipe; it is shorthand for 2>&1 |. <ins>This implicit redirection of the standard error to the standard output is performed after any redirections specified by command1</ins>.

brush-parser/src/parser.rs
39555: Can you give me some examples what to you mean? Now redirect_list is used after compound commands and function bodies

brush-parser/src/parser.rs
reubeno: It looks like `|&` can be used in the same places that `|` can be used:

image

...and not just anywhere output redirection (`>`) can be used, e.g.:

image

I think that means the only place where it should really show up in the grammar, at least as it's currently factored, is in `pipe_sequence()`.

brush-parser/src/parser.rs
reubeno: My understanding is that the `++` combinator _does_ consume the indicated separator (`|&` or `|` in this case). So this might be the right place put this operator, but we need to capture the fact that it was used instead of `|`.

From https://docs.rs/peg/latest/peg/:

> `expression ++ delim` - Delimited repeat (one or more): match one or more repetitions of expression delimited with delim and return the results as a Vec.

brush-parser/src/parser.rs
39555: same in brush. 
```
brush$ ls |&
>
ERROR main: syntax error near token `
' (line 1 col 6)
brush$ |& ls
ERROR main: syntax error near token `|&' (line 1 col 1)
```
`|& ls` is an `io_redirect+` inside cmd_prefix not a `redirect_list`. `redirect_list` is used only at the very end. Maybe I should add input_output_pipe_extension() as a separate parsing step to the compound command, instead of inside the redirect_list.
I don't know how to approach `&|` inside the pipe_sequence: `c:command()++ special_operator('|')`. maybe `map` or something? I will give it a try.

39555:
@reubeno I've just pushed attempt 2 with the logic only inside the `pipe_sequence` rule

brush-parser/src/parser.rs
```diff
@@ -254,8 +254,37 @@ peg::parser! {
             bang:bang()? seq:pipe_sequence() { ast::Pipeline { bang: bang.is_some(), seq } }
         rule bang() -> bool = specific_word("!") { true }
 
-        rule pipe_sequence() -> Vec<ast::Command> =
-            c:command() ++ (specific_operator("|") linebreak()) { c }
+        pub(crate) rule pipe_sequence() -> Vec<ast::Command> =
```
reubeno, line 257: This logic seems a bit cleaner; thanks for reworking the changes.

It's also getting harder to read and easily understand the `pipe_sequence` rule. Could you please break it up into smaller separable chunks? For example, updating a `Command` with the added redirect seems like something that could be handled in a method on `Command`.

For the parsing of the actual separator operator, that feels worthwhile breaking off into a helper rule (e.g., `pipe_operator()`) that encapsulates the `|` vs. `|&` on line 285.

brush-parser/src/parser.rs
```diff
@@ -254,8 +254,37 @@ peg::parser! {
...
+                        ast::Command::Compound(c, l) => {
+                            if let Some(r_list) = l {
+                                r_list.0.push(r);
+                            } else {
+                                let v = vec![r];
+                                *l = Some(ast::RedirectList(v));
+                            }
+
+                        }
+                        _ => { // TODO: syntax error!
```
reubeno: This begs the question of which kinds of commands can be followed by redirection operators. It looks like we already have a gap in `brush` with extended tests and others. For example, the following is accepted in `bash` but generates a syntax error in `brush`:

```
[[ -f path ]] >/dev/null
```

Function definitions can also have redirection lists, and it *seems* like `bash` will accept something like:

```bash
function foo() {
    echo hi
} |& wc -l
```

In other words, it seems like all the different command types should technically come with redirection list that we can add to?

brush-parser/src/parser.rs
```diff
@@ -911,4 +949,29 @@ esac\
...
+            &SourceInfo::default(),
+        )?;
+
+        assert_eq!(seq.len(), 2);
+        assert!(matches!(seq[0], ast::Command::Simple(..)));
+        if let ast::Command::Simple(c) = &seq[0] {
+            let c = c.suffix.as_ref().unwrap();
+            assert!(matches!(
+                c.0[0],
+                ast::CommandPrefixOrSuffixItem::IoRedirect(..)
```
reubeno: Do you think it's worth further asserting that the right Fds are in the `IoRedirect`?

reubeno:
Thanks for the updates; this is looking pretty good to me!

Two last final comments:

* It may be worth reversing the order of the `|` and `|&` cases on the off-chance there's performance benefit. (I'd expect `|` to be the more common case but don't know if the difference has any meaningful impact.) The benchmarks look slightly impacted, but they're pretty noisy.

* I'm supportive of getting this merged; could you file a follow-up issue tracking the redirection gap we discussed around extended tests and/or function definitions. We could tackle the unimplemented branches of your new helper function as part of that.

reubeno:
Merged! Thanks for the contribution :)

Issue #242, opened 2024-10-26
Support `|&` for `function` and `extended test`


39555:
Currently this syntax is not supported:
- `foo() { echo 1; } |& cat`
- `[[ 1 -eq 2 ]] |& cat`

- Furthermore non of redirections are supported for extended tests `[[ 1 -eq 2 ]]  2>&1  | cat`.

PR #244, opened 2024-10-26
feat: implement `|&` for function declarations

39555:
Closed: #242 
WIP. Adding `|&` for function declarations turns out to be quite straightforward. However `extended tests` require some move work

### DIFF
--- a/brush-parser/src/parser.rs
+++ b/brush-parser/src/parser.rs
@@ -847,27 +847,31 @@ fn add_pipe_extension_redirection(c: &mut ast::Command) -> Result<(), &'static s
         ast::IoFileRedirectKind::DuplicateOutput,
         ast::IoFileRedirectTarget::Fd(1),
     );
+
+    #[inline]
+    fn add_to_redirect_list(l: &mut Option<ast::RedirectList>, r: ast::IoRedirect) {
+        if let Some(l) = l {
+            l.0.push(r);
+        } else {
+            let v = vec![r];
+            *l = Some(ast::RedirectList(v));
+        }
+    }
+
     match c {
         ast::Command::Simple(c) => {
             let r = ast::CommandPrefixOrSuffixItem::IoRedirect(r);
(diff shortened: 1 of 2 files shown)


brush-parser/src/parser.rs
```diff
@@ -847,27 +847,31 @@ fn add_pipe_extension_redirection(c: &mut ast::Command) -> Result<(), &'static s
         ast::IoFileRedirectKind::DuplicateOutput,
         ast::IoFileRedirectTarget::Fd(1),
     );
+
+    #[inline]
```
reubeno: Why `#[inline]`? For what it's worth, the best write-up I've seen with guidance on this is here: https://matklad.github.io/2021/07/09/inline-in-rust.html.

reubeno (commented):
Apart from my comment on `#[inline]`, the changes look good. Did you want to find a way to move forward with this PR, or did you want to hang onto it until the extended tests can get resolved?

(Note, per earlier comments, I think we already have an issue with redirection on extended tests in general; I'd recommend that we get `[[ -f somefile ]] >/dev/null` parsing correctly first and then add on `|&` support.

brush-parser/src/parser.rs
39555: This function is for not duplicating both match branches, but I would prefer not to write a `macro!` for this. So I decided to tell the compiler to inline the function body. Actually it might inline it anyway... so I can remove it...

39555:
@reubeno let's go now, with functions only

brush-parser/src/parser.rs
reubeno: >Actually it might inline it anyway... so I can remove it...

Okay, please remove it and then I can do a final review.

Taking inspiration from the page linked above, it makes sense to avoid proactively tagging functions for inlining unless we analyze performance and identify that it would make a meaningful difference. (For what it's worth, release builds of `brush` do have link-time optimization turned on as well.)

reubeno (approved):

PR #246, opened 2024-10-27
build: run cargo udeps in checks

reubeno:
Checks for unneeded dependencies in the project's crates.


reubeno:
Looks like this needs nightly.

PR #250, opened 2024-10-27
Experiment: try to run interactive tests with Alacritty

39555:
WIP. I've connected `expectrl` to the Alacritty terminal. It can run any command with `reedline` backend, for example the test `run_pipeline_interactively`, but the command output is somehow messed up with OSC sequences and prompts, so precise control how it is done in `Session::exec_output("jobs")` is not possible for now

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,6 +40,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "alacritty_terminal"
+version = "0.24.2-dev"
+source = "git+https://github.com/39555/alacritty?branch=spawn-for-testing#8bae9dd121058faf75665f73da8e7d5b8bab0a16"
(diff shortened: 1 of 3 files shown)


reubeno:
Coming back to this one... what's your recommendation for what we should do here? I also see this not-yet-merged PR to `reedline` also looking at using `alacritty` for testing:

https://github.com/nushell/reedline/pull/847

>...command output is somehow messed up with OSC sequences and prompts

Any thoughts on how to deal with this?

39555:
Hm the PR that you linked uses the text from the terminal grid instead of the pty for comparison. Looks much nicer. Maybe that way, the output order problem goes away. If this is the case than the alacritty is the perfect solution for interactive tests

Issue #233, opened 2024-10-24
generalize $RANDOM impl to support dynamic vars like `SHELLOPTS` + `BASHOPTS`


reubeno:
I think it's worth looking at generalizing what we already do for `$RANDOM` -- i.e., the existence of a special variable whose reads are satisfied by custom functions. We could have some way to provide dynamically accurate values of `SHELLOPTS`, for example, without needing the code managing options to remember to update the variable. As a side benefit, this also allows us to do less work at shell initialization time to compute and format the right strings.

reubeno:
This has been implemented in #360.

PR #251, opened 2024-10-27
feat: implement  `$SHLVL` and `$BASH_SUBSHELL`

39555:
Implementation of `$SHLVL` and `$BASH_SUBSHELL` variables. https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html.

### BASH_SUBSHELL

> Incremented by one within each subshell or subshell environment when the shell begins executing in that environment. The initial value is 0. If BASH_SUBSHELL is unset, it loses its special properties, even if it is subsequently reset. 

Im not sure how to test `SHLVL`, because it requires starting a new shell inside the test.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -79,11 +79,17 @@ pub struct Shell {
 
 impl Clone for Shell {
     fn clone(&self) -> Self {
+        let mut env = self.env.clone();
+
+        if !self.options.sh_mode {
+            increment_level_var(&mut env, "BASH_SUBSHELL", 1, false).unwrap();
+        }
+
         Self {
             traps: self.traps.clone(),
             open_files: self.open_files.clone(),
             working_dir: self.working_dir.clone(),
-            env: self.env.clone(),
+            env,
             funcs: self.funcs.clone(),
             options: self.options.clone(),
             jobs: jobs::JobManager::new(),
@@ -279,7 +285,13 @@ impl Shell {
             )?;
         }
 
+        increment_level_var(&mut env, "SHLVL", 1, true)?;
+
         if !options.sh_mode {
+            let mut lvl = ShellVariable::new("0".into());
+            lvl.export();
+            env.set_global("BASH_SUBSHELL", lvl)?;
+
             if let Some(shell_name) = &options.shell_name {
                 env.set_global("BASH", ShellVariable::new(shell_name.into()))?;
             }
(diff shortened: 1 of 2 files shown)


reubeno:
Thanks for working on closing this gap!

_One main question for you_: what do you think about the idea of lazily computing the value of these variables? I've been toying with the idea of generalizing the way we implement `$RANDOM` and having the notion of a "dynamically computed" `ShellValue`. It would let us insert the variable in the environment at initialization time but not spend energy on computing it, stringifying the result, and storing it. When actually needed for an expansion, we can evaluate it via call out to a `Fn` of some sort, which may look at the `Shell` first to generate the correct value. For some other variables like `$BASHOPTS` (tracked by #233) it also avoids us needing to keep updating the variable when its value would change (i.e., whenever options are updated) and reduce it down to just needing a way to poll the correct result at expansion-time.

reubeno:
Ok, I went back and tried to pull on the idea I mention above ^^^. I still think it's worth pursuing, but I think it's sufficiently complicated not to rush it.

I'll do a proper review of the changes in this PR so we can get them in.

brush-core/src/shell.rs
```diff
@@ -1182,3 +1194,21 @@ fn parse_string_impl(
     tracing::debug!(target: trace_categories::PARSE, "Parsing string as program...");
     parser.parse(true)
 }
+
+// parse a numeric variable from a string, increment it by 1, then re-export
+fn increment_level_var(
```
reubeno, line 1199: `shell.rs` is already, well, longer than I'd like. Could you turn this into a `pub(crate)` helper in `variables.rs` or similar?

For what it's worth, it seems more understandable to me to have the `default` value expressing the pre-increment default (i.e., `0` here).

brush-shell/tests/cases/compound_cmds/subshell.yaml
```diff
@@ -23,3 +23,37 @@ cases:
   - name: "Piped subshell usage"
     stdin: |
       (echo hi) | wc -l
+
+  - name: "SHLVL subshell"
```
reubeno, line 27: It seems like we need some tests to ensure that `SHLVL` will work correctly across process invocations (i.e., nested shells and not just subshells). Thoughts on how to do that? Nested invocation of `$0`? Call through `bash` in the middle?

39555: @reubeno I thought about it and created a test using $SHELL but is points to nowhere inside the Oracle. Now, looking at it, I learned about `$0`, so I will try using that instead. I'm not sure yet how interactivity works there.
```
$0
$0
$0 -c 'echo $SHLVL; $0'
```

reubeno:
Thanks for leading the charge here. We ended up taking another pass at the "dynamic variable" idea in #233 and have implemented it in #360, using it to implement `SHLVL` and `BASH_SUBSHELL`. Your changes were helpful in identifying how to implement their semantics.

I'll now close out this PR.

Issue #234, opened 2024-10-25
heredoc doesn't work inside of command substituiton


Quackdoc:
example bash:

```sh
#!/bin/sh

test1=$(cat <<EOF
test
test test
test test test
EOF
)
test2=$(echo test 2)
echo $test1
echo $test2
cat <<TEST3
test 3
TEST3

```

command output
```ps
➜  xe git:(master) ✗ bash test.sh
test test test test test test
test 2
test 3
➜  xe git:(master) ✗ brush test.sh
ERROR main: unterminated here document sequence; tag(s) found at: [line 1 col 14] (detected near line 5 column 1)

test 2
test 3
```

39555:
@reubeno seems like it gets tokenized twice? 
```
$(<<EOF
test
EOF
)
```
after the `tokenize_str` 
```
Word(
        "$(<<EOFtest\n\n)", //  <-- the input inside qoutes already processed by the tokenizer, but it sends it to the subshell as a raw string
        TokenLocation {
            start: SourcePosition {
                index: 0,
                line: 1,
                column: 1,
            },
            end: SourcePosition {
                index: 18,
                line: 4,
                column: 2,
            },
        },
    ),
    Operator(
        "\n",
        TokenLocation {
            start: SourcePosition {
                index: 18,
                line: 4,
                column: 2,
            },
            end: SourcePosition {
                index: 19,
                line: 5,
                column: 1,
            },
        },
    ),
]
```

reubeno:
@39555 Yes. It's not ideal, but for the initial implementation we landed on the tokenizer yielding the full text of the command substitution in the word token that contains it (potentially along with other text, substitutions, etc.). The recursion was added to ensure that we can find the end of the substitution, but that's really all it's needed for here.

I recall having explored alternative options (e.g., support a `Word` storing a tree of sub-tokens) but selected this approach as the most practical to start with.

At this point, the proximate error seems to be that the newline between `EOF` and `test` gets lost in the mix.

39555:
@reubeno when the tokenizer processes the input, it replaces the last EOF with a newline for simplicity. Maybe because of that the subshell's tokenizer throws an error that the 'Here' is not closed? The first tokenizer should skip all the text inside the substitution unchanged.

reubeno:
Good question. I added a quick test case to cover this (and improved the error):

[...]

PR #252, opened 2024-10-28
fix: the heredoc inside a substitution

39555:
Addressed: #234 

I've got your branch and added a test to replicate the issue. `parser::tests::parse_subshell_with_heredoc`. When the subshell is invoked, the ending here tag is alredy stripped.

1. after the initial tokenezing, the word is `$(cat <<EOF 1\n\n)`
2. running the expansion
3. the word becomes `s = "cat <<EOF 1\n\n"`
4. starting a subshell: `let result = subshell.run_string(s, &params).await?;` in `expansion.rs`at the  L530.
5. getting an error:
```
Err( Tokenizing { inner: UnterminatedHereDocuments( "EOF1", "line 1 col 11", ), position: Some( SourcePosition { index: 22, line: 5, column: 1, }, ), },)
```
6. subshell returns exit_code 2

I don't know how to fix this issue yet though

### DIFF
--- a/brush-parser/src/parser.rs
+++ b/brush-parser/src/parser.rs
@@ -1021,4 +1021,90 @@ esac\
         }
         Ok(())
     }
+
+    #[test]
+    fn parse_subshell_with_heredoc() -> Result<()> {
+        let input = r"$(cat <<EOF
+1
+2 3
+4 5 6
+EOF
+)";
+
+        let tokens = tokenize_str(input)?;
+        let seq = super::token_parser::pipe_sequence(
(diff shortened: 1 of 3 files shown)


reubeno:
This is a tough one; thanks for helping to assess the problem. I took a deeper look at it, and I see a number of challenges with how our tokenization of here-documents affects subshells or command substitutions that embed them. In the longer term, we may want to revisit a few aspects of how the tokenizer works, but I think I've got a way to get it working. I'm going to create a separate (draft) PR with those changes.

Issue #254, opened 2024-11-03
Use `insta` or `expect_test` for snapshot based testing of the parser


39555:
What do you think about using either [`insta` ](https://crates.io/crates/insta), see examples:
- https://github.com/dtolnay/syn/blob/a777cff00528f270b43f40b0a58c5c26fc85a2bd/tests/test_parse_quote.rs#L12
- https://github.com/GitoxideLabs/gitoxide/blob/3fb989be21c739bbfeac93953c1685e7c6cd2106/gix/tests/gix/object/tree/diff.rs#L78
- https://github.com/RustPython/Parser/blob/5e9d9853e751240b28138efd165e793a7ee5dad4/parser/src/snapshots/rustpython_parser__context__tests__ann_assign_name.snap
- https://github.com/Kampfkarren/full-moon/blob/aaf57e4dd40e7272dd731cbaa0afa5c8abe6a3e3/full-moon/tests/cases/pass/anonymous-functions-3/ast.snap

or rust-analyzer's [`expect-test`](https://crates.io/crates/expect-test)?
- https://github.com/rust-lang/rust-analyzer/blob/f17a5bbfd0969ba2e63a74505a80e55ecb174ed9/crates/ide/src/syntax_tree.rs#L103

This approach is based on matching a `Debug` or any text representation, which allows us to reduce `assert_matches` boilerplate when matching AST nodes. The downside is that if the shape of any node changes, the affected snapshots need to be rebuilt with `cargo insta`.

reubeno:
I completely agree those parser/tokenizer tests aren't scaling well. I'd previously looked into `insta`; it looks really promising, just didn't seem like the right fit for our *compatibility* integration tests. I'm interested in the idea of using it with component-level tests focused on the tokenizer and parser.

_(Aside: we'd historically been light on the tokenizer/parser tests, as their results had been less stable--but also less important. As long as we provide the right end-to-end semantics that match `bash`, we're good to go. With that said, these tokenizer/parser tests can be extremely helpful in pinpointing regressions and unexpected behavior at lower levels. I frequently use VSCode's test running/debugging experiences to step through the code-under-test in some of those tests. Now that things are relatively stable, I support the idea of strengthening these lower-level tests.)_

[...]

PR #259, opened 2024-11-05
feat(ast): derive `PartialEq` and `Eq` for testing

39555:
#254

Instead of using `insta` with text representation we can actually use `assert_eq!` with a direct snapshot of the data structure. I've added one example `parser::tests::test_parse_program`to show how it looks like. 

The `#[derive]` only applies when `cfg_attr(test)` is enabled.

### benefits over `insta`
- simple
- no external tools needed
- rust-analyzer can rename fields
- syntax highlighting/formatting

### DIFF
--- a/brush-parser/src/ast.rs
+++ b/brush-parser/src/ast.rs
@@ -10,6 +10,7 @@ const DISPLAY_INDENT: &str = "    ";
 /// Represents a complete shell program.
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "fuzz-testing", derive(arbitrary::Arbitrary))]
+#[cfg_attr(test, derive(PartialEq, Eq))]
 pub struct Program {
     /// A sequence of complete shell commands.
     pub complete_commands: Vec<CompleteCommand>,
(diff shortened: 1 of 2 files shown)


reubeno:
I like the idea of deriving `Eq` and `PartialEq` for test configurations! Coupled with use of [`pretty_assertions`](https://crates.io/crates/pretty_assertions), which I started using in `brush-parser`, we'll get decent easy-to-read failure output.

As for the actual tests, I do like the idea of `insta` or similar to make it easier to generate/maintain the snapshots. What do you think?

I'm supportive of moving forward with this change--if you'd like to see it get in--and incrementally later more deeply evaluate adopting `insta` or similar.

39555:
Lets go with Eq/PartialEq for now. If we find we need insta, we can give it a try. I copied the snapshot from the Debug output, so generating is not a big deal.

By the way there is a strange failed test on debian, I don't understand why..

reubeno:
I've queued the failing Debian check for a re-run; we'll see if it succeeds. If it does, then this may mean the test in question is flaky.

> I copied the snapshot from the Debug output, so generating is not a big deal.

For now that's fine, but I don't think that scales. I can also see value in having the test cases defined in their own easy-to-read (and grep) subdirs. Before we add too many more cases, we should look into `insta` or similar.

reubeno (approved):

PR #265, opened 2024-11-08
test(parser): snapsot testing with insta

39555:
This draft includes an example of testing with `insta`. It requires `serde` support because `Debug` snapshots are unmanageably wide:
Screenshot 2024-11-08 at 9 55 28 PM
compared to yaml snapshots:

Screenshot 2024-11-08 at 10 12 17 PM

## Usage
- pass the parsed ast into `insta::assert_yaml_snapshot`
- run with `cargo insta review`
- accept the snapshot
- rerun with `cargo test`
-  if a newly created snapshot differs the test fails

## Cons
One inconvenience is that multiline fields in snap files are not formatted which requires always using `cargo insta` tool to pretty printing snapshots. This might related to https://github.com/mitsuhiko/insta/issues/372 I heven't found a good way to store the text input along with the ast snapshot. 

```
---
source: brush-parser/src/parser.rs
description: "\n#!/usr/bin/env bash\n\nfor f in A B C; do\n\n    # sdfsdf\n    echo \"${f@L}\" >&2\n\n   done\n\n"
---
...
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -316,9 +316,11 @@ dependencies = [
  "cached",
  "criterion",
  "indenter",
+ "insta",
  "peg",
  "pprof",
  "pretty_assertions",
+ "serde",
  "thiserror",
  "tracing",
  "utf8-chars",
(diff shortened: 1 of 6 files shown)


reubeno:
Thanks for plotting the course for us to use `insta`! I've cherry-picked the base of these changes and am moving forward with #545 for the rest.

PR #266, opened 2024-11-08
test(parser): snapshot testing with `expect_test`

39555:
[`expect_test`](https://github.com/rust-analyzer/expect-test) is minimalistic library from the rust-analyzer project that does not require a lot of learning. It can update both inline snapshots and snapshots from files by running `UPDATE_EXPECT=1 cargo test`.  TTo achieve comparable functionality to `insta` I've added `serde_yaml` for serialization although  I don't particularly like using serde here because it doesn't include type names. A similar limitation to insta is that the snapshot file doesn’t include the input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -315,10 +315,13 @@ dependencies = [
  "assert_matches",
  "cached",
  "criterion",
+ "expect-test",
  "indenter",
  "peg",
  "pprof",
  "pretty_assertions",
+ "serde",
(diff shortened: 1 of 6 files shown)


39555:
The second commit contains snapshots in [`ron`](https://github.com/ron-rs/ron) format. It looks better with the struct names

reubeno:
Thanks for helping us explore this path! We're going to proceed with `insta`.

PR #287, opened 2024-12-05
feat(exit): suspended jobs warning and sending kill signals

John-Bush14:
`exit` warns user if user has suspended jobs running and sends SIGHUP's on second try.

### DIFF
--- a/brush-core/src/builtins/exit.rs
+++ b/brush-core/src/builtins/exit.rs
@@ -1,5 +1,7 @@
 use clap::Parser;
 
+use std::io::Write;
+
 use crate::{builtins, commands};
 
 /// Exit the shell.
@@ -15,7 +17,15 @@ impl builtins::Command for ExitCommand {
         context: commands::ExecutionContext<'_>,
     ) -> Result<crate::builtins::ExitCode, crate::error::Error> {
         if !context.shell.jobs.jobs.is_empty() && context.shell.user_tried_exiting == 0 {
+            writeln!(context.stdout(), "brush: You have suspended jobs.")?;
+
             context.shell.user_tried_exiting = 2; // get's decreased this input too so next input will be 1
+
+            return Ok(builtins::ExitCode::Custom(1))
+        }
+
+        for job in &mut context.shell.jobs.jobs {
+            job.kill(Some(nix::sys::signal::SIGHUP))?;
         }
 
         let code_8bit: u8;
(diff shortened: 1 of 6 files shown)


reubeno:
Thanks for your contribution, @John-Bush14 ! It's great to see progress on this feature. Please don't hesitate to reach out here or in the Discussions area of the project if you have any questions along the process. I've also enabled the checks to start running, so you may find some new results that will need to be looked at.

We'll aim to review the changes in detail over the weekend or early next week.

brush-core/src/shell.rs
```diff
@@ -42,6 +42,9 @@ pub struct Shell {
     // Additional state
     /// The status of the last completed command.
     pub last_exit_status: u8,
+    
+    /// User tried exiting in the previous/current input if != 0.
+    pub user_tried_exiting: u8,
```
reubeno, line 47: Is there a reason for this to be an integer instead of a simple bool flag? Are there cases in which you think we'd need to track how many times the user tried exiting?

brush-core/src/jobs.rs
```diff
@@ -408,9 +408,9 @@ impl Job {
     }
 
     /// Kills the job.
-    pub fn kill(&mut self) -> Result<(), error::Error> {
+    pub fn kill(&mut self, signal: Option<nix::sys::signal::Signal>) -> Result<(), error::Error> {
```
reubeno, line 411: You may run into challenges exposing a `nix` type in this function--I recall it being used in non-Unix platforms (e.g., Windows). An alternative approach might be to define a separate abstract `enum` type for signals we can send to jobs. It might require adding some new functions under the `sys` module too.

brush-core/src/builtins/exit.rs
```diff
@@ -14,6 +16,18 @@ impl builtins::Command for ExitCommand {
         &self,
         context: commands::ExecutionContext<'_>,
     ) -> Result<crate::builtins::ExitCode, crate::error::Error> {
+        if !context.shell.jobs.jobs.is_empty() && context.shell.user_tried_exiting == 0 {
+            writeln!(context.stdout(), "brush: You have suspended jobs.")?;
```
reubeno, line 20: Minor code note: would be nice to avoid hard-coding the display name of the shell. If I recall correctly `Shell::shell_name` can be used.

brush-core/src/builtins/exit.rs
```diff
@@ -1,5 +1,7 @@
 use clap::Parser;
 
+use std::io::Write;
```
reubeno, line 3: Right now, a `Ctrl+D` based exit won't actually invoke the `exit` built-in. Do we need to make an additional change to make sure that path also works in the same way?

reubeno:
I'm going to close out this PR for now since there hasn't been any activity on it -- and the code in question has undergone quite a bit of change in the intervening time.

Should you or any other contributors be interested, I'd be happy to see this get reopened and updated.

PR #302, opened 2024-12-30
perf: experiment with immutable strings

reubeno:


reubeno:
@39555 I'm interested to get your take since you've been looking at and thinking about parsing, tokenizing, and strings. Independent of parser approach (I.e. winnow/nom, peg, etc.), what do you think about looking at one of the many "low-cost clone" / CoW  string crates? I encountered imstr but also see arcstr and a variety of others in similar category.

Given how much of the shell is string manipulation, I could see benefit but am concerned about how virally they spread in a code base and how much we'd get locked into them. 

This draft is just an experiment in the parser crate. I imagine we'd see greater benefit by going more widespread and figuring out how to better work with them, where we can use the non-thread-safe variants, etc.

39555:
@reubeno sorry for the long reply. Im currently out of my bandwidth. It is interesting concept! Is it basically an Arc<String>? Do you share something like variable names across the codebase?

PR #349, opened 2025-01-21
fix: correct semantics of = in non-extended test commands

### LINKED ISSUE #340: [ "foo" = "fo*" ] should not have a pattern on the right side of the =
```console
$ brush --version
brush 0.2.14 (git:dd1d39c)

$ brush -c '[ "foo" = "fo?" ] && echo match || echo unmatch'
match

$ bash -c '[ "foo" = "fo?" ] && echo match || echo unmatch'
unmatch
```

reubeno:
In a non-extended test command (e.g., `[` or `test`), the `=`, `==`, and `!=` operators are direct, case-sensitive string comparisons -- with no pattern matching. These changes correct that behavior.

(The incorrect behavior was inadvertently inherited from `[[`-style extended test commands when originally implemented.)

Resolves #340. Thanks to @ko1nksm for identifying the problem and filing a bug report on it!

### DIFF
--- a/brush-core/src/extendedtests.rs
+++ b/brush-core/src/extendedtests.rs
@@ -226,6 +226,26 @@ async fn apply_binary_predicate(
 
             Ok(matches)
         }
+        ast::BinaryPredicate::StringExactlyMatchesString => {
+            let left = expansion::basic_expand_word(shell, left).await?;
+            let right = expansion::basic_expand_word(shell, right).await?;
+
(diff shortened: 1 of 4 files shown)


reubeno:
We're hitting an internal compiler error in the aarch64 builds; I'm going to move forward and merge this PR to unblock it and will separately sort out what's happening with aarch64.

PR #395, opened 2025-03-05
feat: add parser selection

theoparis:
This PR adds support for selecting the parser implementation based on https://github.com/reubeno/brush/pull/283#issuecomment-2532562558. I was having trouble running the compat tests under NixOS though - not sure what I'm missing since it doesn't seem to give backtrace:

```
     Running tests/compat_tests.rs (target/debug/deps/brush_compat_tests-6c008899565a1736)
Error: No such file or directory (os error 2)
error: test failed, to rerun pass `--test brush-compat-tests`

Caused by:
  process didn't exit successfully: `/home/theo/src/brush/target/debug/deps/brush_compat_tests-6c008899565a1736` (exit status: 1)
```

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -178,6 +178,18 @@ pub struct CreateOptions {
     pub verbose: bool,
     /// Maximum function call depth.
     pub max_function_call_depth: Option<usize>,
+    /// The parser implementation to use.
+    pub parser: ParserImplementation,
+}
+
+/// Represents the parser implementation.
+#[derive(Debug, Clone, Default, clap::ValueEnum)]
+pub enum ParserImplementation {
+    #[default]
+    /// The peg parser implementation
+    Peg,
+    /// The winnow parser implementation
+    Winnow,
 }
 
 /// Represents an executing script.
--- a/brush-parser/src/lib.rs
+++ b/brush-parser/src/lib.rs
@@ -11,10 +11,12 @@ pub mod word;
 
 mod error;
 mod parser;
+mod parser2;
 mod tokenizer;
 
 pub use error::{ParseError, TestCommandParseError, WordParseError};
 pub use parser::{parse_tokens, Parser, ParserOptions, SourceInfo};
+pub use parser2::Parser as WinnowParser;
 pub use tokenizer::{
     tokenize_str, tokenize_str_with_options, uncached_tokenize_str, unquote_str, SourcePosition,
     Token, TokenLocation, TokenizerOptions,
--- a/brush-core/src/lib.rs
+++ b/brush-core/src/lib.rs
@@ -37,6 +37,6 @@ mod variables;
 pub use commands::ExecutionContext;
 pub use error::Error;
 pub use interp::{ExecutionParameters, ExecutionResult};
-pub use shell::{CreateOptions, Shell};
+pub use shell::{CreateOptions, ParserImplementation, Shell};
 pub use terminal::TerminalControl;
 pub use variables::{ShellValue, ShellVariable};
--- a/brush-parser/src/parser2.rs
+++ b/brush-parser/src/parser2.rs
@@ -1,0 +1,2 @@
+/// Implements parsing for shell programs using winnow.
+pub struct Parser {}
--- a/brush-shell/src/args.rs
+++ b/brush-shell/src/args.rs
@@ -1,3 +1,4 @@
+use brush_core::ParserImplementation;
 use clap::{builder::styling, Parser};
 use std::io::IsTerminal;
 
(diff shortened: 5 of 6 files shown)


brush-parser/src/lib.rs
```diff
@@ -11,10 +11,12 @@ pub mod word;
 
 mod error;
 mod parser;
+mod parser2;
 mod tokenizer;
 
 pub use error::{ParseError, TestCommandParseError, WordParseError};
 pub use parser::{parse_tokens, Parser, ParserOptions, SourceInfo};
+pub use parser2::Parser as WinnowParser;
```
reubeno: Is there a reason not to just name it `WinnowParser` in the first place? It would be great to get a more descriptive and understandable name than `parser2` when it gets checked in.

brush-core/src/shell.rs
```diff
@@ -178,6 +178,18 @@ pub struct CreateOptions {
     pub verbose: bool,
     /// Maximum function call depth.
     pub max_function_call_depth: Option<usize>,
+    /// The parser implementation to use.
+    pub parser: ParserImplementation,
+}
+
+/// Represents the parser implementation.
+#[derive(Debug, Clone, Default, clap::ValueEnum)]
+pub enum ParserImplementation {
```
reubeno, line 200: Since you're not directly exposing this type through `brush-shell` to clap, does this need to derive `clap::ValueEnum`?

reubeno (commented):
Thanks for diving into this! Once there's a way to select an alternate parser, then incremental progress can begin in earnest. Let's make sure that the new parser is clearly enough marked as experimental that folks should only get it if they know what they're doing.


(Aside: you may have seen this, but note that there's a handful of different peg grammars in `brush_parser`. `token_parser` in `parser.rs` is the *big one*. We have hope that a reimplementation of it will bring good yields. At the same time, some of the smaller ones may be easier to translate first. We're open to progress on any of them, but wanted to call this out.)

reubeno:
Oh, and @theoparis, regarding the issues you're seeing running the tests-- would you mind filing a separate issue to work through that? If possible, would be great if you could share any details on your environment, the specific command line you ran, anything you tried to figure out what was going on, etc.

brush-core/src/shell.rs
theoparis: No. I wasn't sure about keeping the two enums since they are pretty much identical - what would be better?

reubeno:
> Edit: It seems that because NixOS uses /nix/store for packages (e.g. /nix/store/--bash/bin/bash, it can't find the bash executable when the test harness passes the hard coded PATH variable. I can run the tests on a non-NixOS container image for now.

That would explain it. As you've likely found, the compat tests have a hard-coded set of default paths to use for the sanitized environment the shells-under-comparison run in:

https://github.com/reubeno/brush/blob/d9e421e42026ca293a510eead2ddbe29e1206918/brush-shell/tests/compat_tests.rs#L1013-L1016

There's similar defaulting logic in `shell.rs` as well:

https://github.com/reubeno/brush/blob/d9e421e42026ca293a510eead2ddbe29e1206918/brush-core/src/shell.rs#L541-L549

`brush` (and its compat tests) should instead really be calling into [`confstr()`](https://pubs.opengroup.org/onlinepubs/009604499/functions/confstr.html) to query the host platform's default paths. Now that I think about it, I recall #200 brought this topic up, although we hadn't made the connection with our compat tests at the time.

It's a bit of a detour from what you're working on, but if you're interested, I think it's worth the effort to at least get the compat tests running on NixOS. Once we get them running somewhat reliably, I'd be interested to learn if there's some way to extend our distro-specific integration tests to add a NixOS target of some kind. (Today our PR + CI tests run these suites on recent containers of Ubuntu, Debian, Fedora, and Arch.)

brush-core/src/shell.rs
reubeno: I think it's debatable, but you can probably just keep the "inner" one in `brush-core` and directly expose it via `brush-shell` without the wrapper.

reubeno:
Your latest changes look like they're going in the right direction, thanks for iterating. 

Do you have a plan for how the code will switch on the parser type to call the right one? Offhand I don't recall if we've done a good job consolidating all invocations into the parsers onto a single call site; any cleanup around that may be worth doing as part of this. (Also, do you have specific thoughts on whether the new enum will switch *all* the parsers or just the main program parser? If you search for `peg` or `grammar` you'll find the arithmetic one, the word parser, etc.)

I'm happy for us to iterate, but I see value in at least getting enough scaffolding wired up to see the new switch turn into different runtime behavior (even if that behavior is to fail with an error).

lu-zero:
Do we want to mix and match every sub-parser?

lu-zero:
I wonder if it can be pushed further, I might have some spare time to help with it.

lu-zero:
> Do you have a plan for how the code will switch on the parser type to call the right one? Offhand I don't recall if we've done a good job consolidating all invocations into the parsers onto a single call site; any cleanup around that may be worth doing as part of this. (Also, do you have specific thoughts on whether the new enum will switch _all_ the parsers or just the main program parser? If you search for `peg` or `grammar` you'll find the arithmetic one, the word parser, etc.)

Right now the subparsers a quick grep can find:
- arithmetic
- token_parser
- assignments
- prompt_parser
- readline_binding
- test_command
- expansion_parser

Do you want to have a struct to flip each of them independently?

reubeno:
I'm going to close this PR. Thanks for getting things started! It helped to move the conversation forward and we've now merged in enough scaffolding to start filling out the winnow-based parser.

PR #396, opened 2025-03-07
Expand contents of PS4 when tracing

dan4thewin:
This PR calls `expand_prompt_var()` on the contents of PS4 within `trace_command()`.  The addition necessitates `trace_command()` become `async` and calls to it use `await`.  A new test case checks expansion against an oracle.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -1510,7 +1510,7 @@ impl Shell {
         &mut self,
         command: S,
     ) -> Result<(), error::Error> {
-        let ps4 = self.as_mut().expand_prompt_var("PS4", "+ ").await?;
+        let ps4 = self.as_mut().expand_prompt_var("PS4", "").await?;
 
         let mut prefix = ps4.to_string();
 
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -1506,8 +1506,11 @@ impl Shell {
     /// # Arguments
     ///
     /// * `command` - The command to trace.
-    pub(crate) fn trace_command<S: AsRef<str>>(&self, command: S) -> Result<(), std::io::Error> {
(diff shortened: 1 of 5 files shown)


brush-core/src/shell.rs
```diff
@@ -1506,8 +1506,11 @@ impl Shell {
     /// # Arguments
     ///
     /// * `command` - The command to trace.
-    pub(crate) fn trace_command<S: AsRef<str>>(&self, command: S) -> Result<(), std::io::Error> {
-        let ps4 = self.get_env_str("PS4").unwrap_or_else(|| "".into());
+    pub(crate) async fn trace_command<S: AsRef<str>>(
+        &mut self,
+        command: S,
+    ) -> Result<(), error::Error> {
+        let ps4 = self.as_mut().expand_prompt_var("PS4", "+ ").await?;
```
reubeno: From some quick testing, it looks like `bash` defaults the trace prefix to `""` if `PS4` is *unset*, e.g.:

```console
$ echo $PS4
+
$ set -x; echo hi; set +x
+ echo hi
hi
+ set +x
$ unset PS4
$ set -x; echo hi; set +x
echo hi
hi
set +x
```

(Do you see the same?) That tells me that we should pass `""` as the 2nd arg here. If `PS4` was unset on entry into the shell, we should still benefit from the `Shell` initialization code path that gives it a default value of `"+ "` if unset.

reubeno (commented):
Thanks for your contribution! This is a good compat fix, and we appreciate you adding the new test case to cover it.

I've left one comment on the unset case, but the changes otherwise look good.

brush-core/src/shell.rs
dan4thewin: Good catch!  fixed.

reubeno (approved):

Issue #404, opened 2025-03-12
Can't parse empty array where parentheses are split across two lines


idealseal:
The following works in bash

```
FOO=(
)
```

reubeno:
Thanks for reporting this -- it's a good find that escaped our previous attention. I left a few comments on the PR, and we should be able to get this in relatively swiftly.

PR #405, opened 2025-03-12
fix: allow newline in empty array assignment

idealseal:
Closes: #404

### DIFF
--- a/brush-parser/src/parser.rs
+++ b/brush-parser/src/parser.rs
@@ -765,7 +765,7 @@ peg::parser! {
             }
 
         rule array_elements() -> Vec<&'input String> =
-            e:array_element()*
+             linebreak() e:array_element()* { e }
 
         rule array_element() -> &'input String =
             linebreak() [Token::Word(e, _)] linebreak() { e }
--- a/brush-shell/tests/cases/arrays.yaml
+++ b/brush-shell/tests/cases/arrays.yaml
@@ -320,3 +320,8 @@ cases:
       echo "\"\${myarray[*]}\": ${myarray[*]}"
       echo "\${myarray[@]}:" ${myarray[@]}
       echo "\${myarray[*]}:" ${myarray[*]}
+
+  - name: "Empty array with newline"
+    stdin: |
+      declare -a myarray=(
+      )


brush-parser/src/parser.rs
```diff
@@ -765,7 +765,7 @@ peg::parser! {
             }
 
         rule array_elements() -> Vec<&'input String> =
-            e:array_element()*
+             linebreak()? e:array_element()* { e }
```
reubeno: Two quick questions:

1. `linebreak` is already defined as `newline()*`, so is the `?` necessary?
2. Could you confirm that we don't need a `linebreak()` after the `array_element()*` for trailing newlines before the end parenthesis?

(Given the above, it occurs to me that `optional_linebreak` is a more verbose but probably more accurate name for the definition of the `linebreak` rule.)

reubeno (commented):
@idealseal -- thanks for finding this issue **and** for then taking the time to contribute a fix!

I left 2 quick inline questions regarding the code change.

Beyond that, it would be great if you could add any relevant test cases to our compat test suite for the buggy behavior you found; that will ensure this _stays_ fixed over time. Any new cases you add are probably most appropriate to add to [arrays.yaml](https://github.com/reubeno/brush/blob/main/brush-shell/tests/cases/arrays.yaml).

(There's some very minimal docs on the test .yaml format [here](https://github.com/reubeno/brush/blob/main/docs/reference/integration-testing.md) but it should hopefully be straightforward looking at the examples. The PR checks, which I'll unblock, will run the tests; you can also run the tests yourself by following [these steps](https://github.com/reubeno/brush/blob/main/docs/how-to/run-tests.md).)

brush-parser/src/parser.rs
idealseal: > linebreak is already defined as newline()*, so is the ? necessary?

No, I just found it confusing that a `linebreak`  parser is itself optional.

> 2. Could you confirm that we don't need a linebreak() after the array_element()* for trailing newlines before the end parenthesis?

If the array is empty, the first `linebreak`  will cover all newlines. If there is at least 1 element, the trailing `linebreak` will consume all newlines after the last element.

reubeno:
Thanks for the updates; I'll merge when the test run passes.

Issue #408, opened 2025-03-12
Quotes in comments inside of sub shells are not ignored and generate an error if not terminated.


idealseal:
The following is valid in bash:

```bash
var=$(
  # I'm
  echo "Batman!"
)
echo $var
```

brush:

```
ERROR test.bash: unterminated single quote at line 2 col 6 (detected near line 6 column 1)
```

Applies to double quotes as well.

reubeno:
This is another good find.

It looks like the problem is in our tokenizer. In this case, the command substitution syntax (i.e., `$(...)`) results in a recursive call to `self.next_token_until(Some(')'))`. Because of the logic in the following code snippet, we see that `terminating_char` is `Some(')')` and consume the `'#'` char as a normal char and not a comment character--never getting to the `else` case immediately following it.

https://github.com/reubeno/brush/blob/63bea4f589675b16e86feb2bd301d1a433b78b99/brush-parser/src/tokenizer.rs#L1094-L1103

We have plans to overhaul our tokenizer down the road, but we should absolutely fix this now. We need to make sure that we consider `#` and end-of-line comment character when it's the beginning of a token in a command substitution body, but *not* when we encounter it inside a curly-brace expansion expression (e.g., `${...}`).

A quick fix might be replacing the `terminating_char.is_some()` check with `matches!(terminating_char, Some(')')`, although that's capturing a bit more knowledge of the caller's context than might be desirable and we'll need to do some testing to confirm if that results in acceptable behavior when inside an arithmetic expression (e.g. `$((...))`).

reubeno:
@idealseal -- is this an issue you're interested in working on contributing a fix for? If so, we're happy to review / provide pointers / discuss if that would be helpful. And if not, that's totally fine too! We'll otherwise make sure to get it assigned to someone to work on.

idealseal:
I can do that.

idealseal:
The proposed solution didn't work. Also it breaks a couple of existing tests. I'm currently struggling with understanding the tokenizer code so I'll need a bit more time to get a proper fix.

reubeno:
Thanks for diving into this. Admittedly, the tokenizer got more complicated than we had originally intended. We have longer-term aspirations to revisit it, but you're welcome to propose any improvements you see fit.

And if you could use a sounding board or have questions, don't hesitate to post here.

PR #418, opened 2025-03-20
fix: parse '#' as char only if inside a variable expansion

idealseal:
Only in arithmetic expressions '#' is a valid token, in other contexts it is treated as a comment, and all following characters will be ignored by the tokenizer.

Closes: #408

### DIFF
--- a/brush-parser/src/tokenizer.rs
+++ b/brush-parser/src/tokenizer.rs
@@ -1091,11 +1091,11 @@ impl<'a, R: ?Sized + std::io::BufRead> Tokenizer<'a, R> {
                 self.consume_char()?;
             }
             //
-            // N.B. We need to remember if we were recursively called, say in a command
-            // substitution; in that case we won't think a token was started but... we'd
-            // be wrong.
+            // N.B. We need to remember if we were recursively called in a variable
+            // expansion expression; in that case we won't think a token was started but...
+            // we'd be wrong.
             else if !state.token_is_operator
-                && (state.started_token() || terminating_char.is_some())
+                && (state.started_token() || matches!(terminating_char, Some('}')))
             {
                 self.consume_char()?;
                 state.append_char(c);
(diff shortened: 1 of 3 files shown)


brush-shell/tests/cases/subshell.yaml
```diff
@@ -0,0 +1,33 @@
+name: "subshell"
+cases:
+  - name: "Ignore single quote in comment in subshell"
+    stdin: |
```
reubeno: Thanks for adding these tests -- this will go a long way toward making sure we don't regress the fix.

What do you think about adding (or verify we already have) cases to ensure correct tokenization when `#` is inside other similar constructs:

* Command substitutions: `$(...)` (what you already covered)
* Subshell commands: `(...)`
* Backtick-style command substitutions

We should probably also cover nested scenarios,  say, `$#` showing up inside a command substitution (e.g., `$(echo $#)`) or cases with one of the above constructs containing a `#` but also nested in another such construct? We don't need to go *crazy* with the tests, but it would be good to make sure we cover the basics.

brush-shell/tests/cases/subshell.yaml
```diff
@@ -0,0 +1,33 @@
+name: "subshell"
```
reubeno: _Minor comment:_ it's true that `$(...)` gets run in a subshell, but we've found it useful to specifically refer to it as a "command substitution" (following the [POSIX docs](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03)) to distinguish it from other constructs that also end up executing logic in subshells (e.g., `(...)`).

What do you think about renaming the file to `command_substitution.yaml`? We already have some command substitution tests in `word_expansion.yaml` but that file has gotten large and unwieldy. I like the idea of factoring out smaller `.yaml` files and, over time, breaking it up. (Doesn't need to be done in this PR.)

reubeno (commented):
Thanks for contributing a fix for the issue! I have a few comments/questions about testing, but the fix seems like a reasonable, targeted one to go with.

reubeno (approved):
Thanks for making those changes, @idealseal. LGTM

PR #437, opened 2025-04-22
chore(Shell): use relaxed typing for string input

Elsie19:
This should help by moving the need to (possibly) convert to a `String` from the call-site to the function itself, which should help with the ergonomics of calling these methods.


reubeno:
LGTM.

(Thanks for diving in -- I'm sure you'll encounter similar sources of friction; we're happy to take these changes to make it better bit by bit.)

reubeno (approved):

PR #438, opened 2025-04-23
feat(Shell): add `get_env_var` method for more generic variable returns

### LINKED ISSUE #436: `brush_core::Shell` doesn't expose a high-level function to retrieve array variables
I see that there is a way to retrieve a variable (https://docs.rs/brush-core/latest/brush_core/struct.Shell.html#method.get_env_str), but if I try that on an array, I only get `${array[0]}`:

```rust
use brush_core::{CreateOptions, ExecutionParameters, Shell};

#[tokio::main]
async fn main() {
    let text = "hello=(1 2 3 4);echo ${hello[2]}";
    let options = CreateOptions::default();
    let mut shell = Shell::new(&options).await.unwrap();
    shell
        .run_string(text.to_string(), &ExecutionParameters::default())
        .await
        .unwrap();
    let var = shell.get_env_str("hello");
    dbg!(var);
}
```

Output:

```
3
[src/main.rs:13:5] var = Some(
    "1",
)
```

Is there some way I can get the array elements back instead of just the first one?

Elsie19:
### Considerations
* ~~I do use a `.clone()` here, which is probably not ideal, so I am open to suggestions on figuring out how to remove it.~~
* We should consider if we should deprecate `get_env_str` or keep it around as a convenience method.

Closes #436.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -1259,6 +1259,15 @@ impl Shell {
         self.env.get_str(name, self)
     }
 
+    /// Tries to retrieve a variable from the shell's environment.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The name of the variable to retrieve.
+    pub fn get_env_var(&self, name: &str) -> Option<ShellVariable> {
+        self.env.get(name).map(|(_, var)| var.clone())
+    }
+
     /// Returns the current value of the IFS variable, or the default value if it is not set.
     pub(crate) fn get_ifs(&self) -> Cow<'_, str> {
         self.get_env_str("IFS").unwrap_or_else(|| " \t\n".into())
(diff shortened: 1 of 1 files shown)


reubeno:
Thanks for adding this; I agree with returning `Option<&ShellVariable>`.

As for deprecating `get_env_str`, here are my thoughts:

* For the scope of this PR, I think we should leave `get_env_str` as is.
* There are a handful of places in the code base where `get_env_str` feels like the right sort of fit -- they just want to get a variable and coerce it to a string of some kind without being too thoughtful about it. If we did want to consider removing the function from `Shell` in the future, I'd at least consider adding it to some kind of `ShellExt` helper that provides it as a convenience.

For now, let's move forward with this change.

reubeno (approved):

PR #439, opened 2025-04-23
feat(apply_unary_predicate_to_str): implement `-R` operand

Elsie19:
Adds:

* `-R`
* `-ef`
* `-nt`
* `-ot`

### DIFF
--- a/brush-core/src/extendedtests.rs
+++ b/brush-core/src/extendedtests.rs
@@ -300,7 +300,25 @@ async fn apply_binary_predicate(
             Ok(l_md.st_dev() == r_md.st_dev() && l_md.st_ino() == r_md.st_ino())
         }
         ast::BinaryPredicate::LeftFileIsNewerOrExistsWhenRightDoesNot => {
-            error::unimp("extended test binary predicate LeftFileIsNewerOrExistsWhenRightDoesNot")
+            let left = expansion::basic_expand_word(shell, params, left).await?;
+            let right = expansion::basic_expand_word(shell, params, right).await?;
+
+            if shell.options.print_commands_and_arguments {
+                shell
+                    .trace_command(std::format!("[[ {left} {op} {right} ]]"))
+                    .await?;
+            }
+
+            let (l_path, r_path) = (
+                shell.get_absolute_path(Path::new(&left)),
+                shell.get_absolute_path(Path::new(&right)),
+            );
+
+            match (l_path.metadata(), r_path.metadata()) {
+                (Ok(m1), Ok(m2)) => Ok(m1.modified()? > m2.modified()?),
+                (Ok(_), Err(_)) => Ok(true),
+                _ => Ok(false),
+            }
         }
         ast::BinaryPredicate::LeftFileIsOlderOrDoesNotExistWhenRightDoes => error::unimp(
             "extended test binary predicate LeftFileIsOlderOrDoesNotExistWhenRightDoes",
--- a/brush-core/src/extendedtests.rs
+++ b/brush-core/src/extendedtests.rs
@@ -182,7 +182,10 @@ pub(crate) fn apply_unary_predicate_to_str(
         }
         ast::UnaryPredicate::ShellVariableIsSetAndAssigned => Ok(shell.env.is_set(operand)),
         ast::UnaryPredicate::ShellVariableIsSetAndNameRef => {
-            error::unimp("unary extended test predicate: ShellVariableIsSetAndNameRef")
+            match (shell.env.is_set(operand), shell.env.get(operand)) {
+                (false, _) | (_, None) => Ok(false),
+                (_, Some(reffed)) => Ok(reffed.1.is_treated_as_nameref()),
+            }
         }
     }
 }
--- a/brush-core/src/extendedtests.rs
+++ b/brush-core/src/extendedtests.rs
@@ -1,5 +1,5 @@
 use brush_parser::ast;
-use std::path::Path;
+use std::{os::linux::fs::MetadataExt as _, path::Path};
 
 use crate::{
     arithmetic, env, error, escape, expansion, namedoptions, patterns,
@@ -181,12 +181,10 @@ pub(crate) fn apply_unary_predicate_to_str(
             }
         }
         ast::UnaryPredicate::ShellVariableIsSetAndAssigned => Ok(shell.env.is_set(operand)),
-        ast::UnaryPredicate::ShellVariableIsSetAndNameRef => {
-            match (shell.env.is_set(operand), shell.env.get(operand)) {
-                (false, _) | (_, None) => Ok(false),
-                (_, Some(reffed)) => Ok(reffed.1.is_treated_as_nameref()),
-            }
-        }
+        ast::UnaryPredicate::ShellVariableIsSetAndNameRef => match shell.env.get(operand) {
+            Some((_, reffed)) => Ok(reffed.is_treated_as_nameref()),
+            None => Ok(false),
+        },
     }
 }
 
@@ -278,7 +276,28 @@ async fn apply_binary_predicate(
             Ok(s.contains(substring.as_str()))
         }
         ast::BinaryPredicate::FilesReferToSameDeviceAndInodeNumbers => {
-            error::unimp("extended test binary predicate FilesReferToSameDeviceAndInodeNumbers")
+            let left = expansion::basic_expand_word(shell, params, left).await?;
+            let right = expansion::basic_expand_word(shell, params, right).await?;
+
+            if shell.options.print_commands_and_arguments {
+                shell
+                    .trace_command(std::format!("[[ {left} {op} {right} ]]"))
+                    .await?;
+            }
+
+            let (l_path, r_path) = (
+                shell.get_absolute_path(Path::new(&left)),
+                shell.get_absolute_path(Path::new(&right)),
+            );
+
+            if !l_path.exists() || !r_path.exists() {
+                return Ok(false);
+            }
+
+            let l_md = l_path.metadata()?;
+            let r_md = r_path.metadata()?;
+
+            Ok(l_md.st_dev() == r_md.st_dev() && l_md.st_ino() == r_md.st_ino())
         }
         ast::BinaryPredicate::LeftFileIsNewerOrExistsWhenRightDoesNot => {
             error::unimp("extended test binary predicate LeftFileIsNewerOrExistsWhenRightDoesNot")
--- a/brush-core/src/extendedtests.rs
+++ b/brush-core/src/extendedtests.rs
@@ -181,12 +181,10 @@ pub(crate) fn apply_unary_predicate_to_str(
             }
         }
         ast::UnaryPredicate::ShellVariableIsSetAndAssigned => Ok(shell.env.is_set(operand)),
-        ast::UnaryPredicate::ShellVariableIsSetAndNameRef => {
-            match (shell.env.is_set(operand), shell.env.get(operand)) {
-                (false, _) | (_, None) => Ok(false),
-                (_, Some((_, reffed))) => Ok(reffed.is_treated_as_nameref()),
-            }
-        }
+        ast::UnaryPredicate::ShellVariableIsSetAndNameRef => match shell.env.get(operand) {
+            Some((_, reffed)) => Ok(reffed.value().is_set() && reffed.is_treated_as_nameref()),
+            None => Ok(false),
+        },
     }
 }
 
--- a/brush-core/src/extendedtests.rs
+++ b/brush-core/src/extendedtests.rs
@@ -1,5 +1,5 @@
 use brush_parser::ast;
-use std::{os::linux::fs::MetadataExt as _, path::Path};
+use std::path::Path;
 
 use crate::{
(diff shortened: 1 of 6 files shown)


Elsie19:
I'm going to be working on `[[ ]]` a bit more which is why this is a draft.

Elsie19:
Unfortunately I introduced a `.except()` in here which I'm sure you don't want, and I'm unsure of how to move forward.

Elsie19:
I think I am confusing myself more 😄... I'll just wait for some input on how to move forward.

brush-core/src/extendedtests.rs
```diff
@@ -181,9 +181,10 @@ pub(crate) fn apply_unary_predicate_to_str(
             }
         }
         ast::UnaryPredicate::ShellVariableIsSetAndAssigned => Ok(shell.env.is_set(operand)),
-        ast::UnaryPredicate::ShellVariableIsSetAndNameRef => {
-            error::unimp("unary extended test predicate: ShellVariableIsSetAndNameRef")
-        }
+        ast::UnaryPredicate::ShellVariableIsSetAndNameRef => match shell.env.get(operand) {
```
reubeno, line 184: I recall that the docs for `[[` specifically mention that `-R` checks if the named variable is a nameref *and* "set". Even if `.is_treated_as_nameref()` returns `true`, the answer might need to be `false` if the variable is "unset" (i.e., declared, but not initialized with a value), e.g.:

```console
$ declare -n somevar # Declare the variable and mark it nameref but don't initialize it with a value
$ [[ -R somevar ]] && echo "Yes!"
```

I'd recommend extending the [`extended_tests` suite of our compatibility integration tests](https://github.com/reubeno/brush/blob/main/brush-shell/tests/cases/extended_tests.yaml) to add various cases to confirm equivalent behavior with `bash`. We'll want to see that for merging this feature anyhow, and I've found that helpful in confirming the right semantics.

(There's extremely sparse docs [here](https://github.com/reubeno/brush/blob/main/docs/how-to/run-tests.md) on how to run the tests, and [here](https://github.com/reubeno/brush/blob/main/docs/reference/integration-testing.md) on the idea of the compat tests.

brush-core/src/extendedtests.rs
```diff
@@ -275,14 +276,71 @@ async fn apply_binary_predicate(
...
+            let (l_path, r_path) = (
+                shell.get_absolute_path(Path::new(&left)),
+                shell.get_absolute_path(Path::new(&right)),
+            );
+
+            if !l_path.exists() || !r_path.exists() {
+                return Ok(false);
+            }
+
+            let l_md = l_path.metadata()?;
```
reubeno: I don't recall if we handle this correctly elsewhere, but `bash` looks like it quietly ignores any access errors, say a file that might exist but is in a dir that the user doesn't have access to, e.g.:

image

Given this, we can probably skip the calls to `.exists()` above, and just always return false if we get any error on either file.

brush-core/src/sys/unix/fs.rs
```diff
@@ -60,6 +60,14 @@ impl crate::sys::fs::PathExt for Path {
         let file_mode = try_get_file_mode(self);
         file_mode.is_some_and(|mode| mode & S_ISVTX != 0)
     }
+
+    fn ino(&self) -> u64 {
```
reubeno: I think there's a few possible directions that we could take. I'll lay out a few options and give some thoughts, but am interested to hear your thoughts too:

1. Add to `PathExt` a function that does the comparison itself, e.g.:

    `fn is_same_physical_file_as(other: &Path) -> Result<bool, error::Error>`
    
    A variant of this approach might be to have it just return a `bool`; that's a bit more in line with the other functions here, but perhaps it would be better to allow the caller to decide how to handle errors.2

2. Take a dependency on something like the [same-file](https://crates.io/crates/same-file) crate, which tries to tackle the same problem?

3. Go with something like the approach you've started, and add to `PathExt` a single function that retrieves the device and inode numbers for the identified file, e.g.:

    `get_device_and_inode(&self) -> Result<(u64, u64), error::Error>`
    
    Retrieving both in a single call would eliminate a 2nd underlying call (and filesystem access) to `.metadata()`. A variant of this approach might be to have it be more of a `try_get_device_and_inode()` call that returns an `Option<(u64, u64)>`.
    
4. Leave `PathExt` alone and instead add some function that takes an already-retrieved `std::fs::Metadata` and tries to extract the device and inode numbers. This could be done in a regular-old function or in some new custom `Ext` struct that we add, e.g.:

    `try_get_device_and_inode(metadata: &std::fs::Metadata) -> Option<(u64, u64)>`
    
From a portability (and abstraction) perspective, I lean toward 1 (or possibly 2, with some more investigation as to its dependencies, how well maintained it is, etc.). I'm also generally open to 3 and 4.

What's your perspective?

brush-core/src/extendedtests.rs
Elsie19: I think a better idea might be to replace `.exists()` with `.readable()`, because that will cover the existence case and be more in line with bash access errors.

brush-core/src/sys/unix/fs.rs
Elsie19: I think the primary issue I had when trying to read the code was that a lot of implementation vs declaration was split apart, where I think it might be a bit better to put the different implementations together, so instead of having a `src/sys/fs.rs` and `src/sys/unix/fs.rs` for instance, you could simply put them in one file, and have `cfg` guards for each platform. It would look a lot nicer and be easier to reason about, although it is your project so I won't be picky :)

Elsie19: > I think there's a few possible directions that we could take. I'll lay out a few options and give some thoughts, but am interested to hear your thoughts too:
> 
> 1. Add to `PathExt` a function that does the comparison itself, e.g.:
>    `fn is_same_physical_file_as(other: &Path) -> Result<bool, error::Error>`
>    A variant of this approach might be to have it just return a `bool`; that's a bit more in line with the other functions here, but perhaps it would be better to allow the caller to decide how to handle errors.2
> 2. Take a dependency on something like the [same-file](https://crates.io/crates/same-file) crate, which tries to tackle the same problem?
> 3. Go with something like the approach you've started, and add to `PathExt` a single function that retrieves the device and inode numbers for the identified file, e.g.:
>    `get_device_and_inode(&self) -> Result<(u64, u64), error::Error>`
>    Retrieving both in a single call would eliminate a 2nd underlying call (and filesystem access) to `.metadata()`. A variant of this approach might be to have it be more of a `try_get_device_and_inode()` call that returns an `Option<(u64, u64)>`.
> 4. Leave `PathExt` alone and instead add some function that takes an already-retrieved `std::fs::Metadata` and tries to extract the device and inode numbers. This could be done in a regular-old function or in some new custom `Ext` struct that we add, e.g.:
>    `try_get_device_and_inode(metadata: &std::fs::Metadata) -> Option<(u64, u64)>`
> 
> From a portability (and abstraction) perspective, I lean toward 1 (or possibly 2, with some more investigation as to its dependencies, how well maintained it is, etc.). I'm also generally open to 3 and 4.
> 
> What's your perspective?

As to this question, every platform except Windows has `.dev()` and `.ino()` like functions, so what how would you feel about leaving them as `unimp` on windows but implemented for other platforms?

We could add a single function that basically does:

```rs
fn dev(&self) -> u64 {
	if cfg!(unix) {
		use std::os::unix::fs::MetadataExt;
		return self.st_dev();
	} else if cfg!(target_family = "wasm") {
		use std::os::wasi::fs::MetadataExt;
		return self.dev();
	} else {
		error:unimp("bla");
	}
}
```

or something like that.

reubeno: Thanks for the thoughtful input and iteration on this.

Interestingly, I started out with the platform-specific code interleaved in the same `.rs` file. I ended up breaking it out the way it is now after seeing the [crosvm guide on platform-specific code](https://crosvm.dev/book/contributing/style_guide_platform_specific_code.html) and seeing how the Rust standard library structures its `sys` modules, e.g.: [for std::fs](https://github.com/rust-lang/rust/tree/master/library/std/src/sys/fs). I also understand it may be something of a developer-specific preference. I'm open to this being brought back into discussion, but would prefer to do that in a separate issue and leave this PR adhering to the current structure.

The reason my proposals 3 and 4 above merged `dev()` and `ino()` retrieval into a single call was to avoid a second unnecessary underlying call to `.metadata()`, which likely entails filesystem lookup. It also avoids the admittedly _extremely unlikely_ possibility of torn state if the file is updated between the two calls (vs. the guaranteed consistent snapshot data from the single metadata lookup).

Regarding Windows support... I'm okay with us leaving the Windows code path unimplemented for now, but I did look into it a bit more. It looks like there is some kind of counterpart of st_dev/st_ino, with the option of retrieving a possibly-64-bit volume ID and 128-bit volume-relative file ID. If and when we decide to get this working for Windows correctly, I'd suggest that we consider shifting to option 1 above and hide the differences behind an abstraction that can have a platform-specific implementation.

With the above said, I'd propose option 3 for now, which I believe is close to your suggestion.

Elsie19: > Interestingly, I started out with the platform-specific code interleaved in the same `.rs` file. I ended up breaking it out the way it is now after seeing the [crosvm guide on platform-specific code](https://crosvm.dev/book/contributing/style_guide_platform_specific_code.html) and seeing how the Rust standard library structures its `sys` modules, e.g.: [for std::fs](https://github.com/rust-lang/rust/tree/master/library/std/src/sys/fs?rgh-link-date=2025-04-24T04%3A06%3A16Z). I also understand it may be something of a developer-specific preference. I'm open to this being brought back into discussion, but would prefer to do that in a separate issue and leave this PR adhering to the current structure.

I mean if rust does it this way, who am I to judge? I brought it up because it was/is my preference, but I would rather not overhaul your codebase just to fit what I want, and if this is the more idiomatic version, I'll stick to it.

> With the above said, I'd propose option 3 for now, which I believe is close to your suggestion.

I will continue to work on option 3 then!

reubeno: Sounds great, thanks!

brush-shell/tests/cases/extended_tests.yaml
```diff
@@ -110,6 +110,15 @@ cases:
       [[ -w test-file ]] && echo "-w correctly identified writable file"
       [[ -w link ]] && echo "-w correctly identified link to writable file"
 
+  - name: "Variable set and nameref tests"
```
reubeno, line 113: Thanks for adding this test. Could you also add some basic coverage for the other `[[ ... ]]` syntax you implemented?

brush-core/src/extendedtests.rs
```diff
@@ -182,7 +182,10 @@ pub(crate) fn apply_unary_predicate_to_str(
         }
         ast::UnaryPredicate::ShellVariableIsSetAndAssigned => Ok(shell.env.is_set(operand)),
         ast::UnaryPredicate::ShellVariableIsSetAndNameRef => {
-            error::unimp("unary extended test predicate: ShellVariableIsSetAndNameRef")
+            match (shell.env.is_set(operand), shell.env.get(operand)) {
```
reubeno: This should fix the behavior, but it may incur 2 environmental lookups (which might require going through multiple scopes to find the match).

It would be cleaner if we could sort this out with just a single call to `shell.env.get()`. Underneath, the `is_set()` function just checks if the variable exists and doesn't match `ShellValue::Unset(_)`. Either we could do that inline here, or if we don't already have an `is_set()` helper on `ShellValue`, we could just add one. It's a common enough thing that we need to do.

reubeno (changes requested):
I left a couple of comments, but the changes are looking pretty good. Thanks for addressing the earlier feedback.

brush-core/src/extendedtests.rs
Elsie19: > This should fix the behavior, but it may incur 2 environmental lookups (which might require going through multiple scopes to find the match).
> 
> It would be cleaner if we could sort this out with just a single call to `shell.env.get()`. Underneath, the `is_set()` function just checks if the variable exists and doesn't match `ShellValue::Unset(_)`. Either we could do that inline here, or if we don't already have an `is_set()` helper on `ShellValue`, we could just add one. It's a common enough thing that we need to do.

https://github.com/reubeno/brush/pull/439#discussion_r2056680824

I added the `is_set` because I thought you said it had to explicitly be set?

reubeno: To clarify, the change looks functionally correct. My last comment was focused on avoiding a double lookup. 

The `is_set()` call's implementation does an env lookup + check on the retrieved value (if the `ShellValue` object exists), then discards the retrieved value. The `get()` call then performs a second lookup.

My proposal is basically to do exactly what `is_set()` does internally (i.e. still check for existing-but-unset case) but not discard the retrieved value.

Is that clearer?

reubeno (approved):
LGTM; thanks for iterating on it. Once the tests pass, we can get this merged.

Elsie19:
Before merging, if you want, I can do some squash commits to make the merge look a lot less like *that* :)

Elsie19:
I think it's ready now!

PR #444, opened 2025-04-25
chore: remove `uzers` dependency

Elsie19:
You can take this PR or not, it depends on how important `uzers` is for possible future development, but I noticed that most of the functions that `uzers` provide just wrap over `nix`, which we already have.

If you want the simplicity of the `uzers` API though, I'll just close this PR :)


Elsie19:
Oh I see why it's here, because of MacOS.

PR #446, opened 2025-04-25
feat(parser): add gettext enabled quotes

### LINKED ISSUE #442: Gettext enabled strings do not split correctly
`test.sh`:
```bash
# Fails
echo $"Hello"
# Works
echo $'Hello'
```

```
$ ./target/debug/brush test.sh
$Hello
Hello
```

Note that I'm not asking for gettext support, just that it should print `Hello` instead of `$Hello`.

Elsie19:
> [!Note]
> This doesn't translate gettext strings, but it will allow it to be used with it's base string (for now).

---

Closes #442.


brush-parser/src/word.rs
```diff
@@ -623,6 +625,7 @@ peg::parser! {
             double_quoted_text()
 
         rule unquoted_text<T>(stop_condition: rule<T>, in_command: bool) -> WordPiece =
+            s:gettext_double_quoted_sequence() { WordPiece::GettextDoubleQuotedSequence(s) } /
```
reubeno: Minor note -- since we expect to see this less frequently than other string forms, we should probably move it further down in the rule (say, after or before the ANSI C escape string case).

It would also be interesting to see if any of our perf tests would notice the difference.

reubeno:
The change looks right.

Do you know if these quotations can also be embedded inside normal double-quotes? (I'm assuming not?)  I'm trying to reason whether the tokenizer needs to be enlightened as well; hopefully not.

Elsie19:
`$""` should behave exactly like `""` with the exception that there is a dollar sign prefixing it, so all double quoting rules apply to this as well.

reubeno (commented):
What do you think about moving forward with this? I realized I had 1 pending comment that I didn't managed to send earlier (sorry!), but other than that it generally looks good to me.

I could see value in adding a basic compat test case too, just to make sure that the original issue you found got resolved by this.

Elsie19:
> What do you think about moving forward with this? I realized I had 1 pending comment that I didn't managed to send earlier (sorry!), but other than that it generally looks good to me.

That would be good. I haven't figured out how to strip the beginning `$`, but I haven't had time to look into it more.

reubeno:
Thanks for taking another look at this over the weekend, @Elsie19. After you shared your findings, I took a stab at this -- I think there was an issue with rule ordering, as we were surmising. I rebased your changes against latest `main` (in a separate branch) and added some changes in [this commit](https://github.com/reubeno/brush/commit/c9a9a597923571af046b16f4e9bf7c5c727b7216).

It now seems to work:

```console
brush$ echo $"some text from $0"
some text from target/debug/brush
```

If this works for you, feel free to incorporate my change into your PR branch.

reubeno (approved):

PR #447, opened 2025-04-25
feat(Shell): add methods to add environmental variables and builtins

Elsie19:
This will be specifically useful for my cases, where I will be inserting variables and builtins (https://github.com/pacstall/pacstall/wiki/101.1-Variables#built-in-variables) into the environment before I source/run a script, but I'm sure a lot of people wanting to interface with brush will also find it useful.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -1268,6 +1268,30 @@ impl Shell {
         self.env.get(name).map(|(_, var)| var)
     }
 
+    /// Tries to set a global variable in the shell's environment.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The name of the variable to add.
+    /// * `var` - The variable contents to add.
+    pub fn set_env_global(&mut self, name: &str, var: ShellVariable) -> Result<(), error::Error> {
+        self.env.set_global(name, var)
+    }
+
+    /// Register a builtin to the shell's environment.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The in-shell name of the builtin.
+    /// * `registration` - The registration handle for the builtin.
+    pub fn register_builtin<S: Into<String>>(
+        &mut self,
+        name: S,
+        registration: builtins::Registration,
+    ) {
+        self.builtins.insert(name.into(), registration);
+    }
+
     /// Returns the current value of the IFS variable, or the default value if it is not set.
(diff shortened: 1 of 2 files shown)


brush-core/src/shell.rs
```diff
@@ -1268,6 +1268,30 @@ impl Shell {
...
     }
 
+    /// Tries to add a global variable to the shell's environment.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The name of the variable to add.
+    /// * `var` - The variable contents to add.
+    pub fn add_env_var(&mut self, name: &str, var: ShellVariable) -> Result<(), error::Error> {
+        self.env.add(name, var, env::EnvironmentScope::Global)
```
reubeno: `ShellEnvironment` already has a `set_global()` that's probably the right function to call here.

I'd also prefer that we name this function `set_env_global()` or something close to that. I'm not sure if you or other callers will ever have need to mutate variables that aren't in the global scope, but I think it best to clarify that in the name up-front.

brush-core/src/shell.rs
```diff
@@ -1268,6 +1268,30 @@ impl Shell {
...
+    }
+
+    /// Register a builtin to the shell's environment.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The in-shell name of the builtin.
+    /// * `registration` - The registration handle for the builtin.
+    pub fn register_builtin<S: Into<String>>(
```
reubeno, line 1287: Thanks for adding this; while you're at it, would you mind updating the registration for `brushctl` in `brush-shell` to use this new helper?

https://github.com/reubeno/brush/blob/651c185e4446e3421ba22253ba068262ee058487/brush-shell/src/brushctl.rs#L6-L11

brush-core/src/shell.rs
```diff
@@ -1268,6 +1268,30 @@ impl Shell {
...
+        &mut self,
+        name: S,
+        registration: builtins::Registration,
+    ) {
+        self.builtins.insert(name.into(), registration.special());
```
reubeno: I hadn't noticed it before, but I believe the tests are failing because of the `.special()` call. Unless you were intentionally looking for a way to allow registering [*special* built-ins](https://www.gnu.org/software/bash/manual/html_node/Special-Builtins.html), then this should just pass through `registration` without marking it special.

reubeno (approved):

PR #448, opened 2025-04-26
ref(Shell): turn `&Path` parameters into `AsRef<Path>`

Elsie19:
This should loosen the constraints on passing parameters to these functions so that:

```rs
shell
    .run_script(
        &PathBuf::from("/home/elsie/Projects/libpacstall/examples/test.sh"),
        vec!["foo", "bar"].into_iter(),
    )
    .await?;
```

Can look like:

```rs
shell
    .run_script(
        "/home/elsie/Projects/libpacstall/examples/test.sh",
        vec!["foo", "bar"].into_iter(),
    )
    .await?;
```

Taking https://doc.rust-lang.org/std/fs/fn.read_to_string.html as a reference.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -749,9 +749,10 @@ impl Shell {
 
     async fn source_if_exists(
         &mut self,
-        path: &Path,
+        path: impl AsRef<Path>,
         params: &ExecutionParameters,
     ) -> Result<bool, error::Error> {
+        let path = path.as_ref();
         if path.exists() {
             self.source_script(path, std::iter::empty::<String>(), params)
                 .await?;
@@ -769,13 +770,13 @@ impl Shell {
     /// * `path` - The path to the file to source.
     /// * `args` - The arguments to pass to the script as positional parameters.
     /// * `params` - Execution parameters.
-    pub async fn source_script<S: AsRef<str>, I: Iterator<Item = S>>(
+    pub async fn source_script<S: AsRef<str>, P: AsRef<Path>, I: Iterator<Item = S>>(
         &mut self,
-        path: &Path,
+        path: P,
         args: I,
         params: &ExecutionParameters,
     ) -> Result<ExecutionResult, error::Error> {
-        self.parse_and_execute_script_file(path, args, params, ScriptCallType::Sourced)
+        self.parse_and_execute_script_file(path.as_ref(), args, params, ScriptCallType::Sourced)
             .await
     }
 
@@ -787,13 +788,14 @@ impl Shell {
     /// * `args` - The arguments to pass to the script as positional parameters.
     /// * `params` - Execution parameters.
     /// * `call_type` - The type of script call being made.
-    async fn parse_and_execute_script_file<S: AsRef<str>, I: Iterator<Item = S>>(
+    async fn parse_and_execute_script_file<S: AsRef<str>, P: AsRef<Path>, I: Iterator<Item = S>>(
         &mut self,
-        path: &Path,
+        path: P,
         args: I,
         params: &ExecutionParameters,
         call_type: ScriptCallType,
     ) -> Result<ExecutionResult, error::Error> {
+        let path = path.as_ref();
         tracing::debug!("sourcing: {}", path.display());
         let opened_file: openfiles::OpenFile = self
             .open_file(path, params)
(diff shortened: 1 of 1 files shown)


reubeno:
Thanks for looking at this, @Elsie19; I agree that relaxing these constraints will make these functions a fair bit easier to use.

I see a mix of using `impl` vs. adding a new generic parameter. Was there a reason you went with one pattern in some cases and the other elsewhere?

_Broader context..._ I'd like to start establishing some consistent guidelines for the code base, and this is a good case to consider. I know there was precedent in `brush` to go with adding an explicit generic type parameter, but looking through your changes and thinking more about it, I think I prefer the `impl` pattern for cases like this where it's a per-parameter _convenience_ as opposed to an important concept in the function's signature (and might be more likely to be used with turbofish syntax).

Do you have a perspective on this? I tried searching for other projects with clear guidelines on this topic, but my ~5 minute search didn't find anything conclusive yet.

Elsie19:
> Thanks for looking at this, @Elsie19; I agree that relaxing these constraints will make these functions a fair bit easier to use.
> 
> I see a mix of using `impl` vs. adding a new generic parameter. Was there a reason you went with one pattern in some cases and the other elsewhere?

Yeah, if there are methods that already have generic parameters, I add a new one to it, but if it's a method without generics yet, I use the `impl` syntax.

> _Broader context..._ I'd like to start establishing some consistent guidelines for the code base, and this is a good case to consider. I know there was precedent in `brush` to go with adding an explicit generic type parameter, but looking through your changes and thinking more about it, I think I prefer the `impl` pattern for cases like this where it's a per-parameter _convenience_ as opposed to an important concept in the function's signature (and might be more likely to be used with turbofish syntax).

Yep, that's how I think about it, if there's just one parameter that needs some generic, I use an `impl`, but if it's multiple parameters, I wouldn't write something like:

```rs
  async fn parse_and_execute_script_file<S: AsRef<str>, I: Iterator<Item = S>>(
        &mut self,
        path: impl AsRef<Path>,
        args: I,
        params: &ExecutionParameters,
        call_type: ScriptCallType,
    ) -> Result<ExecutionResult, error::Error>
```

> Do you have a perspective on this? I tried searching for other projects with clear guidelines on this topic, but my ~5 minute search didn't find anything conclusive yet.

Yeah:

* One off method parameters should use `impl` whenever possible.
* If you are referencing generics inside generics (such as `S: AsRef<str>, I: Iterator<Item = S>`), that should go inside a where clause.
* Otherwise, use the `<>` syntax.

So to illustrate:

## Using `impl` inline

✅ 
```rs
fn foo(path: impl AsRef<Path>) {}
```
❌
```rs
fn foo<P: AsRef<Path>>(path: P) {}
```
❌
```rs
fn foo<P>(path: P)
where P: AsRef<Path> {}
```

## Generics within generics

✅ 
```rs
fn foo<I, T>(path: I)
where
    I: IntoIterator<Item = T>,
    T: Into<String>, {}
```
❌
```rs
fn foo<I, T: Into<String>>(path: I)
where
	I: IntoIterator<Item = T>, {}
```
❌
```rs
fn foo<I, T: Into<String>>(path: impl IntoIterator<Item = T>) {}
```

## Generic bounds

✅ 
```rs
fn foo<P: AsRef<Path>>(path: P, text: &str) {}
```
❌
```rs
fn foo(path: impl AsRef<Path>, text: &str) {}
```
❌
```rs
fn foo<P>(path: P, text: &str)
where P: AsRef<Path> {}
```

reubeno:
Thanks for the longer-form thoughts. I'm generally aligned, but would like to hear more about your thinking on the last case (labeled "generic bounds"). I'm not sure I follow why you wouldn't have gone with the second option -- i.e. why just adding a `&str` would cause you not to still use `impl AsRef<Path>` on `path`.

reubeno:
I should mention that, for the purposes of this PR, I'm okay moving ahead. I would also like to see us go back and separately make (at least) the exported functions more consistent.

reubeno (approved):

Elsie19:
> Thanks for the longer-form thoughts. I'm generally aligned, but would like to hear more about your thinking on the last case (labeled "generic bounds"). I'm not sure I follow why you wouldn't have gone with the second option -- i.e. why just adding a `&str` would cause you not to still use `impl AsRef<Path>` on `path`.

So the last one has a little bit of wiggle room. If the method is just a tiny one, sure, use `impl` or generic bounds, whatever floats your boat, but really it's whatever makes the amount of cognition to understand a call go down is better.

PR #455, opened 2025-04-29
chore(traps): convert to public interface

### LINKED ISSUE #454: Are there supposed to be non-exported types in `brush_core` public structs?
I'm in the middle of using `brush_core` in an app I'm working on, though I notice that [the documentation has multiple variables whose types aren't publically available](https://docs.rs/brush-core/latest/brush_core/struct.Shell.html) (i.e. `TrapHandlerConfig` in [`Shell::traps`](https://docs.rs/brush-core/latest/brush_core/struct.Shell.html#structfield.traps)).

Is this intended? I feel like the types themselves should be public, assuming the variables themselves are as well.

Elsie19:
I expect that there will be some more discussions as to what else needs/not to be public, but I am making it now to get the ball rolling.

---

Closes #454.

### DIFF
--- a/brush-core/src/lib.rs
+++ b/brush-core/src/lib.rs
@@ -31,7 +31,7 @@ mod terminal;
 mod tests;
 mod timing;
 mod trace_categories;
-mod traps;
+pub mod traps;
 mod variables;
 
 pub use commands::ExecutionContext;
--- a/brush-core/src/traps.rs
+++ b/brush-core/src/traps.rs
@@ -1,3 +1,5 @@
+//! Implements various trapping capabilties.
+
 use std::str::FromStr;
 use std::{collections::HashMap, fmt::Display};
 
@@ -131,6 +133,7 @@ impl TryFrom<&str> for TrapSignal {
     }
 }
 
+/// Error type for [`TrapSignal`] conversions.
 #[derive(Debug, Clone, Copy)]
 pub struct DoesntHaveANumber;
 


reubeno:
I took a look back; I'm supportive of exporting this module but would want to see a couple of minor updates first (see notes below). Totally understood that this is really a small bit of code but I believe it's important to strike a balance between providing flexibility and exposing reasonable interfaces we can maintain over time without breaking (while still changing internals). And, well, this is small enough to use as a case to refine principles for these exported surfaces:

* The public `handler_depth` field sticks out as an internal implementation detail manipulated by `Shell` and some other `brush_core` internals. I'm thinking it should be moved out of this module (or at least out of `TrapHandlerConfig`), since it's more about shell state than it is about trap handler registrations.

* We already provide abstract some accessors for registering/unregistering handlers (`register_handler` and `remove_handlers`). This gives us room--should we need it in the future--to add validation, tracing, etc. I'm considering the value of making the underlying `HashMap` private and exposing any other needed accessors (e.g., a read-only iterator). I have the same feedback for the `functions` module; we've already had to update its internal representation of definitions at least once for performance reasons, and I wouldn't be surprised if we needed to do so again.

I think one of the challenges here -- which I, admittedly, baked into the code :smile: -- is that this module doesn't know if it just wants to store registration data or if it wants to be something more (e.g., manage state, implement handler invocation mechanisms).

Elsie19:
> I thought about this some more, and I'd like to proceed with a slight variant of this PR, with the main difference being marking `TrapHandlerConfig`'s members as `pub(crate)` for now. I don't like the idea of wildly proliferating crate-public fields and types, but that would give us the space to clean it up in the ways I was suggesting above and _without_ holding back a good, pragmatic change like the one in this draft PR.
> 
> @Elsie19 what do you think about that?

Whatever works for you 😎

reubeno:
Do you want to add those `pub(crate)` annotations to the fields in `TrapHandlerConfig` and mark this ready for review, or are you okay if I subsume this in a broader separate PR?

Elsie19:
I think if you got more work, make a new PR. I just made this one to get the ball rolling!

reubeno:
Thanks for getting the ball rolling! I'm closing this out now that #459 has subsumed it.

PR #459, opened 2025-05-04
fix: enable unnameable_types lint and mitigate errors

### LINKED ISSUE #454: Are there supposed to be non-exported types in `brush_core` public structs?

reubeno:
* Adds `unnameable_types` to enforce intentionality on exporting types; if we export a type that depends on another type, that depended-on type must be explicitly exported as well. 
* Marks some additional modules and types as explicitly exported.
* Marks some not-desired-to-be-public struct fields as `pub(crate)` to avoid hitting this lint. Additional work will be required in some of these cases to ensure we’ve exported sufficiently powerful functions to manipulate the state.

Resolves #454


brush-core/src/interp.rs
```diff
@@ -109,32 +109,45 @@ struct PipelineExecutionContext<'a> {
...
     }
 
     /// Returns the file descriptor with the given number.
     #[allow(clippy::unwrap_in_result)]
-    pub fn fd(&self, fd: u32) -> Option<openfiles::OpenFile> {
+    pub(crate) fn fd(&self, fd: u32) -> Option<openfiles::OpenFile> {
         self.open_files.files.get(&fd).map(|f| f.try_dup().unwrap())
     }
+
+    pub(crate) fn stdin_file(&self) -> openfiles::OpenFile {
```
Elsie19, line 139: I know it's kinda late but shouldn't this be:

```suggestion
    pub(crate) fn stdin_file(&self) -> impl std::io::Write {
```

reubeno: Fair question. I'd prefer not to have this fn at all. The change updated `stdin()` et al. to return an `impl` of a trait in `std::io`. Almost all call sites were perfectly fine with that because they were mostly just passing the value to `write!` or similar. 

But... there were a few internal-to-brush-core callers that were doing something more complicated. Rather than sort them out more properly I opted to add a few `pub(crate)` helpers (e.g. stdin_file()` that still returned an `OpenFile` or similar.

In the medium term, I'd like to take a pass over `pub(crate)` functions on exported types (like these) and find a way to remove them or properly export them.

And thanks for taking a look, @Elsie19 🙂 I expedited this one in before the end of the weekend, but am always happy to get the feedback whether it's before or after merge. I'm also happy to let changes like this sit for a bit in the future if I know there's interest / willingness for someone else to read through them.

PR #478, opened 2025-05-09
feat(builtins): implement basic `bind -x` support

reubeno:
There's a lot here. Supporting even the most minimal functionality in `bind` required:

* Some `reedline`-specific integration code in `brush-interactive` that lets us add an indirection between `reedline` and the key bindings. This was needed to let us mutate the key bindings at runtime.
* A pattern and interface for us to let the `bind` builtin in `brush-core` "upcall" back to the interactive shell layer for querying and mutating bindings.
* Implementation of this new trait for the `reedline` input backend.
* Parsing the `readline` binding syntax supported by `bind`.
* Translating the key codes referenced in the `readline` binding syntax into the abstract keys that `reedline` allows configuring.
* Implementing `READLINE_LINE` and `READLINE_POINT` handling in the shell.

With this support in place, projects like [atuin](https://atuin.sh/) are *starting* to work. (There's still gaps, but this is a big step.)

_Relates to: #380_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.15",
+ "getrandom 0.2.16",
  "once_cell",
  "version_check",
  "zerocopy 0.7.35",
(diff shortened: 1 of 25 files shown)


reubeno:
There's a risk of this impacting prompt display performance, due to the indirection required for the mutable edit mode. But given the size of the change, I think we should merge this in sooner than later and give it a bit more time to bake before the next release.

I'd prefer not to do so, but if needed we could add some conditional controls around whether runtime key-binding changes are supported--so either it requires opt-in, or else there's a way to avoid the hit via opt out. If the impact is in the noise, then we can avoid worrying about all that and proceed on.

PR #482, opened 2025-05-10
Implement ulimit

### LINKED ISSUE #473: Implement builtin: `ulimit`
_This was raised by @ghuls in #469 but it deserves its own top-level issue._

We haven't really looked at `ulimit` yet, but I'm hoping it won't be as tricky as some of the other last few builtins. We should look into the [rlimit crate](https://crates.io/crates/rlimit) as well as what the nix crate (and others) expose.

It would also be reasonable to start small, perhaps first just installing the read-only parts of it?

lu-zero:
Closes #473

### DIFF
--- a/brush-core/Cargo.toml
+++ b/brush-core/Cargo.toml
@@ -65,6 +65,7 @@ nix = { version = "0.30.1", features = [
     "term",
     "user",
 ] }
+rlimit = "0.10.2"
 terminfo = "0.9.0"
 uzers = "0.12.1"
 
--- a/brush-core/src/builtins/ulimit.rs
+++ b/brush-core/src/builtins/ulimit.rs
@@ -197,6 +197,13 @@ impl ResourceDescription {
         short: 'v',
         unit: Unit::KBytes,
     };
+    const THREADS: ResourceDescription = ResourceDescription {
+        resource: Resource::Phy(rlimit::Resource::THREADS),
+        help: "the maximum number of threads",
+        description: "number of threads",
+        short: 'T',
+        unit: Unit::Number,
+    };
 
     fn get(&self, hard: bool) -> std::io::Result<String> {
         let (soft_limit, hard_limit) = self.resource.get()?;
@@ -361,8 +368,8 @@ pub(crate) struct ULimitCommand {
     /// Unimplemented
     #[arg(short = 'R', default_missing_value = "", num_args(0..=1))]
     rttime: Option<LimitValue>,
-    /// Unimplemented
-    #[arg(short = 'T', default_missing_value = "", num_args(0..=1))]
+    /// the maximum number of threads
+    #[arg(short = 'T', default_missing_value = "", num_args(0..=1), help = ResourceDescription::THREADS)]
     threads: Option<LimitValue>,
 
     /// argument for the implicit limit (`-f`)
@@ -389,11 +396,7 @@ impl builtins::Command for ULimitCommand {
             }
         };
 
-        if self.threads.is_some()
-            || self.rttime.is_some()
-            || self.npts.is_some()
-            || self.file_lock.is_some()
-        {
+        if self.rttime.is_some() || self.npts.is_some() || self.file_lock.is_some() {
             return crate::error::unimp("Limit unimplemented");
         }
 
@@ -411,6 +414,7 @@ impl builtins::Command for ULimitCommand {
         set_or_get(self.msgqueue, ResourceDescription::MSGQUEUE);
         set_or_get(self.rtprio, ResourceDescription::RTPRIO);
         set_or_get(self.stack, ResourceDescription::STACK);
+        set_or_get(self.threads, ResourceDescription::THREADS);
         set_or_get(self.cpu, ResourceDescription::CPU);
         set_or_get(self.nproc, ResourceDescription::NPROC);
         set_or_get(self.vmem, ResourceDescription::VMEM);
--- a/brush-core/src/builtins/ulimit.rs
+++ b/brush-core/src/builtins/ulimit.rs
@@ -67,6 +67,13 @@ impl ResourceDescription {
         short: 'l',
         unit: Unit::KBytes,
     };
+    const MSGQUEUE: ResourceDescription = ResourceDescription {
+        resource: Resource::MSGQUEUE,
+        help: "the maximum number of bytes in POSIX message queues",
+        description: "POSIX message queues",
+        short: 'q',
+        unit: Unit::Bytes,
+    };
     const RSS: ResourceDescription = ResourceDescription {
         resource: Resource::RSS,
         help: "the maximum resident set size",
@@ -227,6 +234,9 @@ pub(crate) struct ULimitCommand {
     /// the maximum number of open file descriptors
     #[arg(short = 'n', default_missing_value = "", num_args(0..=1), help = ResourceDescription::NOFILE)]
     file_open: Option<LimitValue>,
+    /// the maximum number of bytes in POSIX message queues
+    #[arg(short = 'q', default_missing_value = "", num_args(0..=1), help = ResourceDescription::MSGQUEUE)]
+    msgqueue: Option<LimitValue>,
     /// the maximum scheduling priority (`nice`)
     #[arg(short = 'e', default_missing_value = "", num_args(0..=1), help = ResourceDescription::NICE)]
     nice: Option<LimitValue>,
@@ -264,6 +274,7 @@ impl builtins::Command for ULimitCommand {
         set_or_get(self.rss, ResourceDescription::RSS);
         set_or_get(self.file_open, ResourceDescription::NOFILE);
         set_or_get(self.nice, ResourceDescription::NICE);
+        set_or_get(self.msgqueue, ResourceDescription::MSGQUEUE);
 
         if resources_to_set.is_empty() {
             if resources_to_get.is_empty() {
--- a/brush-core/src/builtins/ulimit.rs
+++ b/brush-core/src/builtins/ulimit.rs
@@ -2,7 +2,10 @@ use clap::{
     builder::{IntoResettable, StyledStr},
     Parser,
 };
-use std::{io::ErrorKind, str::FromStr};
+use std::{
+    io::{self, ErrorKind, Write},
+    str::FromStr,
+};
 
 use crate::{builtins, commands};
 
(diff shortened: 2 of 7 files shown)


reubeno:
Thanks for tackling this one (and for the rustfmt config fix)! I should be able to properly review it this weekend.

lu-zero:
They way I wrote it should make quite easy to add additional entries to newcomers, sadly it hits an annoying corner case in clap or two so there is a bit more code than it could be once those are addressed.

brush-core/Cargo.toml
```diff
@@ -64,6 +64,7 @@ nix = { version = "0.30.1", features = [
     "user",
 ] }
 uzers = "0.12.1"
+rlimit = "0.10.2"
```
reubeno, line 68: For readability, would you please move this one line up to keep the crates in sorted order?

brush-core/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,206 @@
...
+        set_or_get(self.file_size, ResourceDescription::FSIZE);
+        set_or_get(self.rss, ResourceDescription::RSS);
+        set_or_get(self.file_open, ResourceDescription::NOFILE);
+
+        if self.all {
+            resources_to_get = ALL_RESOURCES.into();
+        }
+
+        if !resources_to_get.is_empty() && !resources_to_set.is_empty() {
+            return Err(crate::Error::InvalidArguments);
```
reubeno: Is there an error we could/should report here to `stderr`?

brush-core/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,206 @@
...
+
+#[derive(Clone, Copy, Hash, Debug)]
+enum Unit {
+    Block,
+    KBytes,
+    Number,
+}
+
+#[derive(Clone, Copy, Hash, Debug)]
+struct ResourceDescription {
```
reubeno, line 97: I appreciate the data-driven definition of the resources!

brush-core/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,206 @@
...
+            Some(v) => resources_to_set.push((descr, v)),
+            None => {}
+        };
+
+        set_or_get(self.core, ResourceDescription::CORE);
+        set_or_get(self.file_size, ResourceDescription::FSIZE);
+        set_or_get(self.rss, ResourceDescription::RSS);
+        set_or_get(self.file_open, ResourceDescription::NOFILE);
+
+        if self.all {
```
reubeno: If we folded checking `self.all` into the body of `set_or_get`, that could allow us to remove `ALL_RESOURCES` and leave only one place where we need to explicitly list all resource types (i.e., on lines 169-172). It would also complicate `set_or_get`.

What do you think about that trade-off?

brush-core/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,206 @@
...
+        Ok(v)
+    }
+}
+
+/// Modify shell resource limits.
+///
+/// Provides control over the resources available to the shell and processes
+/// it creates, on systems that allow such control.
+#[derive(Parser, Debug)]
+pub(crate) struct ULimitCommand {
```
reubeno, line 354: For most built-ins we've implemented, we've tried to leave placeholder stubs for all of the options -- so runtime use of them results in an "unimplemented error" that we can get reports on and track (vs. argument parsing errors), e.g.:

https://github.com/reubeno/brush/blob/7a7c642454aa0e26027d2abb436d3c7ba40dd8fc/brush-core/src/builtins/mapfile.rs#L57-L59

I'm supportive of getting this change in as incremental progress without implementing the other options but would like to see similar "unimplemented stubs" here. Would that be feasible?

brush-core/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,206 @@
+use clap::Parser;
+use rlimit::Resource;
+use std::str::FromStr;
+
+use crate::{builtins, commands};
+
+#[derive(Clone, Copy, Hash, Debug)]
```
reubeno: Are `Hash` and `Debug` actually required on these types?

brush-core/Cargo.toml
lu-zero: sure

brush-core/src/builtins/ulimit.rs
lu-zero: We can even avoid that restriction.

brush-core/src/builtins/ulimit.rs
lu-zero: let me add one

brush-core/src/builtins/ulimit.rs
lu-zero: Not needed, initially I thought of having an `HashMap`.

Btw can we remove `Cargo.lock` from git, it causes more churn than anything :/

brush-core/src/builtins/ulimit.rs
lu-zero: It is simple enough

brush-core/src/builtins/ulimit.rs
lu-zero: Overall takes more or less the same time to complete it though...

reubeno: Got it. I agree that, if it takes just as much time to add the stub as the real implementation, then we'd appreciate the real implementation :smile:

_(I wasn't sure if you were intentionally limiting the implementation because (a) you wanted to leave some work for other interested contributors, (b) you were short on time to look at the other options, or (c) the other options were trickier. All of those are, of course, reasonable explanations.)_

lu-zero: Mostly `a`, plus running out of time that weekend. All the options have potentially only the annoyance of having to pick the right name for the resource since some are also synonyms, but it literally boils down to fill in the structure and add the line to use it.

reubeno: I definitely appreciate you being supportive of other contributors :smile:.

I'd prefer stubs of some kind, but in this case I'd compromise on at least getting placeholder `// TODO` comment(s) so we have some way to track the missing functionality and hold ourselves accountable for getting it filled in.

lu-zero: Once I have another timeslice I'll fill in more, shouldn't take much either way. We might need some creative solution to deal with synonyms or contribute the failover to the rlimit crate itself.

lu-zero:
Now I added the unsupported flags, some aren't real rlimits

reubeno:
> Now I added the unsupported flags, some aren't real rlimits

Thanks, @lu-zero! We appreciate you putting more time into this.

I'll be able to re-review with your updates within the next ~day.

lu-zero:
Now I added an example also for a virtual resource

lu-zero:
The best way we have to test the getters is to compare flag by flags with bash ulimit, but I'm fairly sure even in that there may be a discrepancy. Setters might be way trickier

lu-zero:
But I'll need you to tell me how to write such tests

reubeno:
> But I'll need you to tell me how to write such tests

There's some basic documentation [here](https://github.com/reubeno/brush/blob/main/docs/reference/integration-testing.md). The main idea of these compat tests is finding sequences of commands that we'd expect to behave identically between `bash` and `brush`. 

In this case, if you expect the getters to retrieve the same values, then invoking those in a test case would be a great way to pin the behavior. 

To be clear, I'm okay with us deferring some testing to subsequent PRs. If there's a very basic test that you'd recommend, that might be a good addition now. 

What do you think?

lu-zero:
Let me give you an example:
```
❯ ulimit -a
core file size              (blocks, -c) 0
data seg size               (kbytes, -d) unlimited
file size                   (blocks, -f) unlimited
max locked memory           (kbytes, -l) unlimited
max memory size             (kbytes, -m) unlimited
open files                          (-n) 256
pipe size                (512 bytes, -p) 1
stack size                  (kbytes, -s) 8176
cpu time                   (seconds, -t) unlimited
max user processes                  (-u) 10666
virtual memory              (kbytes, -v) unlimited
```
```
❯ cargo run -- -c "ulimit -a"
   Compiling brush-shell v0.2.17 (/Users/lu_zero/Sources/rust/brush/brush-shell)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.01s
     Running `target/debug/brush -c 'ulimit -a'`
core file size                 (block, -c) 0
data seg size                 (kbytes, -d) unlimited
file size                      (block, -f) unlimited
max locked memory             (kbytes, -l) unlimited
max memory size               (kbytes, -m) unlimited
open files                            (-n) 256
pipe size                  (512 bytes, -p) 6
stack size                    (kbytes, -s) 8372224
cpu time                     (seconds, -t) unlimited
max user processes                    (-u) 10666
virtual memory                (kbytes, -v) unlimited
```

Our ulimit is providing a better estimated value for `-p`, while bash somehow is falling back to 1 for reasons I'd rather not figure out, but the pipe size is always a fake limit to I wouldn't care much about it.

For the rest we can basically call `ulimit -{not p} {some sensible value or unlimited}` and see if brush ulimit returns the same value and the other way round.

reubeno:
My recommendation would be to spend a very limited amount of time adding test case(s) that pins the behavior that seems reliably identical. Where there is some deviation, let's leave that as a future concern to worry about (or not, if it's not worth tracking down) 😀

lu-zero:
Looks like the yaml didn't make the CI happy at all.

reubeno:
> Looks like the yaml didn't make the CI happy at all.

Do some of those test cases try to change the limits? The CI tests failing are all running in more constrained containers.

I think we can dig into more of this separately, so I'd recommend limiting to the simpler tests that are non-side-effectful and look like they'll run reliably in CI and filing follow-up issues for any additional coverage you think we should invest in later.

lu-zero:
posix ulimit has a single `-f` limit, so I wouldn't spend more time on it until we have somebody needing it.

brush-core/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,489 @@
...
+            Unit::Block => format!("(block, -{})", self.short),
+            Unit::Bytes => format!("(bytes, -{})", self.short),
+            Unit::HalfKBytes => format!("(512 bytes, -{})", self.short),
+            Unit::KBytes => format!("(kbytes, -{})", self.short),
+            Unit::Micros => format!("(microseconds, -{})", self.short),
+            Unit::Number => format!("(-{})", self.short),
+            Unit::Seconds => format!("(seconds, -{})", self.short),
+        };
+        let resource = self.get(hard).unwrap_or_else(|e| format!("{e}"));
+        println!("{:<26}{:>16} {}", self.description, unit, resource);
```
reubeno: I forgot to note this earlier -- instead of calling `print*`, builtins should use `write*` with `context.stdout()` or `context.stderr()`. That ensures it will go to the right output stream for the current execution context (in case it was redirected).

Ditto for the other calls to `print*` in this file.

reubeno (approved):

reubeno:
Thanks for all the work on this, @lu-zero; I plan to merge this today.

PR #486, opened 2025-05-11
feat(mapfile): Register also as readarray

### LINKED ISSUE #474: Implement builtin `readarray` as wrapper over `mapfile`
Per documentation, `readarray` basically seems to be an alias for `mapfile`, which we already have a partial implementation of.

For the scope of this issue, we should: (1) do some research to see if there *is* a difference in functionality, and (2) wire up `readarray` to call into shared code with `mapfile`. This can be done similar to how we've handled `readonly` being a thin wrapper on `declare`.

We can separately track filling out missing functionality in `mapfile`.

lu-zero:
Closes #474

### DIFF
--- a/brush-core/src/builtins/factory.rs
+++ b/brush-core/src/builtins/factory.rs
@@ -283,6 +283,7 @@ pub(crate) fn get_default_builtins(
         m.insert("enable".into(), builtin::<enable::EnableCommand>());
         m.insert("let".into(), builtin::<let_::LetCommand>());
         m.insert("mapfile".into(), builtin::<mapfile::MapFileCommand>());
+        m.insert("readarray".into(), builtin::<mapfile::MapFileCommand>());
         m.insert("printf".into(), builtin::<printf::PrintfCommand>());
         m.insert("shopt".into(), builtin::<shopt::ShoptCommand>());
         m.insert("source".into(), builtin::<dot::DotCommand>().special());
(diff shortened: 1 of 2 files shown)


reubeno:
Thanks for picking this one up!

Could you please extend the compat tests to ensure we cover this? I don't recall offhand what coverage we already have for `mapfile` but it would be good to get something basic to prevent future regression.

Just holler if you need pointers on the compat tests; there's some minimal documentation under `docs`.

lu-zero:
But definitely I need directions on how to update the reference files.

brush-core/src/builtins/factory.rs
```diff
@@ -283,6 +283,7 @@ pub(crate) fn get_default_builtins(
         m.insert("enable".into(), builtin::<enable::EnableCommand>());
         m.insert("let".into(), builtin::<let_::LetCommand>());
         m.insert("mapfile".into(), builtin::<mapfile::MapFileCommand>());
+        m.insert("readarray".into(), builtin::<mapfile::MapFileCommand>());
```
reubeno, line 286: You'll need to remove lines 321-322. That would also explain the test failures--the stub registered below is overriding the registration here.

reubeno (changes requested):

reubeno:
> But definitely I need directions on how to update the reference files.

There aren't any checked-in reference files. The reference being used is the installed copy of `bash`. The [doc page](https://github.com/reubeno/brush/blob/main/docs/reference/integration-testing.md) explains it a bit, but to make it easier to write tests we run the test on `brush` and `bash` and compare the results (i.e., exit code, `stdout`, `stderr`).

To run tests locally you can follow [these steps](https://github.com/reubeno/brush/blob/main/docs/how-to/run-tests.md).

reubeno:
Thanks for resolving the double-registration. Did the YAML test case you'd added somehow get lost in the process? Assuming you're able to add that back in (and it passes), we can get this merged.

PR #490, opened 2025-05-12
feat(mapfile): implement `-n`, `-d`, `-s`

Elsie19:


Elsie19:
I sort of over-engineered `LineCount`. I can revert the last commit if you want, I just think (barring line count), it looks and feels much nicer to use.

reubeno:
> I sort of over-engineered `LineCount`. I can revert the last commit if you want, I just think (barring line count), it looks and feels much nicer to use.

I agree it looks nicer to use in the core implementation logic of the builtin. I also agree it might be nice to simplify the `LineCount` implementation a bit.

What do you think about using `Range`s to reduce the custom code, e.g.:

```rust
use either;

/// Iterates from 0 up to (but not including) the given `count` value;
/// as a special case, if `count` is 0, continues iterating indefinitely.
fn count_to(count: i32) -> impl Iterator<Item = i32> {
    if count == 0 {
        either::Left(0..)
    } else {
        either::Right(0..count)
    }
}
```

I had to use the `either` crate (or else `Box<dyn...>`) because the indefinite range and the limited range have different concrete types and I placed it in a separate function, but hopefully you get the idea.

_I was also idly wondering whether we might consider placing the actual body of the read loop in an iterator and then be able to use adaptors like `.take(max_count)` to implement the `-n` constraint. With that said, I didn't think too deeply into that and am not sure if we'd be happier with the result._

(Also, shout-out for adding `default_value = "MAPFILE"`. That's a good compat fix!)

Elsie19:
I don't think adding an extra crate just to reduce the line count in one file is particularly useful.

We could also maybe move `LineCount` into `OpenFile` to abstract it away from `mapfile` as well.

I was also thinking about moving the read loop into the iterator. I was actually wondering why it *wasn't* the case in the first place!

reubeno:
> I don't think adding an extra crate just to reduce the line count in one file is particularly useful.

In general, I agree with your hesitance. In this specific case, `either` is something I'd looked at multiple times previously, seems reasonably well supported, doesn't have any dependencies of its own (when used with only default features), and provides usability value. I think it would be acceptable to pull it in here, but I'm okay with other options too.

> We could also maybe move LineCount into OpenFile to abstract it away from mapfile as well.

I'd be fine--and would probably prefer--it living in its own `.rs` file if we don't go with something like the `count_to` function above; I just would want it to be `pub(crate)` and not export it.

> I was actually wondering why it wasn't the case in the first place! I think the implementation of mapfile could be a lot more readable if it was rewritten.

I don't think there was a particular reason. We should be mindful of the fallibility of iterating and make sure we can reasonably implement `-C` / `-c` options for sending lines to a callback in batch... but assuming those seem fine, please feel free to propose a refactoring that you think will be easier to read and maintain.

Elsie19:
I think that should come in a later PR. I will add tests and undraft this.

Elsie19:
Actually I might need to rework all of mapfile, it's not very futureproof and I can't get `-n` to work.

reubeno:
> On a somewhat related note, I think you should look at https://github.com/bminor/bash/tree/master/tests. This would *really* uncover any implementation bugs.

Yep, it's on our list to figure out how to use the core bash tests. (We've got an issue somewhere that briefly mentions this, I think?)

As context, my initial hesitance had been due to licensing concern. Throughout brush's development, I haven't looked at bash sources or taken any code from its code base to avoid issues in that space. But with that said, we should figure out how to at least run the tests without importing them into this repo--perhaps similar to what projects like uutils have done or like what we do with the bash-completion tests. 

I've also separately reached out to POSIX to inquire about getting a temporary license to their official conformance tests. That will likely require a fair bit of work to make sure we really have the right conditionals around all non-POSIX extensions.

Elsie19:
> I've also separately reached out to POSIX to inquire about getting a temporary license to their official conformance tests. That will likely require a fair bit of work to make sure we really have the right conditionals around all non-POSIX extensions.

I think that's a wonderful idea!

Elsie19:
Ugh I still forgot tests, hold on...

Elsie19:
Concurrently with doing tests, I do have something I'd like fixed if possible:

```rs
let orig_term_attr = setup_terminal_settings(&input_file)?;

let mut entries = vec![];
let mut buf_reader = BufReader::new(input_file.clone());
```

Can you think of any ways to remove the clone in `BufReader`? I know that `OpenFile` is small, but it never hurts to remove any copying that isn't strictly necessary.

reubeno:
Thanks for adding some basic tests. I skimmed through the change quickly and had a few broader questions... (I can look more closely within the next ~day or so.)

Is there a reason this change shifts to using `BufReader`? Note that the original implementation wasn't ideal, but it was intentionally reading a single character at a time (without any explicit buffering) to ensure that we don't read too many characters from the underlying file/stream--i.e., that we stop reading as soon as we read the final character. With `BufReader` there's a risk that it will read a larger chunk from the underlying input file, even if we've only read a strict subset of the characters from it.

Also, do we need to bring back some of the special Ctrl-C/Ctrl-D handling? It's hard to test those without a more interactive/pty-style test, but can you at least confirm that they still work as expected?

Elsie19:
> Thanks for adding some basic tests. I skimmed through the change quickly and had a few broader questions... (I can look more closely within  Is there a reason this change shifts to using `BufReader`? Note that the original implementation wasn't ideal, but it was intentionally reading a single character at a time (without any explicit buffering) to ensure that we don't read too many characters from the underlying file/stream--i.e., that we stop reading as soon as we read the final character. With `BufReader` there's a risk that it will read a larger chunk from the underlying input file, even if we've only read a strict subset of the characters from it.

> It can be excessively inefficient to work directly with a Read instance. For example, every call to read on TcpStream results in a system call. A BufReader<R> performs large, infrequent reads on the underlying Read and maintains an in-memory buffer of the results.

https://doc.rust-lang.org/std/io/struct.BufReader.html

> Also, do we need to bring back some of the special Ctrl-C/Ctrl-D handling? It's hard to test those without a more interactive/pty-style test, but can you at least confirm that they still work as expected?

Oh yes I should add that back! Thanks for reminding me.

reubeno:
I'm familiar with the benefits of buffered reads :smile:. I'd absolutely support using it if we could sort out the semantics -- but what I'm saying is that there's a functional requirement for builtins like `read`, `readarray`/`mapfile`, etc. to read no more bytes from the underlying OS file stream than they are supposed to consume. Otherwise, another command running in the pipeline/context afterwards may experience data loss (because the `BufReader` would have already read in those bytes in its buffer, but discarded them when it was dropped).

In case it helps, here's a contrived example of a scenario I'm concerned with:

* User runs something like: `seq 1 3 | (mapfile -n 1; lolcat; echo $MAPFILE)`
* `brush` will establish a single pipe that `seq` writes into, and from which the `mapfile`/`lolcat` subshell will read from.
* `seq` will write `1\n2\n3\n` into the writer end of the pipe.
* `mapfile` runs, with its `stdin` mapped to the reader end of the pipe, and:
  * creates a `BufReader` atop `stdin` (the pipe reader)
  * reads from the `BufReader`
  * under the hood, the implementation of `BufReader` reads _some number of bytes it decides_ from the pipe and fills its buffer; it then returns to its caller the number of bytes requested. Those bytes it read are now only in its buffer and no longer present in the pipe for future reads.
  * sees `1` and then `\n` and realizes it's read enough; it sets `MAPFILE` to `"1"`
  * (implicitly) drops the `BufReader`, which results in the `BufReader`'s internal buffer getting dropped on the floor, along with the bytes that are still in it
* `lolcat` runs, with its `stdin` mapped to the same reader end of the pipe
  * it reads from its `stdin` and *doesn't* see `2`, because `2` had already been read into the `BufReader` buffer above and that byte is no longer present in the pipe

Expected output is something like:

image

With reads buffered in this way, it's entirely possible the `BufReader` could read *all* the bytes from the pipe, and `lolcat` would see nothing -- losing the remaining bytes.

Elsie19:
Oh I wasn't aware of that, but good catch. I'll rework it later. The reason I wanted to use `BufReader` was because it could do `.read_until`, which I thought would be easier to work with.

Elsie19:
Also I cannot find the flag `-r` in `lolcat`. Is that a custom command?

reubeno:
Thanks for the update; I got a bit busy yesterday but should be able to get to this during the weekend.

I'd like to put together a new release sometime next week and include this (and 2 other pending changes) in it, assuming it's ready.

Elsie19:
It's ready!

brush-core/src/builtins/mapfile.rs
```diff
@@ -93,46 +80,71 @@ impl builtins::Command for MapFileCommand {
...
 
-        let mut current_entry = String::new();
-        let mut buffer: [u8; 1] = [0; 1]; // 1-byte buffer
+        let mut entries = vec![];
+        let mut line = vec![];
+        let mut idx = self.skip_count;
+        let mut read_count = 0;
+        let delimiter = self.delimiter.chars().next().unwrap_or('\n') as u8;
+
+        let mut reader = input_file.clone();
```
reubeno: Now that the `BufReader` has been removed, I think you should be able to remove the `.clone()`. For that matter, it will probably be simpler to mark `input_file` as `mut` and remove the `reader` variable.

brush-core/src/builtins/mapfile.rs
```diff
@@ -93,46 +80,71 @@ impl builtins::Command for MapFileCommand {
...
+        let mut buf = [0u8; 1];
+
+        while self.max_count == 0 || entries.len() < usize::try_from(self.max_count)? {
```
reubeno: To streamline iterations of the loop, it's probably worth extracting the `try_from()` call out of the loop, and just executing it once beforehand.

brush-core/src/builtins/mapfile.rs
```diff
@@ -93,46 +80,71 @@ impl builtins::Command for MapFileCommand {
...
+            line.clear();
```
reubeno: If `line` is going to get cleared each iteration anyway, it would be simpler to define it local to the body of the loop (and then no need for the `.clear()`.

brush-core/src/builtins/mapfile.rs
```diff
@@ -93,46 +80,71 @@ impl builtins::Command for MapFileCommand {
...
-            // TODO: Figure out how to restore terminal settings on error?
-            let n = input_file.read(&mut buffer)?;
-            if n == 0 {
-                // EOF reached.
+            if line.is_empty() && !saw_delimiter {
                 break;
             }
 
-            let ch = buffer[0] as char;
+            if read_count < self.skip_count {
```
reubeno, line 122: Minor suggestion: right now we're tracking `read_count` and `idx`. Can we ditch `read_count` and just use `idx` as long as we (1) initialize `idx` to `0` (instead of `skip_count`) and (2) make sure to increment it even for skipped reads (e.g., right before line 125)? I think that will be somewhat more readable.

brush-core/src/builtins/mapfile.rs
```diff
@@ -93,46 +80,71 @@ impl builtins::Command for MapFileCommand {
...
-                if !self.remove_delimiter {
-                    current_entry.push(ch);
-                }
+            let line_str = String::from_utf8_lossy(&line).to_string();
 
-                entries.push((None, std::mem::take(&mut current_entry)));
-            } else {
-                current_entry.push(ch);
+            if let Some(callback) = &self.callback {
+                if (idx - self.origin.unwrap_or(0)) % self.callback_group_size == 0 {
```
reubeno: Is there a risk of division by zero if `self.callback_group_size` is `0`? Not sure how `bash` handles that.

reubeno: Should it be `idx - self.skip_count` (vs. the `idx - self.origin`)? From the docs, I'm not 100% certain. The interaction between skip_count, origin, and callback invocation seems complicated enough that it deserves some tests...

If you'd like to tackle the callbacks (and testing for them) in a separate PR, that's fine with me too.

brush-core/src/builtins/mapfile.rs
```diff
@@ -12,52 +12,43 @@ pub(crate) struct MapFileCommand {
...
     array_var_name: String,
 }
 
 impl builtins::Command for MapFileCommand {
     async fn execute(
         &self,
-        context: commands::ExecutionContext<'_>,
+        mut context: commands::ExecutionContext<'_>,
     ) -> Result<crate::builtins::ExitCode, error::Error> {
-        if self.delimiter != "\n" {
```
reubeno, line 51: Nice! You should update the PR title to reflect that you've also added support for other options too (e.g., `-d`, `-s`).

brush-core/src/builtins/mapfile.rs
```diff
@@ -93,46 +80,71 @@ impl builtins::Command for MapFileCommand {
...
-    fn read_entries(
+    async fn read_entries(
         &self,
-        mut input_file: openfiles::OpenFile,
+        input_file: openfiles::OpenFile,
+        context: &mut commands::ExecutionContext<'_>,
     ) -> Result<variables::ArrayLiteral, error::Error> {
-        let mut entries = vec![];
-
         let orig_term_attr = setup_terminal_settings(&input_file)?;
```
reubeno, line 87: In trying this out a bit, I see a (preexisting) oddity in the terminal settings we go with here. It looks like backspace still does the normal backspace thing in `bash`'s version of `mapfile`, but I'm seeing it emit `^?` in `brush`. That's not specific to this PR or any of your changes, so we can just file a separate issue on it. I'm noting it here mostly so I don't forget to mention it.

reubeno (commented):
The general flow looks good -- I've left a few more specific comments and questions. Thanks for filling out this builtin, it will be in much better shape with this PR.

brush-core/src/builtins/mapfile.rs
Elsie19: It throws an invalid callback quantum error: https://github.com/bminor/bash/blob/6794b5478f660256a1023712b5fc169196ed0a22/builtins/mapfile.def#L259

Elsie19: I think I'll move callbacks to a separate PR, it's a lot more complex than I thought going into this.

Elsie19:
I fixed all your comments, but I removed `-C`/`-c` support (for now, that can be a later PR).

brush-core/src/builtins/mapfile.rs
```diff
@@ -97,42 +84,55 @@ impl MapFileCommand {
...
         mut input_file: openfiles::OpenFile,
     ) -> Result<variables::ArrayLiteral, error::Error> {
-        let mut entries = vec![];
-
         let orig_term_attr = setup_terminal_settings(&input_file)?;
 
-        let mut current_entry = String::new();
-        let mut buffer: [u8; 1] = [0; 1]; // 1-byte buffer
+        let mut entries = vec![];
+        // let mut idx = self.skip_count;
```
reubeno:
```suggestion
```

brush-core/src/builtins/mapfile.rs
```diff
@@ -97,42 +84,55 @@ impl MapFileCommand {
...
-                    current_entry.push(ch);
-                }
+            let line_str = String::from_utf8_lossy(&line).to_string();
 
-                entries.push((None, std::mem::take(&mut current_entry)));
-            } else {
-                current_entry.push(ch);
-            }
+            entries.push((None, line_str));
+            // idx += 1;
```
reubeno:
```suggestion
```

reubeno (approved):
Changes look good, thanks for incorporating the feedback. 2 final comments -- these two commented out lines should be removable at this point, right?

Elsie19:
> Changes look good, thanks for incorporating the feedback. 2 final comments -- these two commented out lines should be removable at this point, right?

Yes, they can be removed.

PR #493, opened 2025-05-15
fix(break,continue): use `default_value_t` for flag values

Elsie19:
`default_value` is for types that can be converted into a `String`, which can be expensive, but `default_value_t` is for stack-allocated types and should be a less expensive to use.

At the level that you use `default_value`, it won't get us that much more performance, but it's still worth it.

#490 does the ones for `mapfile`.


reubeno (approved):
Nice! Righteous change.

PR #495, opened 2025-05-15
chore(edit_mode): Fix clippy lints

lu-zero:


reubeno:
Thanks for catching and fixing these, @lu-zero! Any insights as to why our PR checks haven't caught these? I'm assuming we're missing a "warnings as errors" flag somewhere?

reubeno (approved):

lu-zero:
it is just the new rust bringing in a new clippy :)

PR #505, opened 2025-05-22
docs(readme): a Fish as a shell implemented in Rust

brettcannon:


reubeno:
Thanks for the addition! It's exciting and inspiring to see the work that the `fish` community has done to migrate implementation language in-place.

PR #509, opened 2025-05-23
Allow reading a script from bytes

hwittenborn:
This allows reading input scripts from bytes, in addition to the existing string functionality.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -262,7 +262,7 @@ impl Shell {
         func_name: &str,
         body_text: &str,
     ) -> Result<(), error::Error> {
-        let mut parser = create_parser(body_text, &self.parser_options());
+        let mut parser = create_parser(body_text.as_bytes(), &self.parser_options());
         let func_body = parser.parse_function_parens_and_body()?;
 
         let func_def = brush_parser::ast::FunctionDefinition {
@@ -979,6 +979,19 @@ impl Shell {
         };
         self.run_parsed_result(parse_result, &source_info, params)
             .await
+    }
+
+    /// Parses the given bytes as a shell program, returning the resulting Abstract Syntax Tree
+    /// for the program.
+    ///
+    /// # Arguments
+    ///
+    /// * `b` - The bytes to parse as a program.
+    pub fn parse_bytes<B: Into<Vec<u8>>>(
+        &self,
+        b: B,
+    ) -> Result<brush_parser::ast::Program, brush_parser::ParseError> {
+        parse_bytes_impl(b.into(), self.parser_options())
     }
 
     /// Parses the given string as a shell program, returning the resulting Abstract Syntax Tree
@@ -1716,21 +1729,32 @@ impl Shell {
 }
 
 #[cached::proc_macro::cached(size = 64, result = true)]
+fn parse_bytes_impl(
+    b: Vec<u8>,
+    parser_options: brush_parser::ParserOptions,
+) -> Result<brush_parser::ast::Program, brush_parser::ParseError> {
+    let mut parser = create_parser(&b, &parser_options);
+
+    tracing::debug!(target: trace_categories::PARSE, "Parsing bytes as program...");
+    parser.parse_program()
+}
+
+#[cached::proc_macro::cached(size = 64, result = true)]
 fn parse_string_impl(
     s: String,
     parser_options: brush_parser::ParserOptions,
 ) -> Result<brush_parser::ast::Program, brush_parser::ParseError> {
-    let mut parser = create_parser(s.as_str(), &parser_options);
+    let mut parser = create_parser(s.as_bytes(), &parser_options);
 
     tracing::debug!(target: trace_categories::PARSE, "Parsing string as program...");
     parser.parse_program()
 }
 
 fn create_parser<'a>(
-    s: &'a str,
+    s: &'a [u8],
     parser_options: &brush_parser::ParserOptions,
 ) -> brush_parser::Parser<std::io::BufReader<&'a [u8]>> {
-    let reader = std::io::BufReader::new(s.as_bytes());
+    let reader = std::io::BufReader::new(s);
     let source_info = brush_parser::SourceInfo {
         source: String::from("main"),
     };
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -995,7 +995,10 @@ impl Shell {
         &self,
         reader: R,
     ) -> Result<brush_parser::ast::Program, brush_parser::ParseError> {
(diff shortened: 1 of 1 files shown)


brush-core/src/error.rs
```diff
@@ -51,7 +51,8 @@ pub enum Error {
     #[error("not yet implemented: {0}")]
     Unimplemented(&'static str),
 
-    /// The requested functionality has not yet been implemented in this shell; it is tracked in a GitHub issue.
+    /// The requested functionality has not yet been implemented in this shell; it is tracked in a
```
reubeno: It looks like a number of the changes in the PR are reformatting; are they intentional? And if so, would you please split them off into a separate PR to keep them separated from the more intentional code changes?

reubeno (commented):
The changes look reasonable to me. (See my one comment regarding formatting changes.)

Out of curiosity, would you be willing to share more about how you plan to use this, and what format you already have the script in? For example, would you find it easier / more efficient to be able to directly pass in a `BufReader` rather than fussing about with `Vec<u8>`s?

As broader context, I'd have much preferred to be able to directly take in a `&[u8]` or `&str` rather than an owned byte `Vec` or `String`, but that would be incompatible with the auto-caching. But I think that begs the question as to whether that caching should be the responsibility of the caller and not baked into the core parser function.

Definitely open to further change here, even if we proceed with this PR mostly as-is.

brush-core/src/error.rs
hwittenborn: Yeah, I did run the codebase through the formatter. I was going to make a separate PR for that as well, but didn't want it to seem like I was PR spamming :sweat_smile:. I'll go ahead and get that going though.

reubeno:
Caching has a noticeable positive impact on perf, but I agree it was done in something of an expedient manner. Parsing should be an idempotent operation so I'd assert it's safe but agree it would be better to be less surprising, and at least configurable.

I also like the idea of shifting toward taking `impl std::io::Read` and cleaning up the interface into `brush_core`. _(Aside: We have Grand Plans™️ for overhauling our parsers to support streaming, which will allow us to correctly handle running scripts that have giant binary payloads embedded, among other scenarios. `Read` fits much better with that.)_

For the purposes of this PR, I'd be supportive of:

* Shifting `create_parser` to take in an `impl std::io::Read`.
* Updating the proposed `parse_bytes` to take `impl std::io::Read`; collapse it with `parse_bytes_impl()` and remove the `#[cached]` attribute from it. This will get you a nicer, cleaner API for the future.
* Leaving the existing cached `parse_string()` paths as-is for _now_, to avoid changing existing behavior; file a follow-up issue to clean them up without losing the caching performance.

If you feel compelled to go further and tackle the latter item too, you're welcome to do so -- but I'd recommend breaking that up into multiple separate PRs anyhow.

I'm also open to alternative name proposals for `parse_bytes` if you think, say, `parse_program()` or `parse()` or something else would be clearer.

hwittenborn:
If you're wanting to go down the route of there being some general `std::io::Read` interface as the entrypoint to parse **everything**, I think parse()` would be a good name. In my eyes, stuff like `parse_string()` and `parse_bytes()` would then just be helper functions that call `parse()` internally.

Regardless, I can go ahead and get those first two bullets knocked out in this PR. Let me know if you have any other thoughts/changes in mind! :)

reubeno:
Getting those bullets knocked out sounds great.

I'm okay with `parse()`.

hwittenborn:
The new version has three public methods:
- `parse`: Parsing from a `Read` implementor
- `parse_bytes`: Parsing directly from a `&[u8]`
- `parse_string`: Parsing from a string

Let me know if you want any of those removed - I know we talked about `parse_bytes` turning into what's now `parse`, though it made sense to have all 3 for me. I'm not set on that though, so feel free to have me get that changed up.

brush-core/src/shell.rs
```diff
@@ -981,6 +981,28 @@ impl Shell {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `b` - The bytes to parse as a program.
+    pub fn parse_bytes<B: AsRef<[u8]>>(
+        &self,
+        b: B,
+    ) -> Result<brush_parser::ast::Program, brush_parser::ParseError> {
+        parse_bytes_impl(b.as_ref(), self.parser_options())
+    }
```
Elsie19: https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-for-%26%5Bu8%5D

This method shouldn't exist. `Read` is implemented for `&[u8]`.

Elsie19 (changes requested):

brush-core/src/shell.rs
hwittenborn: Yeah I literally saw that while I was implementing it, don't know why I wasn't thinking of that :sweat_smile:

brush-core/src/shell.rs
```diff
@@ -1715,22 +1724,32 @@ impl Shell {
...
 
+fn parse_impl<R: Read>(
+    r: R,
+    parser_options: brush_parser::ParserOptions,
+) -> Result<brush_parser::ast::Program, brush_parser::ParseError> {
+    let mut parser = create_parser(r, &parser_options);
+
+    tracing::debug!(target: trace_categories::PARSE, "Parsing reader as program...");
+    parser.parse_program()
+}
```
Elsie19: Since there is only one caller of this function, you should inline this into `parse()`.

Elsie19 (changes requested):

Elsie19:
I think later, we could think about removing `parse_string()`, because we could trivially replace all `parse_string("bla")` with `parse("bla".as_bytes())`, but that's not in the scope of this PR.

brush-core/src/shell.rs
hwittenborn: Things like `parse_string` and `parse_string_impl` were separated before this before - do you think they should be inlined @reubeno?

reubeno: The main reason they were separated before was for the caching; we needed to make sure that the parser options got incorporated into the cache "key".

I agree with @Elsie19 -- we might as well collapse `parse` and `parse_impl`.

hwittenborn:
Sorry for the slowness @reubeno, I've been busy with my day job and some summer classes starting up :smiling_face_with_tear:. This should be good to go now though - let me know if you need anything else changed!

Elsie19 (approved):

reubeno:
> Sorry for the slowness @reubeno, I've been busy with my day job and some summer classes starting up 🥲. This should be good to go now though - let me know if you need anything else changed!

No worries! Thanks for addressing the feedback :smile:. I'll get it merged shortly.

PR #510, opened 2025-05-23
Export error types from `brush_parser`

hwittenborn:
This allows the types to be used from functions like `Shell::parse_string`.


reubeno:
I know these types are already all exported directly from `brush_parser`, but as a client of `brush_core`, are you specifically looking to inspect specific parse error types for differentiated error messages / handling?

The reason I ask is that we have future plans to overhaul the parsers in `brush` (#283 has some context) and I'm wondering if we need to clean up / better abstract some of the error types to anticipate that. Some of the error values directly wrap `peg` errors, which is very parser implementation specific right now. (Not ideal, but it grew very organically.) 

Any chance it's mostly just `brush_parser::ParseError` that you need?

brush-core/src/lib.rs
```diff
@@ -36,6 +36,9 @@ pub mod traps;
 pub mod variables;
 
 pub use arithmetic::EvalError;
+pub use brush_parser::{
```
reubeno: I left a more general comment about exposing these, but with that said, I'd lean toward any re-exports of `brush_parser` types be done in a child module (e.g., `brush_core::parser`).

hwittenborn:
Yep, it's just `ParseError` that I'm using currently. I wasn't aware of that overhaul, so we could definitely just export `ParseError` for now.

brush-core/src/lib.rs
hwittenborn: On that note, would it be possible to just export the entirety of `brush_parser` as a `parser` module? This would help to avoid having to list multiple crates in my `Cargo.toml` configuration, especially since the two crates are intertwined with each other.

reubeno: I'd prefer starting with just exporting the types that are already directly passed through `brush_core` to clients of the latter. That includes `ParseError` and the entire `brush_parser::ast` module too. (And per the other thread, I'd like to see those published via a `parser` sub-module of `brush_core`.)

Re-exporting everything in `brush_parser` expands the compatibility contract (and ties the semver) of `brush_core` to `brush_parser`. With plans to revisit how the parser is implemented--and probably some need to clean up some of the exported helper functions in `brush_parser`, I think it's prudent to hold off on that.

Apart from the AST types and the error types, are you finding your code interacting with other parts of `brush_parser` directly?

hwittenborn: It's just the AST and `ParseError` type. I'd still be in favor of export `brush_parser::*` as `brush_core::parser` just so I can use the AST types and whatnot. We could export only the stuff I need from `brush_parser`'s AST module, though there's so much that I don't see the point in simply not exporting the entire crate.

What do you think @reubeno?

brush-core/src/lib.rs
```diff
@@ -36,6 +36,7 @@ pub mod traps;
 pub mod variables;
 
 pub use arithmetic::EvalError;
+pub use brush_parser::ParseError;
```
reubeno, line 39: Apologies that this PR has gotten a bid wedged; let's try to find some initial step we can move forward with. Here's my incremental proposal for now:

```suggestion
/// Module containing the core parser facilities used and re-exported by this crate.
pub mod parser {
    pub use brush_parser::{
        ParseError, ParserOptions, SourcePosition, Token, TokenLocation, TokenizerError, ast,
    };
}
```

I think we can do better in the future, but I'm hoping this will give you a fair bit to go with. Do you think this will meet your needs for the time being, @hwittenborn?

reubeno:
@hwittenborn I'd like to get back to this, and expand it a bit. Are you interested in picking this back up, or are you okay if I shepherd it in? (For reference, please see my draft #784.) We'll make sure to credit you on the work regardless.

reubeno:
_Update:_ I've rebased your changes, made some additions, and have submitted them as a separate PR (#784). I'll close this PR.

PR #512, opened 2025-05-23
feat(AndOrList): add iteration abilities

### LINKED ISSUE #496: Implement Iter for brush_parser::ast::AndOrList
It'd be convenient to `iter()` over an `AndOrList` value!

Elsie19:
Fixes #496.

### DIFF
--- a/brush-parser/src/ast.rs
+++ b/brush-parser/src/ast.rs
@@ -1,7 +1,10 @@
 //! Defines the Abstract Syntax Tree (ast) for shell programs. Includes types and utilities
 //! for manipulating the AST.
 
-use std::fmt::{Display, Write};
+use std::{
+    collections::VecDeque,
+    fmt::{Display, Write},
+};
 
 use crate::tokenizer;
 
@@ -70,6 +73,30 @@ impl Display for AndOrList {
         }
 
         Ok(())
+    }
+}
+
+/// Iterator for [`AndOrList`].
+pub struct AndOrListIter {
+    list: VecDeque<AndOr>,
+}
+
+impl Iterator for AndOrListIter {
+    type Item = AndOr;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.list.pop_front()
+    }
+}
+
+impl IntoIterator for AndOrList {
+    type Item = AndOr;
+    type IntoIter = AndOrListIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        let mut list = vec![AndOr::And(self.first)];
+        list.append(&mut self.additional.clone());
+        AndOrListIter { list: list.into() }
     }
 }
 


brush-parser/src/ast.rs
```diff
@@ -73,6 +76,30 @@ impl Display for AndOrList {
...
+        self.list.pop_front()
+    }
+}
+
+impl IntoIterator for AndOrList {
+    type Item = AndOr;
+    type IntoIter = AndOrListIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        let mut list = vec![AndOr::And(self.first)];
```
reubeno: I think the first item in the list could be either `And` or `Or` (or `None` and use `Option` for that matter), but whichever path we take we should find a way to document to callers what to expect.

brush-parser/src/ast.rs
```diff
@@ -73,6 +76,30 @@ impl Display for AndOrList {
...
+    fn next(&mut self) -> Option<Self::Item> {
+        self.list.pop_front()
+    }
+}
+
+impl IntoIterator for AndOrList {
+    type Item = AndOr;
+    type IntoIter = AndOrListIter;
+
+    fn into_iter(self) -> Self::IntoIter {
```
reubeno, line 130: It's hard to know what the requested use case is--hence my question on the issue--but my thought is that most AST observers are probably immutably reading it vs. consuming it. We could certain use it ourselves internally.

What do you think about starting by adding `fn iter(&self) -> ...` on `AndOrList` instead of `IntoIterator` as the first step? If we wanted to keep it simple perhaps we could have it return `impl Iterator<Item = &AndOr>` and use something like `std::iter::once`, `.chain()`, etc. to stitch together the first element with the remaining ones? I haven't tried it myself so I'm not sure if there's issues with that approach...

Elsie19: I don't like the idea of making a method that shares the same name with one from a notable trait. Seems cheeky and deceptive to a caller + we can't implicitly use it in a for loop, which is likely what people want.

reubeno: Fair point. No deception desired.

I think what I'm looking for is a way to implement `IntoIterator` on `&AndOrList` to allow read-only iteration. Can you see a way to do that?

Elsie19: Yes, we can do something like that I think.

Elsie19: Actually, I don't think we can do non-consuming iterators unless they are cloned first, which makes sense, because pulling from an iterator gives the owned item, which makes it consuming by nature.

Elsie19:
Okay I added implementation notes that explain my reasoning. Anything else you need?

brush-parser/src/ast.rs
Elsie19: Actually my bad, you can do non-consuming iterators if the trait becomes `Iterator<&T>` as opposed to `Iterator<T>`, but `into_iter` still consumes it, meaning you have to `clone` it still.

reubeno:
Thanks for all of your work on this, and for your last comment about an iterator on `&T`. That's more what I had in mind. I do see that the definition of `AndOr` makes it tricky because it's hard to synthesize the `AndOr` for the first element without cloning it.

I experimented a bit and found something that could work (see below). **Pros** -- it avoids any cloning or consuming the input, which I'm hypothesizing as being more practically useful for most cases. **Cons** (maybe?) -- it required the iterator's `Item` being a pair that includes a new simple `enum`.

_(There's also some variants where we do the `.map()` in the the body of `.into_iter()`, etc.)_

```rust
/// Represents a boolean operator used to connect command pipelines in an `AndOrList`.
pub enum PipelineOperator {
    /// The command pipelines are connected by a boolean AND operator.
    And,
    /// The command pipelines are connected by a boolean OR operator.
    Or,
}

/// An iterator over the pipelines in an `AndOrList`.
pub struct AndOrListIter<'a> {
    first: Option<&'a Pipeline>,
    additional_iter: std::slice::Iter<'a, AndOr>,
}

impl<'a> Iterator for AndOrListIter<'a> {
    type Item = (PipelineOperator, &'a Pipeline);

    fn next(&mut self) -> Option<Self::Item> {
        if let Some(first) = self.first.take() {
            Some((PipelineOperator::And, first))
        } else {
            self.additional_iter.next().map(|and_or| match and_or {
                AndOr::And(pipeline) => (PipelineOperator::And, pipeline),
                AndOr::Or(pipeline) => (PipelineOperator::Or, pipeline),
            })
        }
    }
}

impl<'a> IntoIterator for &'a AndOrList {
    type Item = (PipelineOperator, &'a Pipeline);
    type IntoIter = AndOrListIter<'a>;

    fn into_iter(self) -> Self::IntoIter {
        AndOrListIter {
            first: Some(&self.first),
            additional_iter: self.additional.iter(),
        }
    }
}

impl AndOrList {
    /// Returns an iterator over the pipelines in this `AndOrList`.
    pub fn iter(&self) -> AndOrListIter<'_> {
        self.into_iter()
    }
}
```

Elsie19:
I added:

```rs
impl<'a> From<(PipelineOperator, &'a Pipeline)> for AndOr {
    fn from(value: (PipelineOperator, &'a Pipeline)) -> Self {
        match value.0 {
            PipelineOperator::Or => Self::Or(value.1.to_owned()),
            PipelineOperator::And => Self::And(value.1.to_owned()),
        }
    }
}
```

Which should make converting iterator elements into `AndOr` easier, with the caveat that it has to clone the values.

reubeno:
Looks good to me, thanks for the updates. I'll merge this later today.

PR #531, opened 2025-05-29
feat(env): introduce brush specific variable

### LINKED ISSUE #530: Proposal: Provide a brush-specific variable
## Summary:

Add `$BRUSH_VERSION` or another global shell variable to indicate that the shell is brush.

## Motivation:

I want this feature for the gradual adoption of brush, as a workaround for compatibility issues such as #380. Until these compatibility issues are resolved, loading the same `.bashrc` file generates warnings or errors.
We can avoid this behavior by putting a special configuration file for brush; however, managing shared and brush-specific configurations is often inconvenient.
On the other hand, ignoring the whole configuration with `--norc` or `--noprofile` is also not desirable for a daily driver.
So, initially, I added a condition to skip unsupported features with `! command -v brushinfo`. However, this approach often significantly slows down original bash startup.
I'm not sure why, but I suspect it's due to the much longer `$PATH` in my use case (NixOS and/or home-manager. Windows also injects `$PATH` in WSL2). Actually, my bash startup speed increased to 140ms from 90ms.

## Justification:

As far as I know, other famous shells have similar features.

* bash: `$BASH_VERSION`
* zsh: `$ZSH_VERSION`
* fish: `$FISH_VERSION`
* nushell: `$env.NU_VERSION`
* powershell: `$PSVersionTable`
* rusty_bash: This is an exception. Sush does not have a special version variable and [returns its own version via `$BASH_VERSION`](https://github.com/shellgei/rusty_bash/blob/1d34a3145a27eb8092763eb53b7b691a405ea3b5/src/core.rs#L124-L135).

Currently, brush seems to return `$BASH_VERSION`, but this is not its own version. I think keeping the current `$BASH_VERSION` and providing a brush-specific variable like `$BRUSH_VERSION` will be useful for gradual adoption.

kachick:
Resolves #530

I wasn't sure how to add appropriate tests for this feature. For now, I've ensured that if there's a major mistake, the integration tests will fail early.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -189,8 +189,8 @@ pub struct CreateOptions {
     pub max_function_call_depth: Option<usize>,
     /// Key bindings helper for the shell to use.
     pub key_bindings: Option<KeyBindingsHelper>,
-    /// Brush shell version.
-    pub brush_shell_version: String,
+    /// Brush implementation version.
+    pub shell_version: String,
 }
 
 /// Represents an executing script.
@@ -302,9 +302,9 @@ impl Shell {
                 self.env.set_global(k, var)?;
             }
         }
-        let brush_version = options.brush_shell_version.clone();
+        let shell_version = options.shell_version.clone();
         self.env
-            .set_global("BRUSH_VERSION", ShellVariable::new(brush_version.into()))?;
+            .set_global("BRUSH_VERSION", ShellVariable::new(shell_version.into()))?;
 
         // TODO(#479): implement $_
 
@@ -414,6 +414,7 @@ impl Shell {
         self.env.set_global("BASH_VERSINFO", bash_versinfo_var)?;
 
         // BASH_VERSION
+        // This is the Bash interface version. See BRUSH_VERSION for its implementation version.
         self.env.set_global(
             "BASH_VERSION",
             ShellVariable::new(
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -190,7 +190,7 @@ pub struct CreateOptions {
     /// Key bindings helper for the shell to use.
     pub key_bindings: Option<KeyBindingsHelper>,
     /// Brush implementation version.
-    pub shell_version: String,
+    pub shell_version: Option<String>,
 }
 
 /// Represents an executing script.
@@ -303,8 +303,10 @@ impl Shell {
             }
         }
         let shell_version = options.shell_version.clone();
-        self.env
-            .set_global("BRUSH_VERSION", ShellVariable::new(shell_version.into()))?;
+        self.env.set_global(
+            "BRUSH_VERSION",
+            ShellVariable::new(shell_version.unwrap_or_default().into()),
+        )?;
 
         // TODO(#479): implement $_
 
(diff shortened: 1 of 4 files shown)


brush-core/src/shell.rs
```diff
@@ -189,6 +189,8 @@ pub struct CreateOptions {
     pub max_function_call_depth: Option<usize>,
     /// Key bindings helper for the shell to use.
     pub key_bindings: Option<KeyBindingsHelper>,
+    /// Brush shell version.
+    pub brush_shell_version: String,
```
reubeno: For consistency with `shell_name` and `shell_product_display_str`, what do you think about going with `shell_version` as the name of the field and `Option<String>` as the type?

We can clarify in the comment that it's intended to be implementation version.

brush-shell/tests/compat_tests.rs
```diff
@@ -36,10 +36,19 @@ struct TestConfig {
 
 impl TestConfig {
     pub fn for_bash_testing(options: &TestOptions) -> Result<Self> {
-        // Check for bash version.
+        // Check for bash and brush version.
```
reubeno: I appreciate you looking for a way to add an integration test. I don't think the compat test suite is quite the right place, since it's really focused on comparing behavior between across shell implementations -- and as you identified, this is really a feature that's unique to `brush`.

We don't have as many such integration tests as we ought to, but the [interactive tests](https://github.com/reubeno/brush/blob/main/brush-shell/tests/interactive_tests.rs) we have could be a good inspiration, but we wouldn't need the expect-style logic.

We could, for instance:

* Add a new `integration_tests.rs` next to `interactive_tests.rs`
* Use `assert_cmd` (like `interactive_tests.rs`) to spawn a `brush` instance, e.g.:

  https://github.com/reubeno/brush/blob/3fdb482cd595b978146cb41362b9a10c1a5852e7/brush-shell/tests/interactive_tests.rs#L155-L164

* Go with the `-c 'echo ${BRUSH_VERSION}'` approach you took here.
* Minimally observe the returned version.

Looking at the docs of `assert_cmd`, I also see helpers that could make it even simpler (e.g., [CommandCargoExt](https://docs.rs/assert_cmd/latest/assert_cmd/cargo/trait.CommandCargoExt.html)).

In time, we can build on this to properly test `brushinfo` et al. (It's an oversight that there are no tests for it.)

brush-core/src/shell.rs
kachick: Sounds good! 👍 Applied the renaming and adding comments in 002d182324170458301936b350a3cc4e58abb644

brush-shell/tests/compat_tests.rs
kachick: Thank you so much for the guidance. 🙏  It really helped me extract the tests in an appropriate way. 98db07507bb11236a9218c6d596c83376cc7efda

reubeno: The new test looks great, thanks for putting in the time to make the change!

brush-core/src/shell.rs
reubeno: Thanks for the rename; what do you think about `String` -> `Option<String>` as I suggest above? We can simply skip setting `BRUSH_VERSION` if the caller constructing the `Shell` chose not to fill it in.

kachick: Oops, sorry for overlooking the type change suggestion 🙇‍♂️ 
I also agreed the direction 2014fb7178675290e2d4a8fe9f3ed9aaf9c8a82c

reubeno: No problem, the update looks good.

reubeno (approved):

PR #539, opened 2025-06-01
fix(interactive): fix behavior of cmds piped to stdin

### LINKED ISSUE #538: Error when code is piped to stdin of brush
Shells should support code read from stdin:

```
$ echo 'echo hi' | bash
hi
```

brush:

```
$ echo 'echo hi' | ./brush

[4]+  Stopped                 echo 'echo hi' | ./brush
```

I'm using the latest musl binary on an Ubuntu VM:

```
$ ./brush --version
brush 0.2.18 (git:cfa8c5a)
```

reubeno:
This addresses the issues we're seeing when commands are piped into `brush`'s `stdin`. 

Note: this is incremental progress. There are still some issues with how `-i` interacts and we need to ensure we find a reasonable approach to getting back some of the coverage of the reedline impl of `InteractiveShell` that will be lost with this change. We also need to add variants of some of the "basic" tests running with `stdin` being a terminal (say, via pty). We'll sort those out in subsequent PRs.

Resolves #538

### DIFF
--- a/brush-interactive/src/interactive_shell.rs
+++ b/brush-interactive/src/interactive_shell.rs
@@ -1,5 +1,5 @@
 use crate::ShellError;
-use std::io::Write;
+use std::io::{IsTerminal, Write};
 
 /// Result of a read operation.
 pub enum ReadResult {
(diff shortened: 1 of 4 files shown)


reubeno:
I did some research and found that there's a previous report of this against `reedline` itself: https://github.com/nushell/reedline/issues/509.

For now, we'll proceed with a high-level switch not to even use `reedline` if we know that `stdin` isn't a TTY. This may have other ramifications down the road that we'll need to deal with (e.g., needs for `history` management when opted-in even when `stdin` isn't a TTY). We'll need to cross that bridge when we come to it.

PR #548, opened 2025-06-08
fix(getopts): honor `OPTERR`

Elsie19:


Elsie19:
## With Bash

```bash
#!/bin/bash

OPTERR=1

while getopts "a:b:" opt; do
    case "$opt" in
    a)
        echo "You gave -a with value: $OPTARG"
        ;;
    b)
        echo "You gave -b with value: $OPTARG"
        ;;
    \?)
        echo "Unknown option: -$OPTARG"
        ;;
    :)
        echo "Missing argument for -$OPTARG"
        ;;
    esac
done
```

```bash
$ bash test.sh -a foo -x -b
You gave -a with value: foo
test.sh: illegal option -- x
Unknown option: -
test.sh: option requires an argument -- b
Unknown option: -
```

## With Brush

```bash
$ ./target/debug/brush test.sh -a foo -x -b
```

Is getopts in brush fleshed out at all or is it just a pretty basic implementation as of now?

reubeno:
It looks like there's multiple issues going on.

* `OPTERR` isn't being observed, but in this case, we still should have defaulted to displaying errors anyway. Regarding this PR's changes, could you validate if `== 1` is the right check vs `!= 0`? `bash`'s help for `getopts` mentions "nonzero".

* `getopts` in `brush` doesn't look like it does the right thing when no arguments are given (and it's expected for it to look at positional arguments to the shell/script/function). When `self.args` is empty, `context.shell.positional_parameters` should be used instead.

Your test case above is a good one; you should be able to add it to `getopts.yaml` to test that everything is working as expected (although there's a risk that our error messages may not 100% line up). May be easiest to put it all in a helper function (instead of a script) that gets invoked...

reubeno:
Thanks for getting the ball rolling with these changes! I used the changes and test you posted above as the initial inspiration for #1048. We may have more work to do with `getopts`, but I think it should be in a better place now.

PR #558, opened 2025-06-14
feat(mapfile): add `-O` flag

Elsie19:


brush-core/src/builtins/mapfile.rs
```diff
@@ -66,14 +61,20 @@ impl builtins::Command for MapFileCommand {
...
 
-        // Assign!
-        context.shell.env.update_or_add(
-            &self.array_var_name,
-            variables::ShellValueLiteral::Array(results),
-            |_| Ok(()),
-            env::EnvironmentLookup::Anywhere,
-            env::EnvironmentScope::Global,
-        )?;
+        for (elem_idx, result) in results.0.into_iter().enumerate() {
```
reubeno: When `-O` isn't used and the array is sufficiently large, a single `.update_or_add()` call may be more efficient, but I do like the simplicity of having a common code path for the `-O` case and non-`-O` case.

Assuming the semantics work out for all the cases we care about, I'm okay with this approach.  We can always revisit down the road.

brush-core/src/builtins/mapfile.rs
```diff
@@ -66,14 +61,20 @@ impl builtins::Command for MapFileCommand {
...
+            // If the user is getting to wraparounds in *bash*, they got bigger problems.
+            #[allow(clippy::cast_possible_wrap)]
+            let elem_idx = elem_idx as i64;
+            // Assign!
+            context.shell.env.update_or_add_array_element(
```
reubeno: Thanks for tackling this option. I think we're going to need a wide range of compat tests in `mapfile.yaml`, run both with `-O` and without it--and with different values passed to `-O`, e.g.:

* Variable doesn't already exist before call.
* Variable exists but is a non-array, plain string variable.
* Variable exists and is a regular indexed array.
* Variable exists and is an *associative* array.
* `-O` is passed a negative value

I'm sure there may be other interesting cases, but without the tests it's hard to be certain whether the implementation of `.update_or_add_array_element()` will match the required behavior here--particularly with respect to variable conversion.

brush-core/src/builtins/mapfile.rs
Elsie19: For reference, this is how bash's documentation describes `mapfile` updates/inserts (emphasis mine):

> When callback is evaluated, it is supplied the index of the next array element to be assigned and the line to be assigned to that element as additional arguments. callback is evaluated **after the line is read but before the array element is assigned.**

This is from the `-c` flag but it does tell us that `mapfile` loops through and inserts, rather than all in one go.

---

So we are doing half of bash's implementation right, but to get more in line with theirs, we would need to read and assign together, but now we are doing reading, then assigning.

Elsie19:
I don't think I have the time to come up with a test suite for what you asked for so I will defer this to someone else if they want to add tests.

reubeno:
> I don't think I have the time to come up with a test suite for what you asked for so I will defer this to someone else if they want to add tests.

FYI -- I'm going to rebase and start adding some of those tests. I don't know if I have permissions to force-push to your PR branch so I might need to create a separate PR, but will retain your author tag on the changes.

PR #575, opened 2025-06-21
Winnow prompt parser

lu-zero:
Example of the usage of winnow


lu-zero:
After trying a bit more I consider winnow (paired with miette to report the span in an uniform way that is not ad-hoc) to be the current best candidate:
- It is much easier to debug
- It is better documented
- The ~~`precedence`~~ `expression` combinator has implementations and even if it doesn't land immediately in the next week we can either lift the PR locally or spin it as a small crate.

reubeno:
> After trying a bit more I consider winnow (paired with miette to report the span in an uniform way that is not ad-hoc) to be the current best candidate:
> 
> * It is much easier to debug
> * It is better documented
> * The `precedence` combinator has implementations and even if it doesn't land immediately in the next week we can either lift the PR locally or spin it as a small crate.

Thanks for the recommendation -- I'm onboard with this. It's great also seeing the energy going into `winnow` development; that makes it feel like a good bet.

And agreed on the `precedence` bit. I'd love to see it properly supported in the main `winnow` project, but it sounds like we don't have to block on that.

Thanks again for continuing to push this along and help us plot a course--I really appreciate it :smile:. I'll look more closely at these changes over the next couple of days.

lu-zero:
I tried to revive the the PR for the `expression` (current name agreed) combinator, meanwhile do you want me to add the switch machinery so we can select which flavor of parser to use at runtime/compile time?

reubeno:
I'm going to convert this back to draft so we don't lose it, but acknowledge that our eye is on the prize of the *real* parser :smile:.

PR #596, opened 2025-07-01
feat(api): API usability improvements for `Shell::invoke_function`

reubeno:
Gives some love to `Shell::invoke_function`:

* Makes it easier to use with input strings of various forms.
* Requires supplying custom `ExecutionParameters`, allowing per-invocation redirection.
* Adds an example demonstrating how to invoke a function, with optional redirection.

Also properly exports `OpenFiles` and `OpenFile`, which required some cleanup and adding some missing wrappers.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -950,10 +950,10 @@ impl Shell {
     /// * `name` - The name of the function to invoke.
     /// * `args` - The arguments to pass to the function.
     /// * `params` - Execution parameters to use for the invocation.
-    pub async fn invoke_function(
+    pub async fn invoke_function<N: AsRef<str>, I: IntoIterator<Item = A>, A: AsRef<str>>(
         &mut self,
-        name: impl AsRef<str>,
-        args: impl Iterator<Item = impl AsRef<str>>,
+        name: N,
+        args: I,
         params: &ExecutionParameters,
     ) -> Result<u8, error::Error> {
         let name = name.as_ref();
@@ -973,6 +973,7 @@ impl Shell {
         };
 
         let command_args = args
+            .into_iter()
             .map(|s| commands::CommandArg::String(String::from(s.as_ref())))
             .collect::<Vec<_>>();
 


reubeno:
@Elsie19 -- since you were asking about this, do you have a moment to look at this, or part of it? At least (1) the (admittedly breaking) API change to `invoke_function` in `shell.rs`, and (2) the new example that demonstrates how it can be used with redirection?

With these changes, it's also possible to apply custom redirection to the whole `Shell` too -- by reaching into the newly exported `open_files` field of the `Shell` instance similar to what this example does with the `ExecutionParameters`'s `open_files`.

brush-core/src/shell.rs
```diff
@@ -949,10 +949,14 @@ impl Shell {
...
     /// * `args` - The arguments to pass to the function.
-    pub async fn invoke_function(&mut self, name: &str, args: &[&str]) -> Result<u8, error::Error> {
-        // TODO: Figure out if *all* callers have the same process group policy.
-        let params = self.default_exec_params();
-
+    /// * `params` - Execution parameters to use for the invocation.
+    pub async fn invoke_function(
+        &mut self,
+        name: impl AsRef<str>,
+        args: impl Iterator<Item = impl AsRef<str>>,
```
Elsie19: This breaks things such as `&[]` being passed and makes it really difficult to pass things to, so I would suggest changing it to `IntoIterator`.

Elsie19:
```suggestion
    pub async fn invoke_function<S: AsRef<str>>(
        &mut self,
        name: S,
        args: impl Iterator<Item = S>,
```

reubeno: _I vaguely recall having this conversation before, but would appreciate if you could remind me of your perspective on this, @Elsie19?_

As far as I understand it, adding a parameter of type `impl T` for some trait `T` is basically equivalent to:

* Adding a generic type parameter `P`
* Marking the parameter as being of type `P`
* Adding a constraint that `P : T`

_(Minor note -- in the suggestion made here, sharing `S` between `name` and the `Item` of `args` constrains the two string-ish arguments to use the same `AsRef` string type, which isn't strictly required. For more flexibility we'd need a separate generic type parameter for each.)_

From what I gather, `impl` returns weren't supported on older toolchains, and there may be constraints around using it in certain context. It also somewhat hides the type parameters.

My own take is that avoiding `impl` returns in a function like this (which would require adding 3 type parameters for `name`, for `args`, and for the `Item` of `args`) complicates the function signature and distracts the reader a bit--particularly when the generics are just so we can take collections of objects/strings in a more ergonomic way. If we were really doing something more fundamental with the generic parameters, such as a new kind of sort function that is designed to be generic over any kind of comparable type, then I think I'd feel differently and want to see the generic type parameter written out properly.

Ultimately consistency is probably more important, as this is really a stylistic thing (and I know we _have_ already gone with the other approach elsewhere in this source file), but I think it's worthwhile to formulate an intentional opinion on this so I can write it down in our docs somewhere and just refer to that later.

Elsie19:
```rs
fn foo<T: Trait>(f: T) {}; // And by extension, the equivalent `where` clause.
```

and

```rs
fn bar(f: impl Trait) {};
```
are actually subtly different:

```rs
foo::<u32>(0) // this is allowed
bar::<u32>(0) // this is not
```

I also agree that it does hide the type parameters and I usually only use them when there are very few params.

---

I would use `where` clauses when there are intertwined generics, such as this very contrived example:

```rs
fn process_things<'a, T, U, F, G, H>(
    input: &'a [T],
    transform: F,
    filter: G,
    finalizer: H,
) -> Vec<U>
where
    T: Clone + PartialEq + 'a,
    U: Default + Clone,
    F: Fn(&'a T) -> U,
    G: Fn(&T) -> bool,
    H: Fn(U) -> U,
{
    input
        .iter()
        .filter(|x| filter(*x))
        .map(|x| {
            let val = transform(x);
            finalizer(val)
        })
        .collect()
}
```

As opposed to (get ready 😬):

```rs
fn process_things<
    'a,
    T: Clone + PartialEq + 'a,
    U: Default + Clone,
    F: Fn(&'a T) -> U,
    G: Fn(&T) -> bool,
    H: Fn(U) -> U,
>(
    input: &'a [T],
    transform: F,
    filter: G,
    finalizer: H,
) -> Vec<U> {
    input
        .iter()
        .filter(|x| filter(*x))
        .map(|x| {
            let val = transform(x);
            finalizer(val)
        })
        .collect()
}
```

Or even worse:

```rs
fn process_things<'a, T: Clone + PartialEq + 'a, U: Default + Clone>(
    input: &'a [T],
    transform: impl Fn(&'a T) -> U,
    filter: impl Fn(&T) -> bool,
    finalizer: impl Fn(U) -> U,
) -> Vec<U> {
    input
        .iter()
        .filter(|x| filter(*x))
        .map(|x| {
            let val = transform(x);
            finalizer(val)
        })
        .collect()
}
```

#2 may look nicer but remember from the perspective of reading docs, these are how they look:

## #1
image

## #2
image

## #3
image

reubeno: Thanks for the reminder about the turbofish operator difference and the suggestions! I'm going to lean toward being consistent with the other functions already present in `shell.rs` but reserve the right to take another pass at it later :smile:.

Elsie19: Oh and I forgot to say this, but I actually like impl returns when possible. I don't think I need to remind you but your function inputs should be as narrow as possible without being concrete types, and your function outputs should be as wide as possible, so my example *could have* been:

```rs
fn process_things<'a, T, U, F, G, H>(
    input: &'a [T],
    transform: F,
    filter: G,
    finalizer: H,
) -> impl Iterator<Item = U> + 'a
where
    T: Clone + PartialEq,
    U: Default + Clone,
    F: Fn(&'a T) -> U + 'a,
    G: Fn(&T) -> bool + 'a,
    H: Fn(U) -> U + 'a,
{
    input.iter().filter(move |x| filter(*x)).map(move |x| {
        let val = transform(x);
        finalizer(val)
    })
}
```

Which I think is even better than what I showed, because you can even now see the flow and understanding of how generics and lifetimes interact, because you can essentially just go top to bottom and realize a couple things:

1. All functions passed in are constrained to the lifetime of the input.
2. The return type is constrained to the input lifetime.
3. Since `H` is a function that takes a `U` and returns a `U`, it must be constrained to the output as well.

None of these things are easily noticeable in the #2 example where generics are squished into the function parameter, and even worse, clutter the `impl Trait` syntax.

But I do see that this isn't possible all the time, and 80% of the time, you must return concrete types.

Elsie19: Also I will say, you should almost never have a parameter that takes an `Iterator`, it should almost always be `IntoIterator`.

PR #604, opened 2025-07-05
Arch Linux - update installation instruction

roland-5:


reubeno (approved):
Thanks for your contribution! It's been great seeing `brush` make it into `extra` and now more easily installable on Arch.

PR #614, opened 2025-07-25
feat(config): Use platform-specific standard directories for config files, state files

varbhat:
Uses https://crates.io/crates/directories to get config directory of user and loads the brush's config file from that directory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom",
+ "getrandom 0.3.3",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -289,9 +289,10 @@ dependencies = [
  "clap",
  "command-fds",
  "criterion",
+ "directories",
  "fancy-regex",
  "futures",
- "getrandom",
+ "getrandom 0.3.3",
  "homedir",
  "hostname",
  "indexmap",
@@ -335,7 +336,7 @@ dependencies = [
  "brush-core",
  "brush-parser",
  "crossterm 0.29.0",
- "getrandom",
+ "getrandom 0.3.3",
  "indexmap",
  "nu-ansi-term 0.50.1",
  "reedline",
@@ -383,7 +384,7 @@ dependencies = [
  "descape",
  "diff",
  "expectrl",
- "getrandom",
+ "getrandom 0.3.3",
  "git-version",
  "glob",
  "human-panic",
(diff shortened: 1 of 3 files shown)


reubeno:
Thanks for the contribution! This looks like it will set us up for better config management.

I'll look at the code details a bit more closely after the checks run, but a few quick comments from first skim:

* For use of `.bashrc` we should stay compatible with bash's behavior and look under `$HOME`. For `brush`-specific paths, we have room to adopt more standardized paths.

* What do you think about breaking out some helper functions for composing the config paths into a separate module, say a config.rs? I can see us making more use of them in the future.

* I've been debating whether `brush-core` or `brush-shell` crates should be composing the config paths under these standard dirs. For another application that builds on `brush-core`, I can see the value of it specifying its own config/state dir. I'd be okay with this being done in a future set of changes, though.

reubeno:
It looks like `cargo deny` is tripping over the MPL2.0 license of one of the used crates. The license should be fine for us to use but we'll need to update our `deny.toml` and make sure we bundle the right notices when we redistribute our binaries. I can work with you on the former, and we can resolve the latter in a subsequent PR (but before we next release).

reubeno:
> Please do not use it. It breaks horribly support for sharing the home between linux and macos.

@lu-zero -- could you share more about your concerns here? Is your concern with the specific crates in use here? Or the way they're being used?

lu-zero:
The crate is highly opinionated and makes impossible to be sensible and check first for XDG/unix paths first and fail over to MacOS paths.

command line tools work best if they use the above mentioned strategy and given we want to stay compatible with bash I'd find quite peculiar looking for a bashrc in the windows registry or the macos path only.

lu-zero:
https://codeberg.org/dirs/directories-rs/issues/47 for the background problem

https://github.com/lunacookies/etcetera would make things nicer, but still, I see very little gain in diverging from unix-style for this project.

reubeno:
> https://codeberg.org/dirs/directories-rs/issues/47 for the background problem
> 
> https://github.com/lunacookies/etcetera would make things nicer, but still, I see very little gain in diverging from unix-style for this project.

Thanks for the additional context. I'd like to look into `etcetera`, then.

Regarding `bash` compatibility, I 100% agree that we need to support the same paths as `bash` (e.g., `$HOME/.bashrc`). For *brush*-specific files, I can see us doing something like:

* First look under "appropriate" config dirs for `brushrc` (including under `~/.config/brush` on appropriate platforms); fall back to `~/.brushrc` if not found. (Or vice versa.)
* (Future) Support us storing cached data under appropriately configured dirs (e.g., `~/.cache/brush`), say in case we end up finding it a perf win/useful.
* (Future) If we find a need for other config that isn't expressible in bash syntax itself (i.e., can't be expressed in a `brushrc`) support loading that config from an appropriately configured dir.

What's your take on this sort of scoping, @lu-zero ?

lu-zero:
I'd suggest to use [xdg](https://crates.io/crates/xdg) and avoid having special/os-specific cases.

It _might_ make sense to use etcetera only to use the os-specific cache dir to be extra nice to newcomers so the cache would have higher chances to be excluded from backups, but that's a secondary refinement and the library choice would be secondary as long as we have `.brushrc` and friends as primary lookup and then always XDG.

theoparis:
I agree that following the xdg specification would be nice. However the [xdg crate doesn't support windows](https://github.com/whitequark/rust-xdg/issues/28).

lu-zero:
If you are using bash on windows there are very high odds you are on msys2, the git for windows distribution of bash and/or you have the XDG variables set to something proper.

PR #624, opened 2025-08-07
fix(parser): resolve issue with parser confusing subshell for arith expr

reubeno:
Consider the case of `( : && ( (( 0 )) || : ) )`:

* The 3 consecutive open parens `(` tokenize as 3 repeated operators.
* Visually, with space, it looks like it's intended to be an arithmetic command inside a subshell.
* With space removed, when we see 2 `(` chars, we try to parse an arithmetic command.
* When we see the unmatched `)`, the existing logic keeps going on and includes it in the parsed arithmetic expression.

We should instead recognize that this is a subshell command--the unmatched closing parenthesis should have made that obvious.

### DIFF
--- a/brush-parser/src/parser.rs
+++ b/brush-parser/src/parser.rs
@@ -331,8 +331,13 @@ peg::parser! {
             raw_expr:$(arithmetic_expression_piece()*) { ast::UnexpandedArithmeticExpr { value: raw_expr } }
 
         rule arithmetic_expression_piece() =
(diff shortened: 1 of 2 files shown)


reubeno:
@theoparis FYI -- I *think* this addresses the issue you brought up with `makepkg`. (It's a legitimate issue, just wasn't able to confirm that it resolves the specific instance you're seeing.)

PR #625, opened 2025-08-09
fix(expansion): support broader set of nested brace expansions

reubeno:
We previously supported expanding brace expressions like `{a,{x,y}}` but did *not* support something like `{a,b{x,y}}`. This addresses that, adds a few targeted tests, and consolidates the expansion code back into `brush_core`. (Some of it had been in `brush_parser`.)


reubeno:
@theoparis I expect this should address some of the errors you were seeing with braces. Holler if you see more problems with those scripts...

PR #648, opened 2025-08-16
Add minimal miette support

lu-zero:
Ideally we could replace the current `ParseError` and possibly reuse the miette structures that duplicate functionality through newtyping or type aliases later.


lu-zero:
why those 404, github is having more networking troubles?

reubeno:
> why those 404, github is having more networking troubles?

The zizmor failure does look like some kind of transient networking issue accessing GitHub. I've queued a rerun of that step.

The failing deny check is legit; we'll need to update `deny.toml` to allow the ISC license. That seems okay to me.

reubeno:
Looking at the miette overview page, it says the "fancy" feature may only need to be turned on in the top level crate (`brush-shell` in this case). Did I understand the correctly?

lu-zero:
yes, I enabled it only for the example in the brush-parser, the next target would be `brush-core` and possibly extend it so we track the files across source.

deny.toml
```diff
@@ -39,6 +39,7 @@ allow = [
     "Apache-2.0",
     "BSD-2-Clause",
     "BSL-1.0",
+    "ISC",
```
reubeno, line 42: Thanks for adding this. We haven't yet documented this, but could you please also add `ISC` to `brush-shell/about.toml`? We recently added usage of `cargo about` to generate license notices that we can bundle with our prebuilt binaries.

brush-parser/Cargo.toml
```diff
@@ -20,11 +20,13 @@ bench = false
 [features]
 fuzz-testing = ["dep:arbitrary"]
 debug-tracing = ["peg/trace"]
+fancy-diagnostics = ["dep:miette"]
```
reubeno: Thanks for clarifying that this is basically just for the example. Would you mind adding a comment here explaining that this is generally not needed, and instead the consuming app can/should enable it itself?

brush-parser/src/error.rs
```diff
@@ -22,6 +22,49 @@ pub enum ParseError {
     },
 }
 
+#[cfg(feature = "fancy-diagnostics")]
```
reubeno: Is there a reason to put this code behind `fancy-diagnostics`? Is there actually any code in this `mod` that depends on `miette`'s fancy feature (as opposed to just using its structs/types)?

brush-parser/Cargo.toml
lu-zero: This is to make miette optional, so people that do not want to depend on miette they can just use the support provided by thiserror alone.

brush-parser/src/error.rs
lu-zero: `fancy-diagnostics` is a feature for brush-parser, miette/fancy is needed for the example that show how to use brush-parser with miette and the standard fancy visualization.

brush-parser/Cargo.toml
reubeno: Thanks for the clarification; I think what I was stumbling on is that `fancy-diagnostics` doesn't actually enable `miette`'s `fancy` feature.

Would you be open to renaming the feature to something that actually names `miette` or else doesn't use the term `fancy` to avoid confusion with `miette`'s `fancy` feature?

lu-zero: Any name is good to me :) even `diagnostics`

PR #651, opened 2025-08-17
Add a ShellBuilder

lu-zero:


reubeno:
Thanks for throwing this together! Do you have a quick snippet example of what it would look like to use? Or what the generated rust docs look like?

lu-zero:
image
image
image
image

brush-core/src/shell.rs
```diff
@@ -150,50 +150,94 @@ impl AsMut<Self> for Shell {
...
+        name = build_settings,
+        vis = "pub(crate)",
+    ),
+    start_fn(
+        vis = "pub(crate)"
+    )
+)]
 pub struct CreateOptions {
     /// Disabled options.
+    #[builder(default)]
```
reubeno: Looking back through this `struct`, many of the names and types are appropriate for the fields but with a builder pattern might not be the ones we'd typically choose.

For example, with `enabled_options` if we were implementing a builder by hand, I'd probably add a func called `enable_option()` that takes a single `impl AsRef<str>` or similar rather than a single func called `enabled_options` that takes a `Vec<String>`.

Is there a way to customize some of them to make them feel more natural?

While we're at it, do you think there are any other names that you'd recommend we rename for clarity?

brush-core/Cargo.toml
```diff
@@ -19,6 +19,7 @@ workspace = true
 [dependencies]
 async-recursion = "1.1.1"
 async-trait = "0.1.89"
+bon = "3.7.0"
```
reubeno: FYI -- since it took me so long to review this 🥲 there's now a version `3.7.1` out...

brush-core/src/shell.rs
```diff
@@ -150,50 +150,94 @@ impl AsMut<Self> for Shell {
...
+#[derive(Default, bon::Builder)]
+#[builder(
+    builder_type(
+        name = ShellBuilder,
+        doc {
+        /// Builder for [Shell]
+    }),
+    finish_fn(
+        name = build_settings,
+        vis = "pub(crate)",
```
reubeno: I haven't read too much on the `bon` docs; do these functions need to be crate-public vs. module-private?

brush-core/src/shell.rs
```diff
@@ -222,6 +266,11 @@ pub struct FunctionCall {
 }
 
 impl Shell {
+    /// Create an instance of [Shell] using the builder syntax
+    pub fn builder() -> ShellBuilder<shell_builder::Empty> {
```
reubeno, line 321: `Shell` is already somewhat of a dumping ground; what do you think about omitting this and intentionally exposing a builder-specific utility type, or else expose it on the create options?

brush-core/src/shell.rs
lu-zero: > Looking back through this `struct`, many of the names and types are appropriate for the fields but with a builder pattern might not be the ones we'd typically choose.
> 
> For example, with `enabled_options` if we were implementing a builder by hand, I'd probably add a func called `enable_option()` that takes a single `impl AsRef<str>` or similar rather than a single func called `enabled_options` that takes a `Vec<String>`.
> 
> Is there a way to customize some of them to make them feel more natural?

~~[Yes](https://bon-rs.com/reference/builder/member/name), bon is quite flexible.~~ Misread the question, the answer is [almost yes](https://github.com/elastio/bon/issues/190).

> 
> While we're at it, do you think there are any other names that you'd recommend we rename for clarity?

I'll go through it this weekend then :)

brush-core/src/shell.rs
lu-zero: We can use `pub(self)` IIRC.

brush-core/src/shell.rs
lu-zero: The builder pattern is usually `T::builder() -> TBuilder`, `TBuilder::build() -> T`, to maintain the compatibility I wrapped around `CreateOptions` instead of replacing it. 

Ideally if we are happy with the builder pattern `CreateOptions` can be phased out from the API.

brush-core/src/shell.rs
lu-zero: Manually implemented.

lu-zero:
since now it is green, can we land it?

PR #660, opened 2025-08-26
chore: Update MSRV to 1.88

twz123:
When compiling for Windows, the homedir crate uses features that have been stabilized on 1.88, so let the MSRV match that. Also run the source code checks on Windows, so that those things will be caught by it in the future.

This is the error:

    error[E0658]: `let` expressions in this position are unstable
       --> C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\homedir-0.3.5\src\windows.rs:203:16
        |
    203 |             if let Err(e) = GetTokenInformation(token_handle, TokenUser, None, 0, &mut buffer_size)
        |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        |
        = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information

    For more information about this error, try `rustc --explain E0658`.
    error: could not compile `homedir` (lib) due to 1 previous error

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -295,12 +295,14 @@ jobs:
   # Static analysis of the code.
   check:
     name: "Source code checks"
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
         # Test latest stable as well as MSRV.
-        rust-version: ["stable", "1.85.0"]
+        rust-version: ["stable", "1.88.0"]
+        # Test Linux and Windows.
+        os: ["ubuntu-24.04", "windows-2025"]
 
     steps:
       - name: Checkout
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ keywords = ["cli", "shell", "sh", "bash", "script"]
 license = "MIT"
 readme = "README.md"
 repository = "https://github.com/reubeno/brush"
-rust-version = "1.85.0"
+rust-version = "1.88.0"
 
 [workspace.lints.rust]
 warnings = { level = "deny" }


reubeno:
Thanks for the contribution, @twz123! This was a good catch.

We'd intentionally held off upgrading `homedir` to 0.3.6 because it lists its MSRV as 1.88 but didn't realize that 0.3.5 already requires that version (without advertising it).

We're generally relatively conservative about upgrading MSRV for our project unless there's sufficient return for doing so.

What do you think about **downgrading** `homedir` back to 0.3.4 for the time being. I didn't look deeply but my quick read is that should be able to work with our current MSRV. One way or the other, though, your expansion of the source code checks in CI should validate it.

reubeno:
Just saw the check errors. The test failures look unrelated, possibly from flaky tests. The source code check failure *could* be related to your original MSRV change, but I won't swear to that.

reubeno:
@twz123 -- thanks again for your contribution! I'm assuming you haven't had time to follow up, so I've incorporated your changes into #678 but will stick to the current MSRV along with the downgraded `homedir` crate (see comments above). Once I merge that, I'll close this PR. Please do reach out if you have any additional input or feedback.

reubeno:
#678 has now been merged. If we find additional issues on the Windows platform, we'll open up separate issues/PRs. Thanks!

twz123:
Thanks for taking care of this! I was trying to make brush work on Windows as a bash replacement to be used with direnv. Turns out that brush doesn't like direnv's stdlib, which I tried to address. But, of course, this is a can of worms :see_no_evil: and there's only so much time I can dedicate to this. No idea if I get along to upstream my changes. Maybe I'll just open a draft PR with my WIP stuff for reference.

reubeno:
Of course, @twz123; thanks for getting things started.

I'm intrigued to learn more about the issues you ran into with `direnv`. You're definitely welcome to publish a draft PR with your work-in-progress or alternatively an issue with a link to a branch in your fork.

PR #674, opened 2025-09-07
RFC/PoC: JSON-RPC API for Brush

bolinfest:
My high-level goal is to wrap a shell with a programmatic API to enable introspection and automation.

This PR introduces a PoC to show how this could work. If the maintainers open to supporting this functionality, I am happy to take direction and clean this up further.

Overview:

- This adds a `--rpc-socket` flag to specify the path to a UNIX domain socket used for the programmatic API.
- The wire format for the API is newline-delimited JSON-RPC 2.0 (I omit the `"jsonrpc": "2.0"` entry to save bytes).
- The API is roughly defined by the structs defined in `brush-shell/src/rpc.rs`.
- Much of the changes to `brush-core/src` are in service of dispatching events needed to support the API.
- This includes `scripts/socket-json-pretty.js` as a lightweight REPL to test the functionality of the new `--rpc-socket` flag.

You can see the REPL in action here:

https://github.com/user-attachments/assets/6ff2ebd5-382e-4733-8dbb-752e9249d8d1

In the video, you can see the notification/response events sent in response to running:

- `ls` (note `processSpawn` and `processExit`
- `echo foo` (note `builtinSpawn` and `builtinExit`)
- `echo foo > /dev/null` (note there are no `stdout` notifications sent for this case)
- `echo foo && ls` (note both `processSpawn` and `builtinSpawn` notifications are sent)


bolinfest:
I admit that the code could benefit from quite a bit of cleanup, so rather than get into the weeds on that, I want to focus on:

- Is Brush interested in supporting any sort of programmatic API like this?
- I believe the main changes to the core are the introduction of `emit_builtin_event()` (which should probably be a method of `Shell` rather than a global...): is this change to core a reasonable "cost" to pay to support a programmatic API?

lu-zero:
Cannot most of those changes live in a separate crate?

reubeno:
@bolinfest Thanks for posting this RFC! It's great to see ideas like these.

I'm definitely interested in a programmatic interface that allows separate processes to drive a `brush`-hosted shell. I also see the benefits in exposing more events/hooks during execution, for multiple reasons. I'd like us to find a way to extend `brush-core` with the right primitive events/hooks that allow higher-level crates to use them in interesting ways.

Let me take some time and review the proof-of-concept work so I can provide more meaningful input/thoughts.

bolinfest:
@reubeno Thanks for taking a look! Yes, I certainly understand the desire to add as little to your core as possible. I certainly wouldn't expect to land some of the wrapper bits in this PR. I mainly added them as a concrete example of something to play with to motivate the fundamental change.

@lu-zero Some of what is here can live in a separate crate, but the key bit that cannot is the instrumentation of events/hooks in core to build the observability tooling on top.

bolinfest:
@reubeno Just checking in: I'm curious if you have had time to think about what sort of hooks could be added to the core to enable the use cases I described without cluttering it. Happy to help execute if there's a plan!

reubeno:
> @reubeno Just checking in: I'm curious if you have had time to think about what sort of hooks could be added to the core to enable the use cases I described without cluttering it. Happy to help execute if there's a plan!

Hi @bolinfest -- many apologies for taking so long to get back to you. (Your interest really was and is appreciated -- the fall/early winter was very busy for me personally.) I've finally managed to get #838 posted, which embodies some of the ideas we've had for a core hook/observer/filter mechanism that could be used for a host of `brush-core` (e.g., `brush-shell` or something like it) to be able to interpose on key operations in the shell while keeping the core interpreter largely unaware.

If you still have time and/or interest, it would be great to get your perspective and input on this approach and whether you could see it providing some of the base support that would enable some of the interesting applications you had in mind.

PR #694, opened 2025-10-03
docs: first draft of an AI policy in CONTRIBUTING.md

reubeno:
After having read up on what some prominent OSS projects are doing, I've taken a stab at authoring some verbiage in `CONTRIBUTING.md` that touches on AI-assisted contributions.

My take is that it's far better to have a policy than not. I also think it's important to be open-minded and encouraging of new ways of development, while still maintaining accountability and appropriate quality bars.

I'm very open to (ongoing) discussion in this space, should any other participants in this project have opinions :smile:.


lu-zero:
Usually LLM tools are fast in generating first drafts and example code, it would be a good idea to require to not use them to address review comments and even better to require to rewrite once what the LLM produced before submission.

reubeno:
> Usually LLM tools are fast in generating first drafts and example code, it would be a good idea to require to not use them to address review comments and even better to require to rewrite once what the LLM produced before submission.

Thanks for your thoughts on this, @lu-zero. I'm in total agreement that contributors must review any generated content -- and that they must be willing to stand behind what they submit, regardless of what tools they used along the way. If there's anything that doesn't meet standards, they should revise it.

I see those end results as the most important. I think we could make recommendations on how best to responsibly use tools along the way to ensure those results, but I'm hesitant to be too prescriptive.

lu-zero:
LLM tend to diverge a bit too easily when addressing comments, adding unrelated changes if not reigned in. 

Either we spend time to write an `AGENTS.md`, it doubles as newcomers guide, so it isn't a bad idea, to get at least those that use agents that consume that information to reduce the odds of mistakes or we are in for much longer review times.

reubeno:
> LLM tend to diverge a bit too easily when addressing comments, adding unrelated changes if not reigned in.
> 
> Either we spend time to write and `AGENTS.md`, it doubles as newcomers guide, so it isn't a bad idea, to get at least those that use agents that consume that information to reduce the odds of mistakes or we are in for much longer review times.

I'm in strong agreement with you on this. I have a draft `AGENTS.md` that I've been experimenting with, but it definitely seems like something we'll need to iterate on as we see what impact it has on the quality of AI output.

Elsie19 (approved):

PR #704, opened 2025-10-11
Forward token location

lu-zero:


reubeno:
Thanks for all your work on this, @lu-zero. I like the approach you've taken with this and am generally on-board. I also appreciate how some of the code gets simpler with the move to `usize`.

The benchmarks are sometimes noisy, but I think they're indicating legitimate increased overhead with these changes. Any thoughts on what we could do to minimize the cost?

lu-zero:
Once we move to winnow we could have less copies around, so we could consider optimizations after we have the feature.

See https://docs.rs/winnow/latest/winnow/trait.Parser.html#method.with_span

In general right now I wrapped in Arc the location information to not pay for the many tiny allocations.

lu-zero:
It is still not fully complete but it is already useful potentially

reubeno:
Are all these changes in now? (Shall we close it?)

lu-zero:
yes :)

PR #711, opened 2025-10-14
fix: workaround error on nightly

reubeno:


reubeno:
FYI @lu-zero -- just a quick workaround for now.

PR #719, opened 2025-10-28
fix: don't allow builtin to invoke disabled builtin

reubeno:
Note that we intentionally disable the new test added for openSUSE tumbleweed because its build of `bash` appears to select a different default behavior.

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -17,6 +17,10 @@ on:
   pull_request:
     branches: ["main"]
 
+  # Run on pushes into `main` as a way to have "nightly"-ish binaries.
+  push:
+    branches: ["main"]
+
 env:
   CARGO_INCREMENTAL: 0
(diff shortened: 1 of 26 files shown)


reubeno:
@copilot This change is failing a PR check when integration tests are run on openSUSE tumbleweed. The reason appears to be that openSUSE intentionally configures its bash build with support for running disabled builtins. For now, we want to proceed with this change but acknowledge that the failing test case won't run correctly on that distro.

Please do the following:

* Extend the brush-compat-tests test harness in this project to allow a YAML test case to annotate the case as not being applicable to certain target distros. (There's already a mechanism that allows tagging a test as only pertaining to, say, sh mode. We should extend it so that the host system's canonical OS moniker via /etc/os-release is something that we can similarly name.)
* Update the failing test to mark that it doesn't apply to openSUSE (but only openSUSE).

PR #730, opened 2025-11-03
fix: address race conditions in basic input tests

reubeno:
This PR refactors terminal mode management in the `basic` input backend to avoid race conditions due to excessive and poorly timed raw mode toggling. More broadly, it revisits terminal mode management to use RAII guards and high-level settings builders instead of manual state management, consolidating terminal control logic into the `brush-core::terminal` module. 

* Introduced `AutoModeGuard` and `Settings` builder for declarative terminal configuration
* Refactored `read` and `mapfile` builtins to use the new terminal API
* Basic input backend: split terminal and non-terminal line reading into separate implementations

### DIFF
--- a/brush-builtins/src/mapfile.rs
+++ b/brush-builtins/src/mapfile.rs
@@ -86,7 +86,7 @@ impl MapFileCommand {
         &self,
         mut input_file: brush_core::openfiles::OpenFile,
     ) -> Result<variables::ArrayLiteral, brush_core::Error> {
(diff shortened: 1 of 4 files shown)


reubeno:
@lu-zero FYI -- I believe this will resolve the flaky tests we've been seeing. They've been symptoms of poor terminal mode management in the "basic" input backend. Most users of `brush` don't use that backend, but tests use it heavily.

PR #740, opened 2025-11-07
feat(bind): extend key binding support

reubeno:
_Extends the `bind` builtin to support binding key sequences to readline functions (previously it only supported shell commands). Bindings to more complex macros remain unimplemented._

- Adds support for parsing and binding key sequences to readline functions (e.g., `"\C-a":beginning-of-line`)
- Implements translation of `InputFunction` variants to reedline events in the interactive layer
- Refactors the parser to distinguish between shell command bindings and readline target bindings


reubeno:
I've done some manual testing. This change makes things better for `fzf` shell integration. Sadly, `atuin` shell integration isn't looking as good as it did previously, but that seems more to be owing to upstream changes in that project's bash integration script.

Two steps forward, one step back.

With that said, this looks to be a general improvement so I'm going to move forward with merging it. There are still errors from readline "macros", though.

PR #760, opened 2025-11-13
chore: add brush crate + docs publishing

reubeno:
Adds `brush` create as wrapper for `brush-shell`, also producing a binary called `brush`.

This will allow back-compat for anyone `cargo install`ing or `cargo binstall`ing `brush-shell` while opening up the ability to do the same with the `brush` crate name. Over time we'll reevaluate.

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -1,4 +1,4 @@
-a#
+#
 # Based on https://github.com/release-plz/release-plz/blob/a5043c478d46d051c00e4fbc85036ac22510f07e/.github/workflows/cd.yml
 #
 


reubeno:
@copilot There are two categories of issues showing up in the failed PR checks on this change. Please help with both:

1. A new zizmor error on the new GitHub workflow. Please apply best practices and resolve the issue.

2. Failures in build now that we have added --workspace to some of the cargo build invocations in the CI workflows. I'd like to keep that flag but once and for all deal with the fact that some of the crates in the workspace simply aren't buildable for some platforms due to upstream constraints in dependencies. Please don't negatively impact other platforms with your changes.

PR #777, opened 2025-11-21
Avoid allocating in `pre_exec` closure

purplesyringa:
`Error::other` allocates memory (see https://github.com/rust-lang/rust/pull/148971). This is bad in multi-threaded programs, which brush AFAIK is. If the fork occurs while the allocator lock is held by another thread, deadlocks can occur, since there's no one left in the new process to unlock the mutex. I do not believe this is UB, and modern libc offer protections against this issue, but this isn't POSIX-compliant and should preferably be avoided.

Luckily, `nix` provides a non-allocating `impl From<Errno> for std::io::Error`, and it passes the correct error code through to the parent process, instead of using the default `-EINVAL` like `Error::other`.

I'm not sure how to best integrate this into your project. I changed the return type of `move_self_to_foreground` to use `std::io::Error`, which works, but feels hacky. If you have a better idea, feel free to apply your own patch and close this PR.

### DIFF
--- a/brush-core/src/sys/stubs/terminal.rs
+++ b/brush-core/src/sys/stubs/terminal.rs
@@ -67,6 +67,6 @@ pub fn move_to_foreground(_pid: sys::process::ProcessId) -> Result<(), error::Er
 /// Moves the current process to the foreground of the attached terminal.
 ///
 /// This is a stub implementation that returns `None`.
-pub fn move_self_to_foreground() -> Result<(), error::Error> {
+pub fn move_self_to_foreground() -> Result<(), std::io::Error> {
     Ok(())
 }
--- a/brush-core/src/sys/unix/commands.rs
+++ b/brush-core/src/sys/unix/commands.rs
@@ -61,6 +61,6 @@ impl CommandFgControlExt for std::process::Command {
 fn setup_process_before_exec() -> Result<(), std::io::Error> {
     use crate::sys;
 
-    sys::terminal::move_self_to_foreground().map_err(std::io::Error::other)?;
+    sys::terminal::move_self_to_foreground()?;
     Ok(())
 }
--- a/brush-core/src/sys/unix/terminal.rs
+++ b/brush-core/src/sys/unix/terminal.rs
@@ -97,7 +97,8 @@ pub fn move_to_foreground(pid: sys::process::ProcessId) -> Result<(), error::Err
 }
 
 /// Moves the current process to the foreground of the attached terminal.
-pub fn move_self_to_foreground() -> Result<(), error::Error> {
+// This function needs to return `std::io::Error` so that the OS error code can be recovered.
+pub fn move_self_to_foreground() -> Result<(), std::io::Error> {
     if std::io::stdin().is_terminal() {
         let pgid = nix::unistd::getpgid(None)?;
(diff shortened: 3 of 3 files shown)


reubeno:
Thanks for identifying this issue and contributing a fix, @purplesyringa! The latest version of your change seems reasonable given the constraints.

(I read your linked issue and see how many other projects you believe are affected. If you're willing to share, I'm curious to know how you came across this and what ideas you have on how we could mechanically catch these sorts of challenges in the future. (I know that pre_exec is unsafe so I recognize that some risk was already implied.))

purplesyringa:
First of all, I'm confused why the bot mentions `nix::Error` -- I noticed that that wouldn't work on Windows and force-pushed a fix that replaced the type with `std::io::Error`, so that must be outdated. But yeah, this would require a public API change I'm afraid, unless you're willing to `match` on your own `Error`.

> If you're willing to share, I'm curious to know how you came across this

My girlfriend asked me to review her code. I thought `Error::other` is an odd choice for `pre_exec`, but for a different reason, though I don't remember which reason specifically. Eventually I realized that this reason doesn't apply, but that `Error::other` allocates, and that's also a problem. And if I missed this bug on my first review, others might have missed it too. A quick GitHub search later showed that it's quite popular.

> and what ideas you have on how we could mechanically catch these sorts of challenges in the future.

That's a good question. A clippy lint might help, but it's not clear how to decide which functions are allowed to be called, since we don't have effects and there's no annotations of that sort, so it'd just be a blacklist. Something that I think might be worthwhile is to disable the global allocator in `std::process::Command` after `fork`, at least in debug mode/with debug assertions on. I'm not sure if Rust project would be willing to have that in `std`, since a) some libc (try to) guarantee allocation is safe and b) this is actually safe in single-threaded programs, but I can ask. Barring that, we could have a nightly-only crate wrapping any passed allocator, to be used in `#[global_allocator]`.

reubeno:
>First of all, I'm confused why the bot mentions nix::Error -- I noticed that that wouldn't work on Windows and force-pushed a fix that replaced the type with std::io::Error, so that must be outdated.

That's right. Once the checks re-ran, it updated the compat report accordingly. We really only use the `sys` module within this broader project, but did need it public. (We'll be looking into ways to place it behind a feature flag, or otherwise mark it as less stable.) For now, I think it's an acceptable change.

>And if I missed this bug on my first review, others might have missed it too.

Absolutely--I'm one of those. I'm aware of the risks of allocation between `fork` and `exec` in a multithreaded program but didn't stop to think about the safety of constructing an error there. It's a good catch!

>Something that I think might be worthwhile is to disable the global allocator in std::process::Command after fork, at least in debug mode/with debug assertions on. I'm not sure if Rust project would be willing to have that in std, since a) some libc (try to) guarantee allocation is safe and b) this is actually safe in single-threaded programs, but I can ask.

It sounds like it's worth a conversation, but agree that such checks may be overly restrictive for some platforms. I like the idea of a wrapper for debugging/validation. I haven't used any of them before, but I see on `crates.io` a handful of crates that provide debugging-oriented global allocators for verification/tracking/instrumentation. I could envision using that during testing to collect allocation metrics as well as look for anything that's... suspicious.

PR #832, opened 2025-12-05
refactor: shell / interactive-shell layering

reubeno:
Refactors layering of `InteractiveShell` implementations with respect to `Shell`. The UI-specific implementations are now restructured as implementations of `InputBackend` that receive (but do not *construct*) `Shell` instances.

The reused `InteractiveShell` logic has been restructured in an extension trait for use with a `Shell` and does not get instantiated.


reubeno:
@copilot Please review all of the feedback you provided across your reviews of this PR and propose a set of changes (via separate PR into this branch). Make sure you properly understand that tokio's mutex is async-aware, also make sure that the code is all still clippy-clean.

reubeno:
@copilot I've thought about it more. I'd like to find a way to:

* Restore the public export of Shell::new() with its current behavior but mark it deprecated, instead pointing at the builder pattern.
* Update the Shell final build function not to be async -- that will require producing a Shell that *doesn't* load and run RC/profile scripts. Provide an alternate function that will do this running for benefit of consumers that do still need to do this.
* Add builder options to control whether or not the Shell initializes well-known variables.
* Add builder options that enable constructing the Shell with a set of custom variables.
* Add builder option(s) that enable customizing initial open files/fds.

Please think through the above and make sure you have a coherent evolution of initialization design.

PR #838, opened 2025-12-09
feat: introduce `ShellExtensions` trait + filterable op mechanism

reubeno:
Introduces a mechanism for a shell host to register *extensions* with a `Shell` instance. These extensions may include an error formatter. Also, __*experimentally*__, the host may register one or more _filters_ that may observe and/or influence certain key operations (pre- and post-).

* Adds new extensions::ShellExtensions trait for extending shell behavior with pre/post operation hooks.
* Implements filtering infrastructure with `FilterableOp` trait and `with_filter!` macro.
* Enlightens word expansion, simple command execution, and external command spawning to be filterable.
* Migrates error formatting from `ErrorFormatter` to `ShellExtensions::format_error`.
* Adds an example demonstrating how a shell host can observe simple commands being executed.

Future work in this area may include:

* Making additional operations in the shell filterable; likely candidates include: variable adds/updates, function invocation, etc.
* Enabling `ShellExtension` functions related to filterable operations to access more context around the operation (e.g., current call stack, general execution state)
* Adding experimental features to `brush-shell` itself that make use of `ShellExtensions`

Down the road, the aim would be to make `Shell` generic over a type that implements `ShellExtensions`, i.e. `Shell<SE: ShellExtensions>`. This could allow us to avoid `dyn`-invoking the extensions and being able to enable general use of extensions while still keeping low overhead for the no-op case.


reubeno:
@copilot  Unsurprising these changes regress performance for the shell, as evidenced by the benchmark numbers that were posted to this PR thread. Since the benchmarks were written without any filters installed, this indicates that the no-op case received measurable overhead.

Please analyze the changes and benchmark results and propose some changes that decrease the no-op overhead.

k2tomasz:
Hi @reubeno! 👋

This PR caught my attention - the ShellExtensions trait looks very useful. I'm working on a project that wraps brush-core and needs to intercept commands for analysis/filtering purposes.

After reviewing the code, I can see that:

✅ Blocking works via PreFilterResult::Return(Err(...)) - exactly what I need
✅ pre_exec_external_command fires before spawn - perfect timing
✅ SimpleCommand gives me the expanded executable and args

A few things that would make this even more useful for my use case:

Pipeline context - For cmd1 | cmd2, knowing the pipeline position (0, 1) and total length (2) would help with analysis
Redirections - Access to stdin/stdout/stderr targets in the hook data
Expanded variables map - Seeing which variables were expanded (e.g., {"cmd": "curl"} when $cmd resolved)

Happy to test this and provide feedback. Also, if you need help with anything on brush - documentation, testing, features - I'd be glad to contribute. Great project! 🦀

k2tomasz:
Would it be possible to add one more hook? Something like this would solve my problem:

```
fn pre_exec_with_context<'a>(
    &self,
    input: CommandContext<'a>,
) -> PreFilterResult<CommandContextOp<'a>>;

pub struct CommandContext<'a> {
    /// The SimpleCommand being executed
    pub command: SimpleCommand<'a>,
    /// Original command string before any expansion
    pub raw_command: &'a str,
    /// Map of variables that were expanded: {"VAR" -> "value"}
    pub expanded_variables: HashMap<String, String>,
    /// Pipeline info if part of a pipeline
    pub pipeline: Option<PipelineContext>,
    /// Redirections
    pub redirections: Vec<Redirection>,
}

pub struct PipelineContext {
    pub position: usize,  // 0-indexed
    pub total: usize,
}
```

reubeno:
Thanks for your interest, @k2tomasz!

We'll typically have the unexpanded Abstract Syntax Tree available but holding onto all of the intermediate expansions for every simple command could get complicated. But what if the extension could hook: redirection setup, expansion, pipeline execution, and simple command execution? Would that work? That would push more work to the extension/hook implementation to track the context along the way but bring the plus side that you'd only need to track it for what's relevant to you.

As for tracking variable  expansions -- could you share more about what you're trying to accomplish? Full expansion is a lot more than variable replacement; it could include command substitution, brace expansion, string manipulation, nameref lookups, etc. How would you want that to work for your use case?

Perhaps you have some motivating examples of shell script and what you would want to be able to have?

k2tomasz:
@reubeno take a look at this:

https://github.com/reubeno/brush/pull/935

I have added 2 new hooks to your filters:

1. post_expansion_pre_dispatch_simple_command - A post-expansion policy hook that:

- Fires after word expansion and after the pre_exec_simple_command filter chain
- Fires before dispatch to builtin/function/external execution
- Receives &SimpleCommand and &CommandContext (with pipeline info, redirections, dispatch target, raw command)
- Returns a PolicyDecision (Continue or Return early)
- Is read-only — cannot modify the command, only allow/deny execution

2. pre_command_substitution - A pre-command-substitution policy hook that:

- Fires before executing $(...) or backtick substitutions during expansion
- Receives CommandSubstitutionContext (raw body text, syntax type)
- Returns CommandSubstitutionDecision (Allow, ReplaceOutput, or Error)
- Allows blocking or replacing command substitution output to prevent side effects during expansion

Let me know what you think. Maybe there is a better way to do this.

k2tomasz:
@reubeno just want to check and ask when do you think this could be merged into main?

reubeno:
> @reubeno just want to check and ask when do you think this could be merged into main?

Good question. I've been working on it offline some--integrating it with the new `ShellExtensions` in `main`--but likely won't be able to move it forward until this coming weekend.

My plan was to get just one hook point merged so that others interested (perhaps you?) could help work on the other hook points needed in parallel PRs.

k2tomasz:
@reubeno yes I could add more hook points once you merge your work into main.

k2tomasz:
@reubeno any update on this PR?

reubeno:
> @reubeno any update on this PR?

@k2tomasz  -- I've posted draft PR #972. I need to go back through it and it needs proper review, though. Feedback on the shape of it / traits / etc. is certainly welcome. The main goal is to have the extension mechanism enabled without requiring feature flags, but with as close to 0 as possible runtime overhead when not engaged.

PR #865, opened 2025-12-23
Fix true and false with any args

### LINKED ISSUE #861: true --help should be silent
`bash` `true` does not have `--help`.

oech3:
Closes #861

### DIFF
--- a/brush-shell/tests/cases/builtins/true_false.yaml
+++ b/brush-shell/tests/cases/builtins/true_false.yaml
@@ -1,0 +1,7 @@
+name: "true and false"
+cases:
+  - name: "true and false ignore all args"
+    stdin: |
+      builtin false --help; echo "builtin false --help => $?"
+      builtin true --help; echo "builtin true --help => $?"
+      builtin true s87tien5; echo "builtin true s87tien5 => $?"
--- a/brush-builtins/src/factory.rs
+++ b/brush-builtins/src/factory.rs
@@ -91,7 +91,7 @@ pub fn default_builtins(set: BuiltinSet) -> HashMap<String, builtins::Registrati
     #[cfg(feature = "builtin.command")]
     m.insert("command".into(), builtin::<command::CommandCommand>());
     #[cfg(feature = "builtin.false")]
-    m.insert("false".into(), builtin::<false_::FalseCommand>());
+    m.insert("false".into(), simple_builtin::<false_::FalseCommand>());
     #[cfg(feature = "builtin.fg")]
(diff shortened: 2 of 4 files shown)


brush-shell/tests/cases/builtins/builtin.yaml
```diff
@@ -15,8 +15,8 @@ cases:
 
   - name: "builtin passing through results"
     stdin: |
-      builtin false; echo "builtin false => $?"
-      builtin true; echo "builtin true => $?"
+      builtin false --help; echo "builtin false => $?"
```
reubeno: Thanks for adding to the tests.

This test case happens to call `false`/`true`, but it's really focused on testing `builtin`. Will you please:

* Revert the change to this file to keep this test simple.
* Add a new file in this directory called something like `true_false.yaml` that defines a new test set
* In that *new* file add test cases that cover the usage you're fixing with this PR (e.g., basic `false`/`true` usage, usage with non-dash arguments, usage with `--help` or other dash-arguments)

reubeno (commented):
Thanks for the contribution, @oech3!

I've left a comment on the test change and will unblock the PR checks to run.

brush-shell/tests/cases/builtins/builtin.yaml
oech3: Should I use `false --help`, or `! false --help`?

reubeno: `false --help` is fine.

brush-shell/tests/cases/builtins/true_false.yaml
```diff
@@ -0,0 +1,7 @@
+name: "true and false"
+cases:
+  - name: "true and false ignore all args"
+    stdin: |
+      builtin false --help; echo "builtin false --help => $?"
+      builtin true --help; echo "builtin true --help => $?"
+      builtin true s87tien5; echo "builtin true s87tien5 => $?"
```
reubeno, line 7: Does this string (`"s87..."`) mean something?

oech3: No. Just a random string.

reubeno:
Looks like the tests passed but format checks failed; you will need to run `cargo fmt` and update.

oech3:
I mostly copy pasted from `:`. Should fmt it too?

reubeno:
If you run `cargo fmt` from the root of the project repo, it should reformat anything that's needed.

oech3:
Done.

PR #871, opened 2025-12-25
Initial winnow implementation

lu-zero:
It has two implementations, one still using the tokenizer the other directly parsing a str input. 

Most of the code had been machine-ported from the peg implementation and then refined so it is a bit faster than the original, there is still potentially more to refactor but we can start reasoning about it.


reubeno:
Thanks for cleaning up the changes. I still need to actually read through the winnow-based parsers, but what do you think about arranging a runtime switch to select the parser we want to use? If you're open to it, I'm happy to help with some of the glue code for this too.

I'm thinking we could extract a `Parser` trait (or similar name) that must implement a `fn` that takes text as input and yields an AST. We could implement this for the PEG-based impl as well as the winnow-based impl(s). An option on `Shell` could select the desired impl, with default going to the (legacy) PEG one.

We could then arrange to run all of our YAML-based integration tests against *all* parsers without needing to recompile `brush`, and present results. This would give us a way to track functional parity and help with side-by-side comparison. (And ditto for the benchmarks you've set up.) I'd want to update the PR checks to run tests against multiple parsers, but only block if we fail tests on the PEG one. This would give us a clear way to track progress and evaluate when we'd be ready to switch over the default.

lu-zero:
> Thanks for cleaning up the changes. I still need to actually read through the winnow-based parsers, but what do you think about arranging a runtime switch to select the parser we want to use? If you're open to it, I'm happy to help with some of the glue code for this too.

Sounds feasible and shouldn't be too hard to achieve given what we currently have. Just we could clean up a little the options.

> I'm thinking we could extract a `Parser` trait (or similar name) that must implement a `fn` that takes text as input and yields an AST. We could implement this for the PEG-based impl as well as the winnow-based impl(s). An option on `Shell` could select the desired impl, with default going to the (legacy) PEG one.

Yes, I have the two winnow impls because I gradually did the translation and benchmarked and cleaned up till I was confident winnow can be faster and overall nicer than the status quo. If we can ditch the tokenization from the public API we should spare quite a bit of allocations and we can focus on cleaning up even more the winnow_str impl.

> We could then arrange to run all of our YAML-based integration tests against _all_ parsers without needing to recompile `brush`, and present results. This would give us a way to track functional parity and help with side-by-side comparison. (And ditto for the benchmarks you've set up.) I'd want to update the PR checks to run tests against multiple parsers, but only block if we fail tests on the PEG one. This would give us a clear way to track progress and evaluate when we'd be ready to switch over the default.

Yes, we might want to do a bit of refactor so that we can make the location a no-op for the tests since that part will have discrepancies.

lu-zero:
@reubeno this weekend we could try to do a first review pass to make sure all the not too beautiful bits are needed for performance, but it seems that we are getting closer to an acceptable version.

PR #873, opened 2025-12-25
fix: gracefully handle closed stdout

### LINKED ISSUE #869: builtin echo 1 >&- crushes

reubeno:
Reimplements `stdout()` et al. not to `.unwrap()` but instead return a valid object that implements `std::io::Write` but fails every I/O. 

Along the way added a way to create an `OpenFile` that doesn't map to a real host file descriptor.

Resolves #869


oech3:
`$ typo 2>&-` crushed with this commit.

PR #895, opened 2026-01-01
feat: optional TOML-based config file for brush-shell

reubeno:
_TL;DR: This adds a TOML-based config file under standard XDG config dir to configure user defaults for `brush`'s user experience._ 

## Background

`brush` uses `bash`-style profile/rc files for initialization scripts; that support is foundationally present in `brush-core`. However, we've recently found cases where it's useful to influence how `brush-shell` (i.e., the `brush` *binary*) behaves by default -- and in ways that need to be configured before it instantiates and starts running the interpreter.

The best current example is `--enable-zsh-hooks`. As discussed in #445, this flag is experimental and not enabled-by-default. It's also convenient for users who wish to opt into this behavior to be able to configure it as enabled for all `brush` instances run. (In particular, this is a requirement for the case where `brush` is the default shell, registered in `/etc/passwd`.) This isn't the only example, and we anticipate there will be more that are specific to the `brush` shell itself.

To address this, we add initial support for specifying user defaults for some `brush` user experience settings under standard XDG dirs. In practice, this will mean something like `~/.config/brush/config.toml`.

### DIFF
--- a/brush-shell/src/args.rs
+++ b/brush-shell/src/args.rs
@@ -64,11 +64,11 @@ pub struct CommandLineArgs {
     #[clap(long = "version", action = clap::ArgAction::Version)]
     pub version: Option<bool>,
 
-    /// Path to configuration file (overrides default location).
+    /// Path to TOML-based `brush` config file (overrides default location).
     #[clap(long = "config", value_name = "FILE", help_heading = HEADING_CONFIG_OPTIONS)]
     pub config_file: Option<PathBuf>,
 
-    /// Disable loading of configuration file.
+    /// Disable loading of TOML-based `brush` config file.
     #[clap(long = "no-config", help_heading = HEADING_CONFIG_OPTIONS)]
     pub no_config: bool,
 
--- a/brush-shell/tests/integration_tests.rs
+++ b/brush-shell/tests/integration_tests.rs
@@ -10,8 +10,8 @@ use predicates::prelude::PredicateBooleanExt;
 #[test]
 fn get_version_variables() -> anyhow::Result<()> {
     let shell_path = assert_cmd::cargo::cargo_bin!("brush");
-    let brush_ver_str = get_variable(shell_path, "BRUSH_VERSION")?;
-    let bash_ver_str = get_variable(shell_path, "BASH_VERSION")?;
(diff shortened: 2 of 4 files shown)


tessus:
Just a quick note. All available crates that return `XDG_` paths on different OS make one mistake: they return `~/Library/Application Support` on macOS.

**This directory is supposed to be used for GUI applications only and should not be used for terminal/cli applications.** (I believe this is even specified in the official Apple documentation.) Unfortunately the devs who created those crates ignored that part.
On macOS these crates should read the `XDG_CONFIG_DIR` env var and fallback to `~/.config`

I haven't checked how `etcetera` handles the config dir on macOS, so maybe this comment is obdolete.

reubeno:
> Just a quick note. All available crates that return `XDG_` paths on different OS make one mistake: they return `~/Library/Application Support` on macOS.
> 
> **This directory is supposed to be used for GUI applications only and should not be used for terminal/cli applications.** (I believe this is even specified in the official Apple documentation.) Unfortunately the devs who created those crates ignored that part. On macOS these crates should read the `XDG_CONFIG_DIR` env var and fallback to `~/.config`
> 
> I haven't checked how `etcetera` handles the config dir on macOS, so maybe this comment is obdolete.

Thanks for highlighting this, @tessus. Some other contributors (e.g., @lu-zero) previously raised the same concerns, and that helped influence our selection of `etcetera`. It's more flexible than some of the other crates and offers multiple "strategies" for platform-based selection. The strategy we're using is the _first_ one ("base") mentioned in this excerpt from its docs:

> ### [Native Strategy](https://docs.rs/etcetera/latest/etcetera/#native-strategy)
> `choose_base_strategy()` and `choose_app_strategy()` will use the `XDG` strategy on Linux & macOS, and the `Windows` strategy on Windows. > This is used by most CLI tools & some GUI tools on each platform.
>
> If you’re developing a GUI application, you might want to use the “Standard directories” on macOS by using `choose_native_strategy()` instead. Note that if your application expects the user to modify the configuration files, you should still prefer the `XDG` strategy on macOS.

We've got some more testing to do, but I believe this will align with your feedback.

tessus:
Awesome, thanks for the info.

PR #899, opened 2026-01-03
Refactor the PEG parser

lu-zero:
Prerequisite to add the winnow parsers.

### DIFF
--- a/brush-parser/src/parser/snapshots/brush_parser__parser__tests__parse_arith_and_non_arith_parens.snap
+++ b/brush-parser/src/parser/snapshots/brush_parser__parser__tests__parse_arith_and_non_arith_parens.snap
@@ -1,0 +1,126 @@
+---
+source: brush-parser/src/parser/mod.rs
+expression: "ParseResult { input, result: &result }"
+---
+ParseResult(
+  input: "( : && ( (( 0 )) || : ) )",
+  result: Program(
+    complete_commands: [
+      CompoundList([
+        CompoundListItem(AndOrList(
+          first: Pipeline(
+            seq: [
+              Compound(Subshell(SubshellCommand(
+                list: CompoundList([
+                  CompoundListItem(AndOrList(
+                    first: Pipeline(
+                      seq: [
+                        Simple(SimpleCommand(
+                          word_or_name: Some(Word(
+                            value: ":",
+                            loc: Some(SourceSpan(
+                              start: SourcePosition(
+                                index: 2,
+                                line: 1,
+                                column: 3,
+                              ),
+                              end: SourcePosition(
+                                index: 3,
+                                line: 1,
+                                column: 4,
+                              ),
+                            )),
+                          )),
+                        )),
+                      ],
+                    ),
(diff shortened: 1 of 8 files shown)


brush-parser/src/parser/snapshots/brush_parser__parser__peg__tests__parse_arith_and_non_arith_parens.snap
```diff
@@ -0,0 +1,126 @@
+---
```
reubeno, line 1: Are the old versions of these snapshots ones we need to delete too?

brush-parser/src/parser/snapshots/brush_parser__parser__peg__tests__parse_arith_and_non_arith_parens.snap
```diff
@@ -0,0 +1,126 @@
...
+source: brush-parser/src/parser/peg.rs
```
reubeno: Any thoughts on how to make these more generic so we can use them as the snapshots for both parsers? (I'm okay with that being deferred to a separate PR.)

reubeno (commented):
Changes generally look good to me, just 2 quick comments -- look like a few checks failed you'll need to resolve.

brush-parser/src/parser/snapshots/brush_parser__parser__peg__tests__parse_arith_and_non_arith_parens.snap
lu-zero: We can have common tests once we test with no location printed in the snapshot.

brush-parser/src/parser/snapshots/brush_parser__parser__peg__tests__parse_arith_and_non_arith_parens.snap
lu-zero: potentially, I didn't continue today but I wanted to give you a starting point. We should consider if we can hide away the tokenizer to simplify a bit everything.

brush-parser/src/parser/snapshots/brush_parser__parser__peg__tests__parse_arith_and_non_arith_parens.snap
lu-zero: [redactions](https://insta.rs/docs/redactions/) could potentially make it simple.

reubeno: That sounds promising. Let's try to do that in a future PR.

reubeno:
@lu-zero I've gotten the test failures resolved; I'm going to merge this PR as incremental progress. We'll still need follow-on work to loosen the contracts so that `brush-core` doesn't know about existing tokenization APIs and to set up some _stub_ for the new parser that can be called at runtime but doesn't pass tests. (And for us to arrange to run full set of tests against both.) Let's plan that for the next PR.

PR #901, opened 2026-01-04
refactor: introduce `ShellRuntime` and `ShellBehavior` traits

reubeno:


reubeno:
This helped us learn a lot, but also required exposing more shell internals from brush-core than was necessary. This has been supplanted by #941.

Issue #918, opened 2026-01-10
Different exit code at hash /dev/null


oech3:
```
brush$ hash /dev/null
/dev/null: not found # exit 1
bash$ hash /dev/null # exit 0
bash$ hash /dev/nulll # exit 0
```

reubeno:
Thanks for reporting this! This might be a good first issue for someone to look at.

reubeno:
Fixed by #924 with a test case to confirm resolution.

Thanks, @zvyaan!

PR #924, opened 2026-01-11
hash: accept names with slashes without lookup or error

zvyaan:
This PR solves #918

This is my first Rust OSS contribution so I apologise in advance if I didn't go about things the right way, but I have tried my best to follow the contributing guide and follow best practices.

Brush currently attempts to resolve names with slashes in the `hash` builtin, returning an error and exit status 1 if the path does not exist, but this behaviour doesn't align with Bash's.

POSIX specifies that the behaviour for utility names containing slashes is 'unspecified', but exit status should only be non-zero for errors. Bash accepts names with slashes, but performs no lookup nor any output, and exits with status 0.

This change aligns Brush's `hash` builtin with Bash's behaviour: names containing slashes are now accepted without any PATH lookup / error.

As a result, we now have the following:
```
~/brush$ hash /dev/null # exit 0
~/brush$ hash /dev/nulll # exit 0
~/brush$ hash -t /dev/null
/dev/null: not found
```

### DIFF
--- a/brush-shell/tests/cases/compat/builtins/hash.yaml
+++ b/brush-shell/tests/cases/compat/builtins/hash.yaml
@@ -61,3 +61,9 @@ cases:
       hash -r && echo "3. Result: $?"
       hash -t somecmd1 && echo "4. Result: $?"
       hash -t somecmd2 && echo "5. Result: $?"
+
+  - name: "Hash file path"
+    ignore_stderr: true
+    stdin: |
+      hash /dev/null && echo "1. Result: $?"
+      hash /dev/nulll && echo "2. Result: $?"


reubeno:
Welcome, @zvyaan! Thanks for picking up this issue and for the detailed summary. The change seems reasonable.

Our project's way to confirm that we've resolved a compatibility issue (and that it remains resolved) is to add YAML-based test cases under the `brush-shell/tests/cases/compat` directory. The test cases are defined with script snippets that are run against `brush` and `bash`, with their results (stdout, stderr, exit code) auto-compared. It looks like we already have a set of [test cases for the `hash` builtin](https://github.com/reubeno/brush/blob/main/brush-shell/tests/cases/compat/builtins/hash.yaml) but it must not cover this scenario.

Will you please extend `hash.yaml` with one or more cases to cover the behavior that you've fixed? What I typically like to do is the following:

* Checkout `main`
* Add the YAML test case
* Run the tests via `cargo nextest` or `cargo test`; confirm that they *fail* (without the fix).
* Make the fix or checkout the branch with the fix.
* Re-run tests to confirm they pass.

_( Tip: for a quick way to run just a single test, you can run something like: `cargo nextest run -- 'TEST-CASE-NAME'` )_

reubeno:
Oh, and I noticed a failing "OS target test"; that doesn't look related to your changes. Some recent changes introduced some flaky tests related to the `read` builtin that we're working to resolve. We'll work to sort those out, but don't worry about it if you see failures in those tests (`read`).

zvyaan:
Hi @reubeno, thanks for the detailed response!

I've just added a new compatibility test to `hash.yaml` and verified that the test fails on main and passes with the fix applied. The single test case alone fully covers the issue.

PR #936, opened 2026-01-16
perf: random assortment of performance changes

Elsie19:
These are some very easy performance changes I found. I'll probably do some later as well.


reubeno:
Thanks for the improvement, @Elsie19!

The change looks great -- the PR checks caught a clippy warning in the benchmarks; would you mind resolving that when you get a chance?

PR #939, opened 2026-01-17
Add test for shadowing reserved words e.g. for (){ :; }

oech3:


oech3:
Those words broke `yaml` syntax

reubeno:
I don't see a yaml syntax error but now I realize that since these tests are supposed to fail it's hard to know if later commands will fail the same way.

We may need to split these up into separate test cases.

If that's more than you have time for, I'd be okay with just a representative set of cases.

oech3:
later...(or please edit this or reopen another PR

oech3:
+115 line is quite large...

oech3:
I'm not faliliar with this repo's test suite. But can we really merge this (just a bunch of CI failure)?

reubeno:
Thanks for taking a pass at this. 

> I'm not faliliar with this repo's test suite. But can we really merge this (just a bunch of CI failure)?

No; we can add new tests marked with `known_failure: true`. That allows us to add tests but acknowledge that they're not passing yet.

I'm going to end up closing this one since I was able to look at it and get it properly fixed. While I didn't add as thorough a set of tests as your approach, I think we'll catch what we need for now.

Thanks again for finding the issue and getting the ball rolling making progress against it :smile:

PR #946, opened 2026-01-19
fix(*): more direct Intos

Elsie19:
A lot of the time, you put `.as_ref().to_string()` where you could've just made the bound `T: Into<String>`.

### DIFF
--- a/brush-core/src/shell.rs
+++ b/brush-core/src/shell.rs
@@ -1694,11 +1694,13 @@ impl<SE: extensions::ShellExtensions> Shell<SE> {
     /// # Arguments
     ///
     /// * `candidate_name` - The name of the file to look for.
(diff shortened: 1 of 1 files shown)


brush-core/src/shell.rs
```diff
@@ -1690,15 +1694,16 @@ impl<SE: extensions::ShellExtensions> Shell<SE> {
     /// # Arguments
     ///
     /// * `candidate_name` - The name of the file to look for.
-    pub fn find_first_executable_in_path_using_cache<S: AsRef<str>>(
+    pub fn find_first_executable_in_path_using_cache<T: Into<String>>(
```
reubeno: We don't actually always need an owned copy of the string here; if it's a cache hit or doesn't exist in the filesystem, we stick to just using the borrowed reference.

Let's leave this one as is.

PR #958, opened 2026-01-25
Publish from main

oech3:
Instead of just discarding binaries with release profile

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ env:
 
 permissions:
   actions: read
-  contents: read
+  contents: write # publish from main
 
 # Only allow one run of the workflow per branch / PR at a time.
 concurrency:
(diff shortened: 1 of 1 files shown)


oech3:
>What's the problem you're trying to solve?

Because I (and probably you) are not interested in outdated bugs in stable tags.

>they are published as workflow artifacts and downloadable.

Archived artifacts have non frozen URLs. So it is difficult to DL it.

>(1) can become iterating as spam to followers' feeds 

I made this as a pre-release generator instead of release. (Does (1) apply for pre?)

>(2) pollutes the release feed.

I made this to reuse same `tag/main` instead of producing too many tags. Same URL s are reused.

reubeno:
Would an install script meet your needs? E.g.


```
install-brush.sh --channel ci

```

oech3:
Probably yes. But I don't know how to do that.

reubeno:
It's something I've been working on a bit. It may require some prerequisite tools like jq and possibly GitHub access for `main` builds. (Installing official releases will be far simpler and not require them.)

I'm also open to having a prerelease channel for `main` builds we choose to promote for wider early testing.

Ultimately it would be nice for `brush` also to be able to update itself -- that will make it easier to stay updated too. Something like `brushctl update` with an optional channel argument. There are some rust crates that help with this.

oech3:
myself does not use `cargo binstall`. But does it support archives?
Self update is not compatible with package managers.

reubeno:
Binary releases of any kind aren't compatible with traditional package managers.

I think it would be entirely reasonable for the self update mechanism to be feature gated and for distro packaged versions of brush not to enable that feature.

But now I feel like this conversation is getting confused. Are you trying to use brush via package manager or are you trying to get binaries published for `main`? The two aren't compatible for most distros.

oech3:
I think brew, winget, choco and AUR can use binary

PR #967, opened 2026-01-29
fix: handle more readline macro forms

reubeno:


brush-interactive/src/reedline/edit_mode.rs
```diff
@@ -239,38 +242,90 @@ impl UpdatableBindings {
...
+    fn find_longest_prefix_match(&self, bytes: &[u8]) -> Option<(Vec<u8>, &KeyAction)> {
+        // radix_trie's get_ancestor returns the longest prefix that has a value.
+        // We need to search for progressively longer prefixes.
+        let mut best_match: Option<(Vec<u8>, &KeyAction)> = None;
+
+        // Try each prefix length from 1 to bytes.len().
+        for len in 1..=bytes.len() {
+            let prefix: Vec<u8> = bytes[..len].to_vec();
+            if let Some(action) = self.raw_mappings.get(&prefix) {
+                best_match = Some((prefix, action));
```
Elsie19: I haven't checked for compile errors but I think:

```suggestion
            let prefix = &bytes[..len];
            if let Some(action) = self.raw_mappings.get(prefix) {
                best_match = Some((prefix.to_vec(), action));
```

Would be more efficient because you're only making a heap allocation for `prefix` once?

PR #968, opened 2026-01-31
fix: use more `PathBuf`

Its-Just-Nans:
Use `std::env::split_paths`

### DIFF
--- a/brush-core/src/sys/stubs/fs.rs
+++ b/brush-core/src/sys/stubs/fs.rs
@@ -60,13 +60,13 @@ pub(crate) trait MetadataExt {
 
 impl MetadataExt for std::fs::Metadata {}
 
-pub(crate) fn get_default_executable_search_paths() -> Vec<PathBuf> {
+pub(crate) fn get_default_executable_search_paths() -> Vec<std::path::PathBuf> {
     vec![]
 }
 
 /// Returns the default paths where standard Unix utilities are typically installed.
 /// This is a stub implementation that returns an empty vector.
-pub fn get_default_standard_utils_paths() -> Vec<PathBuf> {
+pub fn get_default_standard_utils_paths() -> Vec<std::path::PathBuf> {
     vec![]
 }
 


reubeno:
Thanks for the PR, @Its-Just-Nans ! We appreciate the contribution.

This certainly simplifies code. The one consideration is that it will mean that the (experimental / proof-of-concept) Windows builds will handle platform-native semicolon-separated PATH vars yet be incompatible with other platforms' bash scripts that assume `:` is always the separator. Let me think about that for a little.

In the meantime I've unblocked PR checks and kicked off a copilot code review. (Some of the comments seem bogus so I'll triage them when I get a chance.)

brush-test-harness/src/config.rs
```diff
@@ -25,7 +25,7 @@ pub struct ShellConfig {
 
 impl ShellConfig {
     /// Computes the PATH variable to use for tests.
-    pub fn compute_test_path_var(&self) -> String {
+    pub fn compute_test_path_var(&self) -> OsString {
         let mut dirs = vec![];
```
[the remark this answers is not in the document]
Its-Just-Nans: done

brush-core/src/sys/unix/fs.rs
```diff
@@ -81,40 +69,47 @@ fn try_get_file_mode(path: &Path) -> Option<u32> {
     path.metadata().map(|metadata| metadata.mode()).ok()
 }
 
-pub(crate) fn get_default_executable_search_paths() -> Vec<String> {
-    DEFAULT_EXECUTABLE_SEARCH_PATHS
-        .iter()
-        .map(|s| (*s).to_owned())
-        .collect()
+pub(crate) fn get_default_executable_search_paths() -> Vec<PathBuf> {
+    let default_executable_search_paths = [
```
reubeno: Any reason not to just directly return `vec!["/usr/local/sbin".into(), ...]`?

brush-core/src/sys/unix/fs.rs
```diff
@@ -81,40 +69,47 @@ fn try_get_file_mode(path: &Path) -> Option<u32> {
...
+        if !cs_path.as_os_str().is_empty() {
+            return std::env::split_paths(&cs_path).collect();
         }
     }
-
-    DEFAULT_STANDARD_UTILS_PATHS
-        .iter()
-        .map(|s| (*s).to_owned())
-        .collect()
+    let default_standard_utils_paths = [
```
reubeno: _Same question as above._

brush-core/src/sys/unix/fs.rs
Its-Just-Nans: It was useful for explaining the content of the variable but it's usless
I've added a comment

Maybe I can even remove the comment

brush-core/src/sys/unix/fs.rs
Its-Just-Nans: It was useful for explaining the content of the variable but it's usless
I've added a comment

Maybe I can even remove the comment

reubeno:
LGTM. Will watch check re-runs.

reubeno:
_I've thought about the Windows platform consideration above; let's proceed with the change. We can revisit later for Windows if there's enough interest and if we decide we need different behavior there._

PR #972, opened 2026-02-01
feat: add statically registered filters

reubeno:


hartsock:
@reubeno — following up here from #1183, where you'd pointed me at this PR. Short version:
this is the mechanism we need, and I've closed my competing #1184.

One concrete gap, since you asked for feedback from a real use case: **the `exec` builtin
isn't covered.** This PR touches no `brush-builtins/` files, and
`brush-builtins/src/exec.rs` composes its own `std::process::Command` and calls
`CommandExt::exec`, replacing the shell process image without ever reaching
`execute_external_command`. So a `CmdExecFilter` that denies `rm` is still replaceable by
`exec /bin/rm`.

I have a fix and a regression test for it. The test has to run in a dedicated child process —
a successful `exec` replaces the process image, so a broken implementation takes the test
runner with it. Against unauthorized `exec.rs` it fails with `the denied program ran and
replaced the shell: /tmp/.../marker exists`.

Worth noting the asymmetry it exposes: a *successful* `exec` never returns, so
`post_external_cmd` can never fire for it. An audit log assuming pre/post pairing will
mis-attribute it. Probably worth a line in `filter-architecture.md`.

**I've already done the rebase**, since it was the fastest way to check the above rather
than guess. Cherry-picked onto `30a3bce`: only two files conflicted (`commands.rs`,
`shell/execution.rs`), and the result is green — fmt/clippy/schemas clean, unit 390/390,
integration 2709/2709, compat unchanged. Two things in `execute_via_external` needed carrying
forward that this branch predates: `main`'s `argv0` override, and its `$_` update after
external commands. Branches are on my fork if useful — `demo/972-rebased`, and
`demo/972-exec-fix` with the `exec` coverage on top. Take, adapt, or ignore; it's your design
and I'd rather you drive it.

Two smaller things I hit while porting a real policy onto it:

- **`ExternalCmdParams` exposes neither the command name as written nor `argv0`** — only the
  `ExternalCommand` (program, args, envs, cwd). So a policy can't match on the spelling the
  user typed, and can't see an `exec -a impostor /usr/bin/touch` override. For an
  allow-listing policy those are separately load-bearing; I had to drop two tests in the port
  because the information isn't there.
- **`cmd_exec_filter` is `#[serde(skip, default = ...)]`.** A serialized shell comes back with
  a freshly-defaulted filter rather than the one that was installed. If the default is the
  no-op, a round-trip silently unconfines; if it isn't, a stateful policy still comes back
  without its state. Worth at least a documented contract.

PR #974, opened 2026-02-01
feat: scaffolding for winnow parser

reubeno:
Scaffolding for the winnow parser without an implementation body (yet).

@lu-zero I took your latest changes, stripped out the actual parser implementation, leaving just the scaffolding. I also added a runtime command-line option all the way up through `brush-shell` and some new options that would allow us to run our full suite of compat tests against either implementation, e.g.:

```
$ # Run brush shell with the winnow parser
$ cargo run --features=experimental -- --experimental-parser

$ # Run all tests and enable the winnow parser for tests run through our YAML-based test harness
$ BRUSH_ARGS="--experimental-parser" cargo nextest run --features=experimental
```

What do you think about this?

What I'm looking for is to get enough of the scaffolding in first that we can immediately start with an implementation that is 0% compatible and incrementally improve it. With each improvement that's made, we can track progress. Ideally I'd like to update PR CI workflows to run tests with the winnow parser too, but not gate; we can include in the report which percentage of tests are succeeding vs. failing on the winnow parser to track its progress and ensure we're monotonically moving forward.

Once the parser is working well *enough* to use a lot more, but not still not ready to be the *default* we can individually opt into using it by default in an interactive shell via the `--experimental-parser` option (included when compiling with the right features).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -438,6 +438,7 @@ dependencies = [
  "thiserror",
  "tracing",
  "utf8-chars",
+ "winnow",
 ]
 
 [[package]]
@@ -4129,6 +4130,9 @@ name = "winnow"
 version = "0.7.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a5364e9d77fcdeeaa6062ced926ee3381faa2ee02d3eb83a5c27a8825540829"
+dependencies = [
+ "memchr",
+]
 
 [[package]]
 name = "winsafe"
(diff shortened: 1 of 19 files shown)


lu-zero:
We can do that this way, but the winnow parser itself should be on par with peg at least for the situations I tested.

reubeno:
I'm fine to start bringing in the real implementation pieces shortly thereafter, but I'd like to see the plumbing set up first and for us to actually hook it up to the interpreter and have the capability to run scenario / end-to-end tests. 

Do you have any feedback on the changes I've made outside of brush-parser? The brush-parser changes are taken directly from your PR.

lu-zero:
it does look fine to me :)

PR #982, opened 2026-02-08
refactor: use let chains

reubeno:


xtqqczze:
There are still at least 62 instances of [`clippy::collapsible_if`](https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if) that could be fixed.

reubeno:
> There are still at least 62 instances of [`clippy::collapsible_if`](https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if) that could be fixed.

You've probably noticed that's one of the few handfuls of clippy rules we mark as allowed. I don't recall offhand how much that was intentional vs. practical to get the code to a baseline.

We're open to contributions that make changes in this area, provided that they improve the quality of the code base. (There are some other clippy rules that seem less clearly valuable to address.)

xtqqczze:
I was planning to address the `collapsible_if` lints, but I noticed that in many cases the code could benefit from a refactor instead; for example, in https://github.com/reubeno/brush/pull/1001.

xtqqczze:
> You've probably noticed that's one of the few handfuls of clippy rules we mark as allowed. I don't recall offhand how much that was intentional vs. practical to get the code to a baseline.

That consideration may have shifted as `collapsible_if` can now suggest let chains.

reubeno:
> > You've probably noticed that's one of the few handfuls of clippy rules we mark as allowed. I don't recall offhand how much that was intentional vs. practical to get the code to a baseline.
> 
> 
> 
> That consideration may have shifted as `collapsible_if` can now suggest let chains.

That's good to know! It sounds worthwhile looking at. I'm open to re-enabling it if someone is interested and willing to reduce some of the existing flagged issues first.

PR #987, opened 2026-02-11
fix(env): Return None in the get_str on unset vars

lu-zero:
Fixes IFS handling when unset in local scope and potentially makes the HOME fallback more correct.

The get_str method was incorrectly converting unset variables to empty strings instead of returning None. This caused issues with IFS handling where unset IFS variables should fall back to default behavior but were instead treated as empty IFS.


reubeno:
Thanks for the fix, @lu-zero!

Is this separate from #988, then? Could you please add a YAML-based test case that fails without this fix?

lu-zero:
we can close this one and keep the other. (sorry for not noticing the reply till now)

PR #988, opened 2026-02-11
Improve IFS support

### LINKED ISSUE #295: only honor IFS for splitting portion of word that had expansions
The POSIX standard specifies that field splitting should be done before quote removal, so in theory the following
```sh
IFS=\'
echo a'b'c
```
should print `a b c`. Although it looks like bash doesn't do this even for "simpler" cases, like 
```sh
IFS=:
echo a:b:c
```
outputs `a:b:c`. But it works if we do
```sh
x=a:b:c
IFS=:
echo $x
```
which outputs `a b c`.

lu-zero:
I tried to use brush(winnow) to parse the portage tree and this is one of the issue found and solved. 

Fixes: #295

### DIFF
--- a/brush-core/src/env.rs
+++ b/brush-core/src/env.rs
@@ -266,6 +266,9 @@ impl ShellEnvironment {
     /// Tries to retrieve the string value of the variable with the given name in the
     /// environment.
     ///
+    /// Returns `None` if the variable doesn't exist or if its value is unset.
+    /// See [`ShellValue::try_get_cow_str`] for details on string conversion.
+    ///
     /// # Arguments
     ///
     /// * `name` - The name of the variable to retrieve.
@@ -276,7 +279,7 @@ impl ShellEnvironment {
         shell: &Shell<SE>,
     ) -> Option<Cow<'_, str>> {
         self.get(name.as_ref())
-            .map(|(_, v)| v.value().to_cow_str(shell))
+            .and_then(|(_, v)| v.value().try_get_cow_str(shell))
     }
 
     /// Checks if a variable of the given name is set in the environment.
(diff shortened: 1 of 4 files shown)


brush-core/src/env.rs
reubeno, line 273: This is a commonly enough used function that we should document this behavior explicitly in its header comment. We don't need a big paragraph, just to bring clarity that `None` is returned if the variable is not set, even if it's been declared.

brush-core/src/expansion.rs
```diff
@@ -608,23 +608,46 @@ impl<'a, SE: extensions::ShellExtensions> WordExpander<'a, SE> {
...
-            tracing::debug!(target: trace_categories::EXPANSION, "  => brace expanded to '{brace_expanded}'");
+            tracing::debug!(target: trace_categories::EXPANSION, "  => brace expanded to {brace_words:?}");
         }
 
-        // Expand: tildes, parameters, command substitutions, arithmetic.
-        let mut expansions = vec![];
-        for piece in brush_parser::word::parse(brace_expanded.as_ref(), &self.parser_options)? {
-            let piece_expansion = self.expand_word_piece(piece.piece).await?;
-            expansions.push(piece_expansion);
+        // When there's no actual brace expansion (single word), preserve all expansion
```
reubeno: The comment says "when there's no actual brace expansion" but it and the code below only checks for how many words came back. `{1..1}` is a concrete example where brace expansion *does* happen but only a single word is returned (i.e., `1`).

reubeno (commented):
A few comments... but apart from them I think I understand what this is about and why it would fix some cases. Could you update the PR description or commits with a brief summary of what the actual issue(s) were? I want to make sure I really understand, and it will also be useful for composing the merge commit. (Your original commit for the env fix had a great description. The other commit was a bit more sparse in its details.)

reubeno:
I've taken the liberty of rebasing this against latest `main` and also adding a few layered commits to it. Planning to merge as long as it passes PR checks.

PR #990, opened 2026-02-13
add an ls builtin

cataggar:
I'm not intending to merge this. I was just doing some research. We have a lot of automation that uses `bash` and `make`, but none of it runs on Windows even though the other tooling is compatible. I was exploring switching to `brush` and [just](https://github.com/casey/just), but on Windows there are a few commands that are not available. These are not a part of bash, but useful for our bash scripts. Nushell offer `ls` and this is based on theirs.

I'll close this soon. I just wanted to bring this up for discussion.

### DIFF
--- a/brush-experimental-builtins/Cargo.toml
+++ b/brush-experimental-builtins/Cargo.toml
@@ -14,7 +14,8 @@ rust-version.workspace = true
 bench = false
 
 [features]
-default = ['builtin.save']
+default = ['builtin.ls', 'builtin.save']
+'builtin.ls' = []
 'builtin.save' = ["brush-core/serde", "dep:serde_json"]
 
 [lints]
--- a/brush-experimental-builtins/src/lib.rs
+++ b/brush-experimental-builtins/src/lib.rs
@@ -1,5 +1,7 @@
 //! Experimental builtins.
 
+#[cfg(feature = "builtin.ls")]
+mod ls;
 #[cfg(feature = "builtin.save")]
 mod save;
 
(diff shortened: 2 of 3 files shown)


reubeno:
@cataggar I've considered something like this previously -- particularly for platforms (like Windows) that don't have standard `coreutils` available.

Should you really be interested in this, I think a winning approach could be to wrap some/all/most of the commands in [`uutils`](https://github.com/uutils/coreutils) (a.k.a. the Rust reimplementation of `coreutils`) to be `brush` builtins. I *think* each of the commands implemented (e.g., cat, date, ls) are programmatically embeddable and could be projected into builtin implementations, and available in one of the `brush` builtins crates behind a flag.

oech3:
`make` would require more tools e.g. `grep`,`uutils/{find,diff}utils,sed`.
But uutils does not have `grep`.

oech3:
Also Windows needs to care about `*.exe` at file names if you are on non-cygwin `make`.
uutils does not support it natively i.e. `cp cp.exe cp2` does not copy `cp.exe`. cygwin target exists, but needs cygwin specific *.dll.

reubeno:
> `make` would require more tools e.g. `grep`,`uutils/{find,diff}utils,sed`.
> 
> But uutils does not have `grep`.

Agreed that `make` would be a heavier lift (without using Cygwin, say). Per the original submitter's post, though, a `brush` + `just` combo could be an easier (and rusty!) adaptation.

cataggar:
> Should you really be interested in this, I think a winning approach could be to wrap some/all/most of the commands in [uutils](https://github.com/uutils/coreutils) (a.k.a. the Rust reimplementation of coreutils) to be brush builtins. I think each of the commands implemented (e.g., cat, date, ls) are programmatically embeddable and could be projected into builtin implementations, and available in one of the brush builtins crates behind a flag.

That sounds like the right direction.

PR #994, opened 2026-02-17
fix(test): resolve bash to absolute path before running compat tests

lu-zero:
When the test harness clears the environment with env_clear(), PATH gets reset which can cause 'bash' to resolve to /bin/bash (macOS system bash 3.2) instead of the expected modern bash from Homebrew.

This fixes many false test failures related to features unsupported in old bash versions (e.g., declare -A for associative arrays).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,6 +492,7 @@ dependencies = [
  "uuid",
  "version-compare",
  "walkdir",
+ "which",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


reubeno:
Change looks good, just some complaints from the linter in the checks?

In case it helps we now have a handful of `cargo xtask` commands that mimic what the PR checks run.  If someone is enterprising we could wire them up as pre-commit checks. (I'm curious to try using the `prek` project.)

reubeno:
FYI-- If the nix tests look to be failing consistently, we'll ignore them for now and separately investigate. They shouldn't block other tests, nor merging.

PR #995, opened 2026-02-17
Heredoc quote preservation

lu-zero:


brush-core/src/expansion.rs
```diff
@@ -525,6 +546,28 @@ impl<'a, SE: extensions::ShellExtensions> WordExpander<'a, SE> {
...
+    /// Apply parameter expansion, command substitution, and arithmetic expansion
+    /// to a heredoc body, preserving literal quotes.
+    async fn basic_expand_heredoc_to_str(&mut self, word: &str) -> Result<String, error::Error> {
+        Ok(String::from(self.basic_expand_heredoc(word).await?))
+    }
+
+    async fn basic_expand_heredoc(&mut self, word: &str) -> Result<Expansion, error::Error> {
+        tracing::debug!(target: trace_categories::EXPANSION, "Heredoc expanding: '{word}'");
+
+        if !word.contains(['$', '`', '\\']) {
```
reubeno, line 587: It would probably help for posterity to get some comments in this function's body that capture why it's not shared with the standard expand function (e.g., no brace expansion).

brush-shell/tests/cases/compat/here.yaml
```diff
@@ -114,7 +114,6 @@ cases:
       echo "${test1}"
 
   - name: "Here doc in a command substitution with quotes"
-    known_failure: true # Issue #421
```
reubeno, line 117: Thanks for resolving this long-reported issue!

Are there other cases we should add here that would capture some of the other differences? For example, I see brace expansion is something else that differs.

reubeno:
@lu-zero Thanks for working through this one! I left a couple of comments but it's otherwise looking good.

The NixOS error should also be gone when you next rebase against `main`.

brush-shell/tests/cases/compat/here.yaml
lu-zero: the wip branch `for-portage-repo` has plenty plus the fixes :)

reubeno: My general preference is to eagerly pull forward as many YAML tests as possible but mark them as `known_failure: true`. There's no reason to block their check-in, they decouple from the fixes well, and they'll serve as a positive sign that the tests do, in fact, fail without the fix and succeed with it.

brush-core/src/expansion.rs
reubeno: @lu-zero I think this round of (what looks like) AI-generated docs went in the wrong direction. I wasn't looking for a self-contained spec on what heredoc expansion looks like -- per my previous comment I was more looking for some contextually relevant summaries of how heredoc expansion differs from normal expansion, both to document the differences and to justify why it needs its own entry points.

brush-shell/tests/cases/compat/here.yaml
lu-zero: I can extract them in a separate PR then.

brush-core/src/expansion.rs
lu-zero: I can reduce the summary to "heredoc expansion do not break on `"` and `'`" but the full summary felt good to not have to read back to the posix spec. Let me know which way you prefer.

reubeno:
Merged via #1014

PR #998, opened 2026-02-17
chore(clippy): fix nightly lints

xtqqczze:


reubeno:
Thanks for proactively contributing this!

The only check failure is unrelated to your change; it's something we're tracking as an upstream issue caused by recent runner changes. We won't worry about that.

PR #1003, opened 2026-02-19
fix: readonly should operate on global scope

lu-zero:
In bash, 'readonly VAR' always operates on the global scope, even when called inside a function. Previously, brush was treating 'readonly' like 'declare -r', which creates a local variable.

This fixes the issue where 'readonly ARR' inside a function would create an empty local variable instead of marking the existing global array as readonly.


reubeno:
Good catch! LGTM

PR #1004, opened 2026-02-19
fix(peg): support legacy $[expr] arithmetic syntax

lu-zero:
Bash supports the legacy `$[expr]` syntax for arithmetic expansion as an alternative to `$((expr))`. 

Apparently it moved from being deprecated to be grandfathered in, so we should support it as well.

Add support for this syntax in the PEG parser.

### DIFF
--- a/brush-test-harness/src/execution.rs
+++ b/brush-test-harness/src/execution.rs
@@ -137,6 +137,8 @@ impl TestCase {
         // Clear all environment vars for consistency.
         test_cmd.args(&self.args).env_clear();
 
+        // Set locale to C for consistent behavior across systems.
+        test_cmd.env("LC_ALL", "C");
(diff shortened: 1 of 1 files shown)


reubeno:
Is touching just the peg parser sufficient? If I recall correctly the tokenizer has some logic to deal with the fact that $((...)) enters an arithmetic context, which changes the meaning of something things. For example << is a left shift operator and not a bash redirection operator...

With that said, that logic isn't quite right. (This should be a note for the new parser too. Things parse differently in arithmetic contexts, which can appear in array indices, arithmetic for loops, arithmetic expressions and statements, substring offsets, etc.)

lu-zero:
it is not. fixes coming soon.

reubeno (approved):

PR #1005, opened 2026-02-19
fix(unset): add support for negative array indices

lu-zero:
Add bash-compatible negative indexing for array unset operations. Negative indices count from the end of the array (-1 is last element). Out-of-bounds negative indices produce a 'bad array subscript' error.

### DIFF
--- a/brush-core/src/variables.rs
+++ b/brush-core/src/variables.rs
@@ -395,7 +395,7 @@ impl ShellVariable {
 
         match &mut self.value {
             ShellValue::IndexedArray(arr) => {
-                let key: u64 = array_index.parse().unwrap_or(0);
+                let key = get_key_for_indexed_array(arr, array_index.as_str())?;
 
                 if append {
                     let existing_value = arr.get(&key).map_or_else(|| "", |v| v.as_str());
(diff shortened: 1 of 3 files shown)


reubeno:
Is unset the only command that takes negative indices (and which we don't already support)?

brush-builtins/src/unset.rs
```diff
@@ -67,7 +67,23 @@ impl builtins::Command for UnsetCommand {
                             context.shell.env_mut().unset(name.as_str())?.is_some()
                         }
                         brush_parser::word::Parameter::NamedWithIndex { name, index } => {
-                            unset_array_index(context.shell, name.as_str(), index.as_str())?
+                            match unset_array_index(context.shell, name.as_str(), index.as_str()) {
```
reubeno: Is this special-casing of the specific error needed, or is there a way for us to keep the logic simpler? (If we're okay diverging our error message from `bash` is it still required?)

brush-shell/tests/cases/compat/builtins/unset.yaml
```diff
@@ -90,6 +90,63 @@ cases:
       declare -p myarray
       echo
 
+  - name: "Unset array element with negative index"
```
reubeno, line 93: Thanks for adding all of these tests! :smile:

brush-core/src/variables.rs
```diff
@@ -511,7 +511,19 @@ impl ShellVariable {
             ShellValue::String(_) => Err(error::ErrorKind::NotArray.into()),
             ShellValue::AssociativeArray(values) => Ok(values.remove(index).is_some()),
             ShellValue::IndexedArray(values) => {
-                let key = index.parse::<u64>().unwrap_or(0);
+                let original_index = index.parse::<i64>().unwrap_or(0);
+                let mut index_value = original_index;
+
+                #[expect(clippy::cast_possible_wrap)]
+                if index_value < 0 {
```
reubeno: Do we need negative subscript handling in other operations that affect indexed arrays?

If so, what do you think about factoring out a helper routine to resolve the actual index from the input string (and also ensure it's in-bounds)?

reubeno:
@lu-zero FYI -- in the interests of expediency, I've pushed some adjustments to your PR branch that also expand the fix to cover array setting too. Once the checks pass, I'll merge the change. Thanks for working on this!

PR #1006, opened 2026-02-20
feat(parser): handle inline comments in PEG command substitution

lu-zero:
Exclude '#' from literal text when inside command context and add comment handling to command_piece() rule. This allows comments within
 command substitutions to be properly parsed.

### DIFF
--- a/brush-parser/src/word.rs
+++ b/brush-parser/src/word.rs
@@ -815,7 +815,8 @@ peg::parser! {
         rule unquoted_literal_text_piece<T>(stop_condition: rule<T>, in_command: bool) =
             is_true(in_command) extglob_pattern() /
             is_true(in_command) subshell_command() /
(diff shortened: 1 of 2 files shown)


reubeno (changes requested):
I ran the 2 added YAML tests without the fix:

* `Comment in command substitution without newline fails` - it *does* indeed fail, just with a different error than `bash` (so I'd expect the test to pass with `ignore_stderr: true`)
* `Comment in command substitution with newline` - passes even without any code changes

Can you identify what case, in particular, you believe is a problem in the current code base?

lu-zero:
If it passes without the change, it is just me repeating the fix I had to make in winnow, but that's strange I was quite sure the original from Gentoo failed even with the stand alone tokenizer...

lu-zero:
we can just keep the tests and consider later that bit if we want to emit comments and keep the peg parser

PR #1029, opened 2026-02-28
feat: add coproc AST and placeholder execution

lu-zero:
Add AST support for coproc command:
- Add CoprocClauseCommand struct with name, body, and loc fields
- Add CoprocClause variant to CompoundCommand enum
- Implement Execute for CoprocClauseCommand returning unimplemented error

Note: The PEG parser does not support coproc. This provides the AST structure for future parser implementations.

### DIFF
--- a/brush-core/src/interp.rs
+++ b/brush-core/src/interp.rs
@@ -693,19 +693,20 @@ impl Execute for ast::CompoundCommand {
             Self::UntilClause(u) => (WhileOrUntil::Until, u).execute(shell, params).await,
             Self::Arithmetic(a) => a.execute(shell, params).await,
             Self::ArithmeticForClause(a) => a.execute(shell, params).await,
-            Self::CoprocClause(c) => c.execute(shell, params).await,
+            Self::Coprocess(c) => c.execute(shell, params).await,
         }
     }
 }
 
 #[async_trait::async_trait]
-impl Execute for ast::CoprocClauseCommand {
+impl Execute for ast::CoprocessCommand {
     async fn execute(
         &self,
         _shell: &mut Shell<impl extensions::ShellExtensions>,
         _params: &ExecutionParameters,
     ) -> Result<ExecutionResult, error::Error> {
-        error::unimp("coproc")
+        // Coprocesses are not yet implemented in this shell.
+        error::unimp("coproc is not yet supported in this shell")
     }
 }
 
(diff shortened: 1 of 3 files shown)


brush-core/src/interp.rs
```diff
@@ -1178,6 +1190,19 @@ impl<SE: extensions::ShellExtensions> ExecuteInPipeline<SE> for ast::SimpleComma
         if let Some(CommandArg::String(cmd_name)) = args.first().cloned() {
             let mut stderr = params.stderr(&context.shell);
 
+            // Set $_ to the last argument before executing the command.
```
reubeno: Did this leak in from a different PR?

brush-parser/src/ast.rs
```diff
@@ -446,6 +446,8 @@ pub enum CompoundCommand {
     WhileClause(WhileOrUntilClauseCommand),
     /// An until clause, which loops until a condition is met.
     UntilClause(WhileOrUntilClauseCommand),
+    /// A coproc clause, which runs a command as a coprocess.
+    CoprocClause(Box<CoprocClauseCommand>),
```
reubeno: Why a `Box`?

brush-parser/src/ast.rs
```diff
@@ -845,6 +851,46 @@ impl Display for ElseClause {
...
+)]
+pub struct CoprocClauseCommand {
+    /// The optional name for the coprocess.
+    #[cfg_attr(
+        any(test, feature = "serde"),
+        serde(skip_serializing_if = "Option::is_none", default)
+    )]
+    pub name: Option<Word>,
+    /// The command to run as a coprocess (can be simple or compound).
+    pub body: Box<Command>,
```
reubeno, line 869: Why `Box`?

brush-core/src/interp.rs
lu-zero: Looks like so I tried to extract two low hanging fruits

reubeno:
FYI @lu-zero I've applied some minor renaming to the new AST nodes. I took a look back at the other commands and it seems to be closer to, say, `Subshell` and `SubshellCommand` -- and less like one of the "clause" commands ... hence the names `Coprocess` and `CoprocessCommand`.

I also went back through the docs and this seems to be in line with it.

I hope that won't cause you too much rework anywhere you've already taken a dependency on the names.

PR #1030, opened 2026-02-28
feat: implement $_ well-known variable for last command argument

### LINKED ISSUE #479: Difference from bash: `:` followed by `$_`
I can use `bash` (and `zsh`, but not POSIX `sh`) to find the number of days in current month:

```bash
$ bash -c ': $(cal); printf "»%s«\n" "$_";'
»31«
```

However, `brush` doesn't seem to assign anything to `$_` in this context:

```bash
$ brush -c ': $(cal); printf "»%s«\n" "$_";'
»«
```

----

```sh
$ bash --version | head -1
GNU bash, version 5.2.37(1)-release (x86_64-pc-linux-gnu)
$ brush --version
brush 0.2.17 (git:c6993c6)
```

lu-zero:
Add support for the $_ well-known variable, which expands to the last argument of the previous command.

Resolves #479.

### DIFF
--- a/brush-shell/tests/cases/compat/special_parameters.yaml
+++ b/brush-shell/tests/cases/compat/special_parameters.yaml
@@ -75,9 +75,9 @@ cases:
   - name: "$_ - nested function dispatch"
     stdin: |
       inner() { echo innerdone; }
-      outer() { inner ix iy; }
+      outer() { inner ia ib; }
       echo seed
-      outer ox oy
+      outer oa ob
       echo "[$_]"
 
   - name: "$_ - assignment-only function body"


brush-shell/tests/cases/compat/special_parameters.yaml
```diff
@@ -1,5 +1,14 @@
 name: "Special parameters"
 cases:
+  - name: "$_ - last command argument"
```
reubeno: Thanks for implementing this! It was long-requested as #479.

Do the tests added in #480 pass now? Are there any other tests that are worth adding to ensure appropriate scoping/update of the variable across subshell/command/pipeline boundaries?

lu-zero: Let me think about it tomorrow.

lu-zero:
Now it should be cleaner

brush-core/src/commands.rs
```diff
@@ -548,6 +565,9 @@ impl<'a, SE: extensions::ShellExtensions> SimpleCommand<'a, SE> {
             &self.args[1..],
         );
 
+        // Update $_ after command execution.
+        shell.update_last_arg_variable(last_arg.as_deref());
```
reubeno: In each of these cases, we have an owned `String` of the last argument (or at least an `Option` of one). Couldn't we save a string clone by having the setter in `Shell` require the owned `String`?

brush-shell/tests/cases/compat/special_parameters.yaml
```diff
@@ -1,5 +1,14 @@
...
+    stdin: |
+      : 0
+      echo "$_"
+      : 0 1 2
+      echo "$_"
+      echo "first command"
+      echo "$_"
```
reubeno: Could you please add a few test cases that cover other scenarios? e.g.:

* Pipeline with multiple commands
* Commands with `&&` / `||`
* Command substitution / subshell
* Builtin invocation
* Function invocation
* Commands that fail

We typically don't need to go so broad but because there are multiple code paths for handling different kinds of commands here, I think these will be important to ensure compatibility (and prevent us from regressing it in the future).

reubeno (commented):
Looking pretty good! Just a few comments.

Issue #677, opened 2025-09-12
Bash incompatibility: printf does not recognize "%(datefmt)T"


kseistrup:
In `bash` (and `ksh` and `osh`) I can use `printf "%(datefmt)T"` to output a datetime formatted by the `datefmt` specifier, e.g.:

```bash
$ printf '%(%F %T)T\n'
2025-09-12 15:54:11
```

This does not work in brush 0.2.23 (git:7967c78):

```bash
$ printf '%(%F %T)T\n'
ERROR printf: printf parsing error: %(: invalid conversion specification
```

----

From bash(1):

```txt
      printf [-v var] format [arguments]
          ⋮
              %(datefmt)T
                     causes  printf  to  output  the date-time string resulting
                     from using datefmt as a  format  string  for  strftime(3).
                     The  corresponding argument is an integer representing the
                     number of seconds since the epoch.  This format  specifier
                     recognizes  two special argument values: -1 represents the
                     current time, and -2 represents the time the shell was in‐
                     voked.  If no argument is specified, conversion behaves as
                     if -1 had been supplied.  This  is  an  exception  to  the
                     usual printf behavior.
```

reubeno:
Thanks for reporting this issue, @kseistrup.

We use the `uucore` crate from the `uutils` project to provide our `printf` builtin implementation. As you note, though, there are (at least?) 3 `printf` extensions unique to `bash`'s builtin (i.e., not supported by `/usr/bin/printf`):

* `%b` - expands escape sequences like `echo -e`
* `%q` - auto-quotes the argument so it can be used as input
* `%(datefmt)T` - formats a date/time string using a `strftime`-style format

To resolve this, I believe we'd either need to:

(1) Work with the `uutils` project and see if they'd be willing to accept contributions of these custom formats as optional extensions (off by default for standard `coreutils` usage).
(2) Figure out how to replace/handle them before passing the format string and arguments to `uucore::format()`.

We presently have some special-case handling of `"%q"` that's basically a hack for some common cases, but we need to properly resolve this.

oech3:
See GNU's thought. https://github.com/coreutils/coreutils/issues/142#issuecomment-3677768479 .
If GNU does not add it, I think uutils should not add it too (until BSD coreutils add it).

@kseistrup What application does actually use it?

kseistrup:
[...]

PR #1031, opened 2026-03-01
feat: add optional bundling of coreutils builtins behind experimental feature flag

cataggar:
This is a first pass at adding the coreutils as builtins per the suggestion https://github.com/reubeno/brush/pull/990#issuecomment-3895790164 .

cargo install --locked brush-shell --features coreutils-builtins

```
brush.exe-0.4$ help
brush version 0.3.0 (git:4db4716) - https://github.com/reubeno/brush

The following commands are implemented as shell built-ins:
   .                      factor                 set
   :                      false                  sha1sum
   [                      fc                     sha224sum
   alias                  fg                     sha256sum
   arch                   fmt                    sha384sum
   b2sum                  fold                   sha512sum
   base32                 getopts                shift
   base64                 hash                   shopt
   basename               head                   shred
   basenc                 help                   shuf
   bg                     history                sleep
   bind                   hostname               sort
   break                  jobs                   source
   brushctl               join                   split
   brushinfo              let                    sum
   builtin                link                   sync
   caller                 ln                     tac
   cat                    local                  tail
   cd                     logout                 tee
   cksum                  ls                     test
   comm                   mapfile                times
   command                md5sum                 touch
   compgen                mkdir                  tr
   complete               mktemp                 trap
   compopt                more                   true
   continue               mv                     truncate
   cp                     nl                     tsort
   csplit                 nproc                  type
   cut                    numfmt                 typeset
   date                   od                     uecho
   dd                     paste                  ufalse
   declare                popd                   unalias
   df                     pr                     uname
   dir                    printenv               unexpand
   dircolors              printf                 uniq
   dirname                ptx                    unlink
   dirs                   pushd                  unset
   disown                 pwd                    uprintf
   du                     read                   upwd
   echo                   readarray              utest
   enable                 readlink               utrue
   env                    readonly               vdir
   eval                   realpath               wait
   exit                   return                 wc
   expand                 rm                     whoami
   export                 rmdir                  yes
   expr                   seq
brush.exe-0.4$ uname
Windows_NT
```


reubeno:
Very cool! Thanks for putting this together :smile: With all these commands present, how well does the shell seem to work on Windows? (I'm assuming from your output above that's where you've tested this?)

I skimmed through the changes and they look quite reasonable. (I wish there were a way to dynamically get the command short-description from `uutils`, but I don't think it's a deal-breaker.) I'll look at the macros a bit more closely later this weekend.

I've thought about it a bit more since and one thing we'll need to keep in mind is that much of `brush`'s shell state isn't reflected back to the hosting process's OS state. As specific examples: when a variable is set in `brush` it's not observable via OS env APIs; also, file redirections aren't actually reflected back to the real host file descriptor table or OS-level notions of stdout/stderr/stdin. That's a benefit for OS portability and for being able to spawn subshells in-process, but I could foresee some issues with coreutils commands not being able to "see" those changes.

With that said, this seems interesting to enough to keep pursuing. I'm inclined to bring these in under an "experimental" qualifier, as we've done with some of the other speculative features/extensions. (That just might entail prefixing the feature in `brush-shell` with the `experimental-` prefix. We should also explore the `uutils` code base to see if there might be some way of overriding env and/or stdio access. At worst, we could explore other creative options that entail running the coreutils commands in a child process despite being available in-binary.

cataggar:
I created a minimal Azure Linux Dockerfile to try it out and it works well.
https://github.com/cataggar/brush/pull/1

reubeno:
@cataggar I'd like to move toward getting this PR merged but would like to qualify the coreutils integration as "experimental" -- particularly because of known challenges with stdio redirection.

Are you up for making some adjustments? Alternatively, I'm open to making some of the adjustments myself if you're okay with me pushing changes to your PR branch. In particular here are the items that we'll need to sort out:

* Rename feature `coreutils-builtins` to `experimental-coreutils-builtins` to indicate its experimental status.
* Get PR checks passing (I see the format checks are failing); `cargo xtask ci pre-commit` can be used locally to help catch some of these.
* Rebase on latest `main`.
* Ensure that the license file is symlink'd under the new crate dir like it is for others

oech3:
You should exclude ~unix specific~some progs from ~windows and~ wasm* target. (cygwin target is OK).

reubeno:
> You should exclude ~unix specific~some progs from ~windows and~ wasm* target. (cygwin target is OK).

@oech3 can you share more about what you're thinking here? Or a concrete example?

oech3:
https://github.com/uutils/coreutils/blob/de5e4f54f224a25875bdc207cd57f035a58fbe0a/Cargo.toml#L174-L225
You cannot build other utils for wasip1.

oech3:
uutils has policy to avoid `process::exit` for this usecase. But still used. Does it exit `brush` itself?

reubeno:
> uutils has policy to avoid `process::exit` for this usecase. But still used. Does it exit `brush` itself?

Yes, I'd expect it would. It's for reasons like this that I'd consider this integration experimental. The input/output redirection issues I mention above are another class of challenges.

casey:
I'm the author of `just`, and I just (ha ha, never gets old) stumbled across this PR and the original conversation in #990.

There's been a longstanding desire to embed a shell and common utilities in `just` in order to make it truly cross platform. I was researching what options are available and found this issue.

The combination of brush and `uutils/coreutils` as a Rust-native busybox is really appealing!

There are definitely potential issues. The binary would be absolutely gargantuan, and `just` has very strict backwards compatibility requirements. But this could be sidestepped by having users install the brush + coreutils binary separately and use it via `set shell := ['brush']` in their justfile.

Anyways, just wanted to say that I think this is a super cool direction to explore, and I'm very interested in how things develop.

reubeno:
> I'm the author of `just`, and I just (ha ha, never gets old) stumbled across this PR and the original conversation in #990.
> 
> There's been a longstanding desire to embed a shell and common utilities in `just` in order to make it truly cross platform. I was researching what options are available and found this issue.
> 
> The combination of brush and `uutils/coreutils` as a Rust-native busybox is really appealing!
> 
> There are definitely potential issues. The binary would be absolutely gargantuan, and `just` has very strict backwards compatibility requirements. But this could be sidestepped by having users install the brush + coreutils binary separately and use it via `set shell := ['brush']` in their justfile.
> 
> Anyways, just wanted to say that I think this is a super cool direction to explore, and I'm very interested in how things develop.

Thanks for reaching out, @casey! (Big fan of `just` by the way :smile:. We don't use it for `brush` currently but I've used it for other projects.)

I'm also interested in the idea of a(n emeddable?) variant build of `brush` that's more intentionally busybox-ish and self-contained. (For unix-y platforms I'm hesitant for the default `brush` binary to embed coreutils; I expect that users typically want to use their system installation of coreutils instead. We may choose otherwise on other platforms like Windows, though, on which those standard utilities generally aren't present.)

Do you have a point of view on which shell functionality / common utilities you and your users would want to see `just` have access to? For example, there's a fair bit of logic in shells like `bash` and `brush` to support interactive use, and which typically isn't required for scripts/automation. We've tried to keep interactive logic separatable in `brush`'s core crates; I've long wanted to build a version of `brush` that's limited to script execution, and which excludes all the code + dependencies only required for interactive usage.

casey:
> I'm also interested in the idea of a(n emeddable?) variant build of `brush` that's more intentionally busybox-ish and self-contained. (For unix-y platforms I'm hesitant for the default `brush` binary to embed coreutils; I expect that users typically want to use their system installation of coreutils instead. We may choose otherwise on other platforms like Windows, though, on which those standard utilities generally aren't present.)

I think it makes sense for those to be different binaries. `brush` with no userland, and `brushbox` which is brush plus coreutils.

> Do you have a point of view on which shell functionality / common utilities you and your users would want to see `just` have access to? For example, there's a fair bit of logic in shells like `bash` and `brush` to support interactive use, and which typically isn't required for scripts/automation. We've tried to keep interactive logic separatable in `brush`'s core crates; I've long wanted to build a version of `brush` that's limited to script execution, and which excludes all the code + dependencies only required for interactive usage.

If it was embedded in `just`, than it would be nice to get the binary size down, so interactive logic could be stripped out. However, that's only to get the binary size down, and I think that would only be really important if brush were embedded. In the end it's probably more flexible to have a separate `brushbox` binary, in which case it's less important to get the binary size down, and users would probably want the interactive features in the standalone binary.

Then you would just do:

```
set shell := ["brushbox", "brush"]
```

In your justfile, or whatever the command line would be.

casey:
Oh, and sorry, I didn't fully answer this question:

> Do you have a point of view on which shell functionality / common utilities you and your users would want to see just have access to? 

I'm not actually sure about which utilities users would want. I think there are probably some obscure ones that don't get used frequently, but I'm not sure how to think of the tradeoff between a larger binary with all utilities, and a smaller binary without some obscure utilities.

oech3:
Can we bind uutils/date to https://github.com/reubeno/brush/issues/677 ?

brush-coreutils-builtins/src/lib.rs
```diff
@@ -0,0 +1,420 @@
...
+#[cfg(feature = "coreutils.yes")]
+coreutils_builtin!(YesCommand, uu_yes, "output a string repeatedly until killed");
+
+// ── Public API ──────────────────────────────────────────────────────────────
+
+/// Returns the set of coreutils built-in commands enabled by feature flags.
+#[allow(clippy::too_many_lines)]
+pub fn coreutils_builtins<SE: brush_core::extensions::ShellExtensions>()
+-> HashMap<String, builtins::Registration<SE>> {
+    let mut m = HashMap::<String, builtins::Registration<SE>>::new();
```
Elsie19, line 319:
```suggestion
    let mut m = HashMap::<&'static str, builtins::Registration<SE>>::new();
```

Is there no reason this could be used?

PR #1039, opened 2026-03-03
feat(expansion): implement nameref resolution for variable access

lu-zero:
When a variable is declared as a nameref (declare -n ref=target), accessing or assigning to $ref now correctly redirects to $target.

- Add resolve_nameref_target() helper to resolve nameref chains
- Resolve namerefs in expand_parameter for all variable access patterns:
  - Simple variable expansion ($ref)
  - Indexed array access (${ref[index]})
  - All indices (${!ref[@]})
  - All values (${ref[@]})
- Resolve namerefs in assign_to_parameter for write operations
- Resolve namerefs in apply_assignment for direct assignments

This enables proper support for nameref variables used in functions and scripts that pass array references.

Adds comprehensive YAML tests for nameref functionality including:
- Basic read/write through reference
- Array and associative array access
- Pass-by-reference function arguments
- Array iteration helpers

Updates existing helper functions to use nameref properly now that it is implemented.

### DIFF
--- a/brush-core/src/expansion.rs
+++ b/brush-core/src/expansion.rs
@@ -4,6 +4,7 @@ use std::borrow::Cow;
 use std::cmp::min;
 
 use brush_parser::word::{ParameterTransformOp, SubstringMatchKind};
+use itertools::Either;
 use itertools::Itertools;
 
 use crate::ExecutionParameters;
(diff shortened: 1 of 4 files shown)


reubeno:
Thanks for your patience with this one, @lu-zero.

I looked into it a bit -- this gets us some partial progress, but I think the nameref implementation needs to be plumbed through more deeply (and widely). There's other code paths that can assign to or query variables (e.g., builtins that update variables) and which also need to be enlightened to honor namerefs.

My perspective: `env.rs` / `variables.rs` will need to know about namerefs, or at least some fundamental aliasing. We'll also need to validate whether a nameref chain (a -> b -> c) needs to be handled, how to deal with cycles, etc.

If I have some time, I'm game to take a crack at the above -- using your changes as a starting point. That may not be until a future weekend, though.

Was this resolving something that was blocking something you were working on?

lu-zero:
it is one of the changes needed to support Gentoo portage parsing so yes ^^;

PR #1043, opened 2026-03-09
Regex contention

lu-zero:
when running my example of portage repo regenerating the whole repository cache I found out that cached is actually impairing performance.

We could later consider switching to https://crates.io/crates/moka or similar if I manage to hit more contention points.


reubeno:
Thanks for fixing the doc bugs too! The changes look reasonable to me.

Agreed also that we could look at moka, memoize, or other similar crates in the future.

We should probably also (at *some* point) look at all the other caching we enable by default.

PR #1044, opened 2026-03-10
feat: add /dev/null handling

puffnfresh:
This allows some scripts to run on Windows, which otherwise fail because the /dev/null device does not exist. This is similar to the existing special /dev/fd handling which Brush already contains.

### DIFF
--- a/brush-core/src/sys/windows/fs.rs
+++ b/brush-core/src/sys/windows/fs.rs
@@ -25,7 +25,7 @@ pub fn open_null_file() -> Result<std::fs::File, error::Error> {
 /// Gives the platform an opportunity to handle a special file path (e.g. `/dev/null`).
 pub fn try_open_special_file(path: &Path) -> Option<Result<std::fs::File, std::io::Error>> {
     if path == Path::new("/dev/null") {
-        Some(open_null_file().map_err(|e| std::io::Error::other(e)))
+        Some(open_null_file().map_err(std::io::Error::other))
     } else {
         None
     }


oech3:
I think it is something Cygwin should care about instead of shell.

puffnfresh:
@oech3 is that an argument specifically for `/dev/null` handling? Brush already does [/dev/fd](https://github.com/reubeno/brush/blob/ff73ad4fe9ff54aec270060792c672de23f59af9/brush-core/src/sys/unix/fd.rs#L16) and [Bash provides a few more](https://cgit.git.savannah.gnu.org/cgit/bash.git/tree/redir.c#n578). Where should we draw the line?

I think `/dev/null` is such a common idiom, it's a shame to have to use a Cygwin environment for many scripts which otherwise run fine on Windows.

oech3:
Hmm... I did't know that `bash` has such hacks.

oech3:
Are you planning to suggest this to upstram `bash` too?

reubeno:
@puffnfresh thanks for the contribution! It's good to know that some simple changes would unblock more scripts from running natively on Windows under `brush`.

I also support the idea of being able to use more scripts in a cross-platform way without requiring emulation layers like Cygwin.

Regarding the change, you're right that we have precedent with `/dev/fd` handling. It wasn't ideal but we were forced to do that because of `brush`'s design of abstracting and decoupling the file descriptors addressable within the shell from the host process's native file descriptors even on platforms like Linux and macOS.

I'm supportive of special handling for `/dev/null` and similar special files that are ubiquitously accessed in scripts, but we should scope that emulation to the platforms that require it (e.g., Windows).

We use the `sys` module to abstract platform-specific behaviors. What do you think of adding a new call into the `sys` module in this code path to give platform-specific code an opportunity to optionally handle special files. We would have the Unix-y platforms implement a no-op, but Windows and others could have a special case there for `/dev/null`.

This would allow Unix platforms to open the real null file, which is likely to look more consistent with other shells, while still enabling scripts to achieve the desired effect elsewhere.

reubeno:
Merged! Thanks for working on this, @puffnfresh. There was just one last clippy warning that I fixed for you so I could get it merged.

If you end up using this on Windows, would be great to hear what your experience is like / what you're able to do. If you're up for it, feel free to post issues / discussions here on GitHub or on our Discord.

Issue #997, opened 2026-02-17
`compgen -A command` don't contain builtins or keywords


takeshiD:
### brush version

brush 0.3.0

### bash version (for comparison)

GNU bash, version 5.3.9(1)-release (x86_64-pc-linux-gnu)

### Platform / OS

Linux

### Steps to reproduce

# bash
```bash
$ compgen -A command -- compgen
compgen
$ compgen -A builtin -- compgen
compgen
```

# brush
```bash
$ compgen -A command -- compgen

$ compgen -A builtin -- compgen
compgen
```

### Actual behavior (brush)

# brush
```bash
$ compgen -A command -- compgen
# return 0, but not output command
$ compgen -A builtin -- compgen
compgen
# this working
```

### Expected behavior

# brush
```bash
$ compgen -A command -- compgen
compgen
$ compgen -A builtin -- compgen
compgen
```

reubeno:
Good find! Thanks for reporting this.

Is this something you're interested in working on, @takeshiD ?

takeshiD:
@reubeno 
Yes, I'd like to investigate the cause of this issue.
If I'm able to identify and fix the problem, I'll submit a PR.

PR #1027, opened 2026-02-23
fix(builtins): add builtin commands to `compgen -A command`(#997)

takeshiD:
## Changes
- Kept external command completion logic.
- Added builtin-command names to candidates when completion.

## Result
Resolves issue #997, `compgen -A command` aligns more closely with `bash`.

PR #1047, opened 2026-03-14
fix(builtins): add reserved keywords to `compgen -A command`

takeshiD:
## Changes
- Add reserved keywords(e.g `[[`, `]]`, `case`, `if`, etc) to the result of `compgen -A command`.

## Testing
- Added tests for verify bash compatible, excluding `brushctl` and `brushinfo`.

## Result
Resolves issue #997, `compgen -A command` aligns more closely with `bash`.

This PR is follow-up to #1027.


takeshiD:
@reubeno 
It seems that the CI tests are failing on `linux/aarch64` and `linux/x86_64`, likely because X11 is present in those environments.

I haven't yet identified why X11 is included there, so I’m planning to open a separate issue to investigate the root cause.

In the meantime, would it make sense to mark these tests as `known_failure`?
This change still improves test coverage, so I think merging it would still be beneficial.

image

reubeno:
@takeshiD That's an interesting find! I'd suggest splitting the test case and having multiple test cases. You could construct test cases that constrain the command environment (e.g., constraining `PATH` such that visible commands are limited to what you control, constraining the token being completed, etc.) and confirm that keywords and builtins and commands come back--and hopefully that test or tests would pass. 

You could then take the current formulation and add it as a separate test marked as a known failure. That would increase the coverage, track a possibly legitimate problem deserving of follow-ups, and still have a passing test case ensuring we don't regress the functionality you've added.

takeshiD:
@reubeno Thanks for the excellent suggestion!

I had been assuming `--noprofile` and `--norc` were sufficient, which limited my perspective. I've now added deterministic tests with a constrained environment, and marked the original test as a `known_failure`.

takeshiD:
Opps,,,I'll investigate and fix it.

takeshiD:
@reubeno 
I added some test cases for constrained Path environments.

The unconstrained Path case fails due to including X11 command on Ubuntu(linux/aarch, linux/x86_64), while it passes on ArchLinux, Fedora, OpenSuse, AzureLinux, Macos.
Since this is strongly environment-dependent, I marked it as `skip` rather than `known_failure`.
I'll create a separate issue for this.

brush-shell/tests/cases/compat/builtins/compgen.yaml
```diff
@@ -232,10 +232,43 @@ cases:
...
+  - name: "compgen -A command"
+    # Temporarily marked unconstrained Path case as `skip: true` to keep CI stable.
+    # Environment-dependent due to host PATH contents(Ubuntu includes X11, while Arch and others do not).
+    skip: true
     stdin: |
+      compgen -A command | sort | grep -v brushctl | grep -v brushinfo
+
+  - name: "compgen -A command in a constrained PATH"
+    stdin: |
+      SORT=$(command -v ort)
```
reubeno: Is that a typo for `sort`?

brush-shell/tests/cases/compat/builtins/compgen.yaml
```diff
@@ -232,10 +232,43 @@ cases:
...
+      GREP=$(command -v grep)
+      PATH=$(pwd)
+      compgen -A command | $SORT | $GREP -v brushctl | $GREP -v brushinfo
```
reubeno: To simplify, what do you think about just excluding any matches for `brush`?

brush-shell/tests/cases/compat/builtins/compgen.yaml
takeshiD: Sorry, this is typo for `sort`. I'll fix it.

brush-shell/tests/cases/compat/builtins/compgen.yaml
takeshiD: I considered various scenarios, but excluding them with `^brush` seems to work fine. 
I'll make the changes and commit them.

reubeno:
Thanks for the fixes! FYI the PR branch looks like it got messy with merges. I'm going to squash-merge, which should ignore that. For next time, if you haven't been doing so, perhaps rebase to keep your work-in-progress branch updated?

PR #1070, opened 2026-03-22
fix: supports Android and 32-bit

YumeYuka:
Tested on Android devices and works fine

Commit by @ShIroRRen

### DIFF
--- a/brush-core/src/sys/unix/fs.rs
+++ b/brush-core/src/sys/unix/fs.rs
@@ -8,6 +8,10 @@ use crate::error;
 
 pub use std::os::unix::fs::MetadataExt;
 
+#[cfg(target_os = "android")]
+// _PATH_DEFPATH in https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/paths.h
+const ANDROID_DEFPATH: &str = "/product/bin:/apex/com.android.runtime/bin:/apex/com.android.art/bin:/apex/com.android.virt/bin:/system_ext/bin:/system/bin:/system/xbin:/odm/bin:/vendor/bin:/vendor/xbin";
+
 impl crate::sys::fs::PathExt for Path {
     fn readable(&self) -> bool {
         nix::unistd::access(self, nix::unistd::AccessFlags::R_OK).is_ok()
@@ -70,15 +74,22 @@ fn try_get_file_mode(path: &Path) -> Option<u32> {
 }
 
 pub(crate) fn get_default_executable_search_paths() -> Vec<PathBuf> {
-    // standard hard-coded defaults for executable search path
-    vec![
-        "/usr/local/sbin".into(),
-        "/usr/local/bin".into(),
-        "/usr/sbin".into(),
-        "/usr/bin".into(),
-        "/sbin".into(),
-        "/bin".into(),
-    ]
+    #[cfg(target_os = "android")]
+    {
+        std::env::split_paths(ANDROID_DEFPATH).collect()
+    }
+    #[cfg(not(target_os = "android"))]
+    {
+        // standard hard-coded defaults for executable search path
+        vec![
+            "/usr/local/sbin".into(),
+            "/usr/local/bin".into(),
+            "/usr/sbin".into(),
+            "/usr/bin".into(),
+            "/sbin".into(),
+            "/bin".into(),
+        ]
+    }
 }
 
 /// Retrieves the platform-specific set of paths that should contain standard system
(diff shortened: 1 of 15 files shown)


brush-core/src/sys/unix/fs.rs
```diff
@@ -105,6 +105,14 @@ pub fn get_default_standard_utils_paths() -> Vec<PathBuf> {
     ]
 }
 
+#[allow(clippy::unnecessary_wraps)]
+#[cfg(target_os = "android")]
+fn confstr_cs_path() -> Result<Option<PathBuf>, std::io::Error> {
+    Ok(Some(PathBuf::from(
+        "/product/bin:/apex/com.android.runtime/bin:/apex/com.android.art/bin:/system_ext/bin:/system/bin:/system/xbin:/odm/bin:/vendor/bin:/vendor/xbin",
```
reubeno: Can you please add a comment explaining how these paths were derived?

Also note I see https://android.googlesource.com/platform/bionic/+/be5088e4adab806eac07e3a3f82b401b0c1f5553 added another path: `/apex/com.android.virt/bin` about a year ago. If we're trying to sync with that default path list, let's include this one as well.

reubeno (commented):
Thanks for the contribution, @YumeYuka! The PR title mentions 32-bit support. Is that still accurate? I don't recall seeing any 32-bit specific changes included.

Also, would you be able to share some details on what this allowed you to do on Android -- and any instructions for how to build and run on Android to validate that it works?

One additional consideration is that we might break this in the future without some GitHub checks ensuring that we can at least build for an android target. Do you know if it's as simple as adding another case to our cargo build matrix in our PR checks, or if there's more complexity?

brush-core/src/sys/unix/fs.rs
ShIroRRen: > Can you please add a comment explaining how these paths were derived?
> 
> Also note I see https://android.googlesource.com/platform/bionic/+/be5088e4adab806eac07e3a3f82b401b0c1f5553 added another path: `/apex/com.android.virt/bin` about a year ago. If we're trying to sync with that default path list, let's include this one as well.

I directly used the default PATH of my Android 14 device and didn't pay much attention to the original definitions in the system, so I missed it (my device doesn't have com.android.virt yet)

ShIroRRen:
> Thanks for the contribution, @YumeYuka! The PR title mentions 32-bit support. Is that still accurate? I don't recall seeing any 32-bit specific changes included.
> 
> Also, would you be able to share some details on what this allowed you to do on Android -- and any instructions for how to build and run on Android to validate that it works?
> 
> One additional consideration is that we might break this in the future without some GitHub checks ensuring that we can at least build for an android target. Do you know if it's as simple as adding another case to our cargo build matrix in our PR checks, or if there's more complexity?

Currently, there is a misuse of mode_t (it is u16 on 32-bit and u32 on 64-bit, but in nix it is always u32, because nix uses c_uint in its definition). I only fixed this part so that it can build successfully on 32-bit (only tested building 32-bit Android). I am not sure if there are other issues. 

For Android builds, I use cargo-ndk, which can successfully build Android arm64, arm, x64, and x86. I only added Android checks in places where it only checked for Linux, without doing very detailed testing. If you have functions that might cause problems on Android, I can try them out.

brush-core/src/sys/unix/fs.rs
```diff
@@ -1,186 +1,195 @@
...
-/// Opens a null file that will discard all I/O.
-pub fn open_null_file() -> Result<std::fs::File, error::Error> {
-    let f = std::fs::File::options()
-        .read(true)
-        .write(true)
-        .open("/dev/null")?;
-
-    Ok(f)
-}
+//! Filesystem utilities.
```
reubeno: It looks like this file ended up with CRLF sequences and thus the large diff. Would you please convert it back (e.g., via `dos2unix`)?

brush-core/src/sys/unix/fs.rs
```diff
@@ -1,186 +1,195 @@
...
+    path.metadata().map(|metadata| metadata.file_type()).ok()
+}
+
+fn try_get_file_mode(path: &Path) -> Option<u32> {
+    path.metadata().map(|metadata| metadata.mode()).ok()
+}
+
+pub(crate) fn get_default_executable_search_paths() -> Vec<PathBuf> {
+    // standard hard-coded defaults for executable search path
+    vec![
```
reubeno: With the larger diff visible, I noticed this; will we need to update this as well for Android? The two hard-coded sets of search paths are intentionally separate and used in different code paths.

ShIroRRen: Indeed, I didn't notice it before, and now I have also corrected these two

reubeno: Thanks for the changes. We may refactor this a bit in the future as we find other platforms have their own hard-coded paths, but this looks good for now.

brush-core/src/sys/unix/fs.rs
ShIroRRen: I changed CRLF to LF, but I found that other files seem to be CRLF? (In the IDE they are all CRLF, only `fs.rs` has this issue because of CRLF) I'm not sure about the reason. If you need it to be fixed to LF, I think you can add a `.gitattributes` and also set it in `rustfmt.toml`

ShIroRRen:
@reubeno I have a small suggestion. No release has been published this year yet, but I think there have been quite a few changes. At the very least, the patch version could be updated to allow dependency from `crate.io` (Rust has an unstable feature `bindeps`, and I am using it to depend on brush to build it. It's a very convenient feature)

Issue #1073, opened 2026-03-28
system() libc call does not work when brush is /bin/sh


ArtikusHG:
### brush version

0.3.0

### bash version (for comparison)

5.3.0

### Platform / OS

Linux

### Steps to reproduce

Install brush as /bin/sh and try running a `system()` call in a C program

### Actual behavior (brush)

For context, I was trying to run brush as my system shell on Fedora Atomic, but I noticed that after replacing /bin/sh with brush, I got no sound - ALSA wasn't seeing any sound cards. Digging deeper, I found out that udev was failing to load a kernel module - and the default error message from brush for `-c` being empty appeared in the logs. I dug even deeper, and found that `modprobe` uses a `system()` call to run commands, and compiled a simple test program to check if what I found was true:

```c
#include <stdlib.h>

int main(void) {
    system("ls");
    return 0;
}
```

Output on brush:

```sh
error: a value is required for '-c <COMMAND>' but none was supplied

For more information, try '--help'.
```

Output on bash: contents of directory (runs ls)

### Expected behavior

system() should work, just as in every other shell

### Additional context

FWIW, my system actually boots and works fine, and ALSA sound seems to be the only issue so far! I'm guessing this should be an easy fix and hopefully I'll be able to use a rust-based /bin/sh soon!

reubeno:
Thanks for reporting this, @ArtikusHG! Glad to hear things are otherwise looking positive, but this is definitely a compatibility bug.

From my initial investigation, `system()` in `glibc` invokes `sh` with something like `sh -c -- <ARGS_HERE>`. Sure enough, that causes a command-line parsing error for `brush`, even with latest sources:

<img width="779" height="164" alt="Image" src="https://github.com/user-attachments/assets/8b950df4-e4cb-4b44-9c30-d3e3c9b2bc60" />

PR #1076, opened 2026-03-28
fix(cmdline): handle `--` as option terminator for `-c` flag

reubeno:
In bash, `-c` treats `--` as an option terminator: the command string comes from the first argument after `--`, and `--` itself is consumed. This is unique to `-c`. Previously, `brush -c -- 'echo hi'` failed with a parse error.

--
_Aims to address #1073 but will need more testing._

### DIFF
--- a/brush-shell/src/entry.rs
+++ b/brush-shell/src/entry.rs
@@ -10,6 +10,7 @@ use crate::productinfo;
 use brush_builtins::ShellBuilderExt as _;
 #[cfg(feature = "experimental-builtins")]
 use brush_experimental_builtins::ShellBuilderExt as _;
+use clap::CommandFactory;
 use std::sync::LazyLock;
 use std::{path::Path, sync::Arc};
 use tokio::sync::Mutex;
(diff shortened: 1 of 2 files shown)


ArtikusHG:
Wow, thanks for the quick bugfix! Trying to build an image with the produced artifact right now to see if the issue is fixed.

reubeno:
> Wow, thanks for the quick bugfix! Trying to build an image with the produced artifact right now to see if the issue is fixed.

To be fair, I haven't actually tried the scenario you ran earlier -- this PR fixes a legitimate bug, and I *believe* it's at least partly responsible for the behavior you saw :smile:. Let us know what you find!

ArtikusHG:
Happy to report this fixes the issue! I'm writing this from a booted deployment of my Fedora Atomic image with brush as /bin/sh, while listening to music from my laptop's speakers - modprobe no longer fails, so that means this fixed the system() issue! There's some issues with my custom $PATH from .bashrc not being applied properly, but I'm still figuring out the cause and I'll open a separate issue for that.

It's super impressive how good the compatibility with bash is, considering a lot of startup scripts have a hard dependency on a shell, some even with specific bashisms - I'll keep using this as my system shell and try helping solve as many issues as I can!

reubeno:
> Happy to report this fixes the issue! I'm writing this from a booted deployment of my Fedora Atomic image with brush as /bin/sh, while listening to music from my laptop's speakers - modprobe no longer fails, so that means this fixed the system() issue! There's some issues with my custom $PATH from .bashrc not being applied properly, but I'm still figuring out the cause and I'll open a separate issue for that.
> 
> It's super impressive how good the compatibility with bash is, considering a lot of startup scripts have a hard dependency on a shell, some even with specific bashisms - I'll keep using this as my system shell and try helping solve as many issues as I can!

Excellent :smile:. Many thanks for the quick testing and for reporting the issue, @ArtikusHG 

(And please do keep the reports coming -- positive or negative :smile:.)

PR #1080, opened 2026-03-29
fix(-c): call EXIT traps on exit; add tests

### LINKED ISSUE #1079: async commands (&) run via `-c` aren't spawned before brush exits
### brush version

sh -c -- fix debug build

5.3.0

### Platform / OS

Linux

### Steps to reproduce

brush -c "some-command &" - does not run at all

### Actual behavior (brush)

When trying to run a command with -c that has an & symbol at the end, the command doesn't run. If there's another command after it, it does run, but the commands run in a seemingly random order, sometimes some of them don't run at all.

```sh
[~] sh -c "echo 1"
1
[~] sh -c "echo 1 &" # no output :(
[~] sh -c "echo 1 & echo 2 & echo 3 & echo 4"
1
2
3
4
[~] sh -c "echo 1 & echo 2 & echo 3 & echo 4"
4
[~] sh -c "echo 1 & echo 2 & echo 3 & echo 4"
4
1
2
[~] sh -c "echo 1 & echo 2 & echo 3 & echo 4"
1
4
[~] sh -c "echo 1 & echo 2 & echo 3 & echo 4"
4
2
1
[~] sh -c "echo 1 & echo 2 & echo 3 & echo 4"
4
1
2
```

Note that in an interactive shell both running a single command and multiple commands works fine.

From my understanding, the fact that they're in random order is expected, but the fact that sometimes only *some* of them run isn't great, and the fact that if I try to run a single command in the background, it simply ignores that request, prevents me from being able to use a lockscreen using swayidle.

### Expected behavior

commands with & run correctly, just like in interactive shell

### Additional context

Sorry for the load of bug reports :/ I'm probably the only one crazy enough to try to use this as their system shell lol

reubeno:
This resolves an inconsistency between the `-c` code path and paths for interactive execution as well as script execution.

Also adds tests, some of them skipped by default until we resolve #1079.

### DIFF
--- a/brush-core/src/shell/execution.rs
+++ b/brush-core/src/shell/execution.rs
@@ -194,7 +194,7 @@ impl<SE: crate::extensions::ShellExtensions> crate::Shell<SE> {
 
         self.end_command_string_mode()?;
 
-        // Give the shell a change to run on-exit tasks, but ignore the result.
+        // Give the shell a chance to run on-exit tasks, but ignore the result.
         let _ = self.on_exit().await;
 
         Ok(result)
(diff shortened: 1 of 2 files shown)


ArtikusHG:
Duplicating this here just in case - this PR doesn't solve the issue.
image

PR #1104, opened 2026-04-19
fix(windows): fix /dev/null handling

theoparis:
This makes brush closer to behaving correctly with starship on Windows.

Before:
```
PS D:\wfs> brush
error: command not found:
error: failed to redirect to D:/dev/null: The system cannot find the path specified. (os error 3)

D:wfs
❯ ls
brush  portage	redbox
error: failed to redirect to D:/dev/null: The system cannot find the path specified. (os error 3)
```

After:
```
PS D:\wfs> .\brush\target\debug\brush
error: command not found:

D:wfs
❯ ls
brush  portage	redbox
```

### DIFF
--- a/brush-core/src/sys/windows/fs.rs
+++ b/brush-core/src/sys/windows/fs.rs
@@ -148,7 +148,7 @@ pub fn open_null_file() -> Result<std::fs::File, error::Error> {
 
 /// Gives the platform an opportunity to handle a special file path (e.g. `/dev/null`).
 pub fn try_open_special_file(path: &Path) -> Option<Result<std::fs::File, std::io::Error>> {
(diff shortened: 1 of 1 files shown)


reubeno:
Thanks for the contribution, @theoparis! We definitely appreciate you working to fix one of the issues you'd identified.

The PR check failures seem unrelated to your changes -- I'll sort out what's going on. (Looks like new clippy warnings in existing code -- and also some flaky tests.)

reubeno (approved):
Tests are passing now.

I suspect we may need to refine this later, but for now I'm going to merge this since it's definitely a win for the Windows builds (and starship usage).

PR #1109, opened 2026-04-26
fix: bang formatted without space leads to invalid command

cilki:
When formatting a negated command like:

```
! $0
```

`Pipeline::fmt` results in:

```
!$0
```

which is incorrect.


reubeno:
Thanks for the contribution! The change seems reasonable.

I’ve queued the PR checks to run so we can see their results.

reubeno (approved):
The failing PR checks are unrelated; we'll resolve those separately.

Change looks good.

PR #1116, opened 2026-05-01
feat: use bstr::BString for shell values to fix UTF-8 lossy conversion

### LINKED ISSUE #1115: lossy utf8 conversion usage
### Description

Inspired by [this](https://corrode.dev/blog/bugs-rust-wont-catch). We are using utf8 lossy conversion in a bunch of places. Let's see if they are all fine.

lu-zero:
## Summary

- Convert `ShellValue`, `ShellValueLiteral`, and `ArrayLiteral` from `String` to `bstr::BString` so arbitrary bytes are preserved throughout the shell variable system
- Fix silent data corruption in `echo -e`, `mapfile`, ANSI-C quoting (`$'...'`), `${var@E}`, PWD/OLDPWD, and `From<OsString>` for `ShellValue`

Closes #1115

## Details

Bash treats all data (filenames, variable values, command output) as raw byte sequences. Rust's `String` requires valid UTF-8, so every `from_utf8_lossy` or `to_string_lossy` call was a potential silent data corruption point.

The core change replaces `String` with `bstr::BString` (backed by `Vec<u8>`) in `ShellValue::String`, `AssociativeArray`, and `IndexedArray`, plus all related types. Key fixes:

- **`echo -e '\x80'`**: now writes raw bytes to stdout instead of lossy-converting
- **`mapfile`**: stores raw bytes as `BString`
- **ANSI-C quoting** (`$'\x80'`): preserves non-UTF-8 bytes
- **`${var@E}`**: same as above
- **PWD/OLDPWD**: uses `OsStrExt::as_bytes()` on Unix to preserve raw path bytes
- **`From<OsString>`**: uses `into_vec()` on Unix instead of `to_string_lossy()`

24 files changed, all tests passing (49 unit + 224 parser + 25 builtins + 1696 compat).

Co-Authored-by: opencode <opencode@anomaly.co>
Assisted-by: glm-5.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1609,9 +1609,9 @@ dependencies = [
 
 [[package]]
 name = "fancy-regex"
-version = "0.18.0"
+version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e1e1dacd0d2082dfcf1351c4bdd566bbe89a2b263235a2b50058f1e130a47277"
+checksum = "476de73bddf2ef8490aa4ee8f1cf40b430bf1d56c48c22080e5186952cd580e6"
 dependencies = [
  "bit-set",
  "regex-automata",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -434,6 +434,7 @@ dependencies = [
  "async-trait",
  "bon",
  "brush-parser",
+ "bstr",
  "cached",
  "cfg-if",
  "check_elevation",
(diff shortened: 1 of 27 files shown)


brush-builtins/src/echo.rs
```diff
@@ -48,32 +48,32 @@ impl builtins::Command for EchoCommand {
         let mut trailing_newline = !self.no_trailing_newline;
         let mut s;
         if self.interpret_backslash_escapes {
-            s = String::new();
+            s = Vec::new();
```
Elsie19:
```suggestion
            s = Vec::with_capacity(self.args.len());
```

If you do this you can avoid a couple reallocations.

Elsie19 (changes requested):

PR #1131, opened 2026-05-04
fix: source $BASH_ENV / $ENV for non-interactive non-login shells (with privilege-mode hardening)

### LINKED ISSUE #1130: bug: BASH_ENV / $ENV not sourced for non-interactive non-login shells (returns "not yet implemented")
## Summary

When `BASH_ENV` (or `ENV` in `--sh` mode) is set in the environment, brush returns an "unimplemented" error instead of sourcing the named file, breaking non-interactive `-c` invocations.

## Reproduction

```bash
echo 'echo loaded' > /tmp/empty.sh
BASH_ENV=/tmp/empty.sh brush -c 'echo hello'
```

**Expected:**
```
loaded
hello
```

**Actual:**
```
error: not yet implemented: load config from $ENV/BASH_ENV for non-interactive, non-login shell
```

Workaround: pass `--noenv`.

## Why this matters

It's common for users to have `BASH_ENV` exported from their login profile (e.g., to point at a shared init script for non-interactive bash). Once `BASH_ENV` is set, *every* non-interactive `brush -c` invocation fails. That makes brush hard to use as a drop-in `bash -c` replacement from other tools (CI runners, editor "run shell command" features, parent processes that exec a shell to run a one-liner) without surgery on the caller's environment.

## Source

The bailout is at `brush-core/src/shell/initscripts.rs` and is already TODO'd:

```rust
// TODO(well-known-vars): look at $ENV/BASH_ENV; source its expansion if that
// file exists
return error::unimp(
    "load config from $ENV/BASH_ENV for non-interactive, non-login shell",
);
```

## Bash semantics worth mirroring

Per `bash(1)` (INVOCATION section):

- For non-interactive non-login shells: if `BASH_ENV` is set in the environment, expand the value and source the resulting file (if it exists and is readable)
- In POSIX/`sh` mode: same behavior using `ENV` instead of `BASH_ENV`
- The value undergoes parameter expansion, command substitution, and arithmetic expansion before being used as a filename
- Bash silently continues if the file does not exist (no error)
- Behavior is suppressed when the effective UID/GID differ from the real ones (bash skips env-file reads in privileged contexts)

## Environment

- brush 0.4.0 (installed via Homebrew on macOS)
- Darwin 25.4.0, arm64

Happy to take a stab at this — would follow up with a draft PR using the existing `source_if_exists` helper and the basic expansion path used elsewhere for shell variables.

melderan:
## Summary

Implements the existing `TODO(well-known-vars)` in `brush-core/src/shell/initscripts.rs`: when `BASH_ENV` (or `ENV` in `--sh` mode) is set on a non-interactive non-login shell, expand the value and source the resulting file if it exists, instead of returning `unimp`.

Also adds bash-equivalent privilege-mode hardening to `load_config`: when running with privileges (effective UID/GID different from real), refuse to source any init file whose location is environment-controlled. This closes both the new BASH_ENV vector this PR would otherwise introduce, and the pre-existing equivalent vector via `$HOME`-derived init files (`~/.bash_profile`, `~/.bashrc`, etc.) — `$HOME` is itself env-controlled, so without the gate an unprivileged caller of a setuid brush could direct the privileged shell to source attacker-owned files.

Closes #1130.

## Two commits, one story

1. **`fix: source $BASH_ENV / $ENV for non-interactive non-login shells`** — implements the missing init-file load. Adds three compat tests under `well_known_vars.yaml`. Verified parity with bash for parameter expansion, `$PWD` substitution, tilde, and missing-file handling.

2. **`fix: skip env-controlled init files when running with privileges`** — adds `sys::users::is_privileged()` (Unix: `getuid() != geteuid() || getgid() != getegid()`; Windows / WASI: const false, no setuid vector on those platforms), and uses it to gate `BASH_ENV` / `ENV`, the login-shell home-derived branch (`~/.bash_profile` / `~/.bash_login` / `~/.profile`), and the interactive non-`--rcfile` branch (`~/.bashrc` / `~/.brushrc`). System-wide files (`/etc/profile`, `/etc/bash.bashrc`) and CLI-controlled `--rcfile` paths remain honored. Mirrors bash's behavior under `-p`.

Happy to squash if the maintainer prefers a single commit.

## Behavior — matches `bash(1)` INVOCATION

- Reads `$BASH_ENV` (or `$ENV` in `--sh` mode) from the shell's environment
- Applies basic shell expansion (parameter / command / arithmetic / tilde) to the value via `expansion::basic_expand_word` — verified that `BASH_ENV='~/script.sh'` and `BASH_ENV='$PWD/script.sh'` both behave identically to bash
- Sources the resulting file via the existing `source_if_exists` helper — silently skips if it doesn't exist (matches bash, which doesn't error on missing `BASH_ENV` files)
- Skipped entirely under privilege (along with all other env-controlled init paths)

## Trade-off worth flagging

Removing the `unimp` early-return means that if a user's `BASH_ENV` file contains a bash construct brush hasn't implemented yet, they'll see a confusing mid-source error rather than the clear "not yet implemented" wall. Net better — brush becomes usable as a `bash -c` drop-in for users with `BASH_ENV` set in their login profile — but I wanted to call it out so it isn't a surprise.

## Notes on the privilege check's scope

- **Linux file capabilities** (`cap_setuid+ep` etc.) granted without the setuid bit are intentionally not detected — bash itself doesn't detect them either, so this matches established shell-hardening behavior. Documented in the doc-comment on `is_privileged()`.
- **Supplementary groups** likewise not considered, matching bash.
- **Windows / WASI** privileges are not propagated via the env-as-attack-surface path that motivates this gate, so the platform stubs return const `false` (with rationale documented inline).

## Tests

Added compat tests under `brush-shell/tests/cases/compat/well_known_vars.yaml`:
- `BASH_ENV` pointing at an existing file → file is sourced (parity with bash)
- `BASH_ENV` pointing at a non-existent path → no error, shell continues (parity with bash)
- `BASH_ENV` value containing `$PWD` → expansion applied before source attempt (parity with bash)

The privileged-mode gate isn't easily testable in CI without root + setuid setup. I verified non-privileged behavior is unchanged via the existing test suite, and traced the gating manually with the privilege check forced on/off in a debug build. Happy to add a privileged-mode integration test if there's an existing pattern for it in the project I missed.

## Local verification

- `cargo xtask ci pre-commit` — fmt, build, lint, deps, schemas, unit tests all green; integration tests pass except for one pre-existing `brush-interactive-tests::run_in_bg_then_fg` flake unrelated to this change (also fails on `main` on the same machine; the dev environment's interactive `.bashrc` references `mise`/`direnv`/`atuin`/`zoxide` hook functions that aren't present in CI)
- The original repro `BASH_ENV=/tmp/empty.sh brush -c 'echo hi'` now prints the sourced output and the command output, with parity against `bash --norc --noprofile -c '...'`
- Verified `HOME=/tmp/fake brush --login --noprofile -c '...'` correctly sources `/tmp/fake/.bash_profile` in the non-privileged case

Filed as a draft for early input — happy to address feedback before flipping to ready-for-review.

---

### DIFF
--- a/brush-core/src/shell/initscripts.rs
+++ b/brush-core/src/shell/initscripts.rs
@@ -1,6 +1,6 @@
 //! Init script support for shells.
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::{Shell, error, extensions, interp};
 
@@ -125,14 +125,21 @@ impl<SE: extensions::ShellExtensions> Shell<SE> {
                     "BASH_ENV"
                 };
 
-                if self.env.is_set(env_var_name) {
-                    //
-                    // TODO(well-known-vars): look at $ENV/BASH_ENV; source its expansion if that
-                    // file exists
-                    //
-                    return error::unimp(
-                        "load config from $ENV/BASH_ENV for non-interactive, non-login shell",
-                    );
+                // Per bash(1) INVOCATION: when BASH_ENV is set on a non-interactive
+                // non-login shell (or ENV in POSIX/sh mode), the value is subjected to
+                // parameter expansion, command substitution, and arithmetic expansion,
+                // and the resulting filename is sourced if it exists. If the file does
+                // not exist or is not readable, bash silently continues.
+                //
+                // FIXME: bash also suppresses this load when the shell is running with
+                // privileges (effective UID/GID different from real); that protection
+                // is not yet implemented here.
+                if let Some(raw_value) = self.env_str(env_var_name).map(|v| v.into_owned()) {
+                    let expanded =
+                        crate::expansion::basic_expand_word(self, &params, raw_value).await?;
+                    if !expanded.is_empty() {
+                        self.source_if_exists(Path::new(&expanded), &params).await?;
+                    }
                 }
             }
         }
--- a/brush-core/src/shell/initscripts.rs
+++ b/brush-core/src/shell/initscripts.rs
@@ -55,6 +55,20 @@ impl<SE: extensions::ShellExtensions> Shell<SE> {
         let mut params = self.default_exec_params();
         params.process_group_policy = interp::ProcessGroupPolicy::SameProcessGroup;
 
+        // When running with privileges (setuid/setgid where effective UID/GID
+        // differs from real), refuse to source any init file whose location is
+        // controlled by the (untrusted) environment. That includes BASH_ENV and
+        // ENV (env-controlled directly) as well as anything under `$HOME` —
+        // because `$HOME` is itself environment-controlled, an unprivileged
+        // caller can otherwise direct the privileged shell to source
+        // attacker-owned `.bash_profile` / `.bashrc` / etc.
+        //
+        // System-wide init files (`/etc/profile`, `/etc/bash.bashrc`) and
+        // explicit `--rcfile` paths remain honored — those are root-owned or
+        // CLI-controlled, not attacker-controlled. This mirrors bash's
+        // behavior under `-p`.
+        let privileged = crate::sys::users::is_privileged();
+
         if self.options.login_shell {
             // --noprofile means skip this.
             if matches!(profile_behavior, ProfileLoadBehavior::Skip) {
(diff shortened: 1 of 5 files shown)


reubeno:
@melderan Thanks for the PR!

I've unblocked the PR checks to run.

I'll want to review this a bit more carefully when you're ready to take it out of draft but am generally supportive of the changes you're introducing here. 

We may also want to think a bit more about how to (repeatably) test some of the hardening logic--I know that may be less turnkey with our current test infrastructure.

Issue #1135, opened 2026-05-04
bug: `history -a/-w/-n/-r` silently no-op on unset HISTFILE; bash emits diagnostic


melderan:
## Summary

When `HISTFILE` is unset and no explicit file argument is provided, brush's `history -a`, `history -w`, `history -n`, and `history -r` silently no-op with exit 0. Bash emits a diagnostic to stderr in the same situation (and also returns 0). Minor bash-parity gap; the missing diagnostic is the only difference.

## Reproduction

```bash
$ unset HISTFILE
$ bash --norc --noprofile -c 'history -n'
bash: line 1: history: HISTFILE: parameter null or not set

$ brush -c 'history -n'      # silent
$ echo $?
0
```

Same observable for `-a`, `-w`, `-r` (anything that calls `get_effective_history_file_path` with no fallback).

## Suggested fix

In `brush-builtins/src/history.rs`, when `get_effective_history_file_path` returns `None` for the file-touching flags, emit something like:

```
brush: history: HISTFILE: parameter null or not set
```

to stderr before returning success. Match bash's exact wording for compat-script friendliness.

## Why minor

The functional behavior is correct (don't crash, don't write to a wrong place). The diagnostic is informational only. Some `__history_sync` style PROMPT_COMMAND functions look at the exit code, not stderr, so this doesn't break common workflows. Filed for completeness; happy to follow up with a PR if welcome.

## Environment

- brush 0.4.0 + the `history -n`/`history -r` PR-in-flight (which surfaces this gap; same gap exists pre-PR for `-a`/`-w`).

reubeno:
Thanks for reporting this, @melderan.

We don't guarantee character-for-character identical diagnostic messages, but I agree that it would be better not to silently ignore this case.

Issue #100, opened 2024-06-28
Implement history expansion (word designators, event designators)


reubeno:
In `bash`, there's a set of non-`$` expansions (e.g., `!!`) that will expand based on the current shell history. None of this is implemented in `brush` as of yet.

This will need to go into word expansion. Some work may need to be done to better integrate shell history into the core brush crate; right now much of it is handled by `brush-interactive`.

reubeno:
Now that history management has matured in `brush`, this should be easier to start making happen!

PR #1136, opened 2026-05-04
fix: implement history -n and history -r

### LINKED ISSUE #1134: bug: `history -n` and `history -r` not implemented; emits stderr error per prompt cycle
## Summary

`history -n` and `history -r` return `error: not yet implemented` instead of doing what bash does. Both flags are parsed and dispatched in `brush-builtins/src/history.rs`, but the implementation bodies are `error::unimp(...)` stubs.

## Reproduction

```bash
$ brush -i <<< 'history -n; exit'
error: history: not yet implemented: history -n is not yet implemented

$ brush -i <<< 'history -r; exit'
error: history: not yet implemented: history -r is not yet implemented
```

## Why this matters

A common pattern in shell config is a `PROMPT_COMMAND` function that calls `history -a` (works in brush) and `history -n` (errors in brush) to keep multi-terminal history in sync. With `history -n` unimp'd, every prompt cycle in such configs prints a stderr error.

A representative example:

```bash
__history_sync() {
    local last_cmd last_in_file
    last_cmd=$(HISTTIMEFORMAT='' history 1 | sed 's/^[ ]*[0-9]*[ ]*//')
    last_in_file=$(grep -v '^#[0-9]*$' ~/.bash_history | tail -1)
    if [[ "$last_cmd" != "$last_in_file" ]]; then
        history -a
    fi
    history -n
}
export PROMPT_COMMAND="__history_sync"
```

This sources cleanly in brush today, but every interactive prompt emits an `error: history -n is not yet implemented` line.

## Bash semantics to match

- `history -n [filename]`: Read all history lines NOT ALREADY READ from the history file into the in-memory list. Track per-session "where we last read to."
- `history -r [filename]`: APPEND the file contents to the in-memory history (despite the man page wording "use them as the current history" — bash actually appends; verified empirically).
- Both flags accept an optional explicit file argument; absent that, fall back to `$HISTFILE`.

Also worth doing in the same touchpoint: replace the `error::unimp(...)` stub for `-p` with `error::unimp_with_issue(..., 100)` since #100 is the correct tracker for that work.

## Environment

- brush 0.4.0 (Homebrew on macOS, also reproduced from local cargo build)
- Darwin 25.4.0, arm64

Happy to follow up with a PR.

melderan:
## Summary

Implements `history -n` and `history -r`, which previously returned `error: not yet implemented`. Also retags the `history -p` stub via `unimp_with_issue(..., 100)` so anyone hitting it lands on #100 (history expansion) directly.

The motivating case: a `__history_sync` function in `PROMPT_COMMAND` (a common pattern for keeping multi-terminal history coherent) calls `history -a` (works) and `history -n` (errors today). With this fix, those configurations stop spamming stderr on every prompt cycle.

Closes #1134.

## Behavior — matches bash

- **`history -n [filename]`**: reads all entries from the file that haven't already been read into this session, appending them to the in-memory history. Tracks a per-session "last read offset" via a new `last_file_read_count` field on `History`.
- **`history -r [filename]`**: appends the full file contents to the in-memory history. Note: bash's man page says "use them as the current history," but the actual bash implementation **appends rather than replaces** — verified empirically with `bash --norc --noprofile -c 'history -s in-mem; history -r src.txt; history'`. The tests follow the empirical behavior.
- Both flags accept an optional explicit file argument; absent that, fall back to `$HISTFILE`.
- **Truncation recovery**: if the history file shrinks externally between two `-n` calls (log rotation, another session doing `history -w`), `last_file_read_count` is reset to the new (smaller) count so subsequent appends remain visible.
- **1 GiB file-size guard** is applied to `-n` and `-r`, matching the existing `load_history` behavior. While we were there, the constant was lifted to a single `pub(crate) const` so the threshold stays in lockstep across the three import paths.

## Design notes worth flagging

- **`last_file_read_count` is `serde(skip)`**: it's session-local state. Resetting to 0 across serialization boundaries is safe — a subsequent `-n` just re-reads everything from the file (the conservative outcome).
- **`clear()` resets `last_file_read_count` to 0** (so `history -c; history -n` reads everything from the file, matching bash). `next_id` is intentionally **not** reset — comment in code explains why (reedline ID stability across clear/reload cycles).
- **Items loaded from the file are constructed via direct `Item { dirty: false, ... }` literal**, not `Item::new()` (which sets `dirty: true`). That ensures `history -a` doesn't double-write entries that came from the file.
- **Reedline-clone investigation**: `brush-interactive/src/reedline/history.rs` accesses `History` exclusively through `MutexGuard` borrows; no `History::clone()` calls. The `Clone` derive on the struct exists but is not exercised by the reedline integration, so `last_file_read_count` doesn't get reset mid-session.

## Tests

- 8 new compat test cases under `brush-shell/tests/cases/compat/builtins/history.yaml` covering: load file, append-vs-replace, explicit file argument, incremental `-n` reads, idempotence, post-`history -c` behavior, and truncation recovery. All use `args: ["-o", "history"]` for deterministic non-interactive sequencing.
- 2 new unit tests (`test_parse_dash_n`, `test_parse_dash_r`) mirroring the existing `test_parse_dash_a`.

The new tests use `skip_oracle: true` + `expected_stdout` rather than bash-oracle comparison. Reason: brush loads `$HISTFILE` at startup in non-interactive mode while bash doesn't, which causes oracle divergence on incidental output unrelated to the feature being tested. This is a pre-existing brush↔bash divergence, not introduced by this PR. Happy to switch to oracle comparison if the maintainer prefers; this seemed like the lower-risk choice.

## Known related gap (not in this PR)

When `HISTFILE` is unset and no explicit file argument is given, `-a`/`-w`/`-n`/`-r` silently no-op (exit 0). Bash emits a diagnostic to stderr in the same situation (also exit 0). Filed as #1135. Doesn't affect the motivating `__history_sync` case (where `HISTFILE` is set), but worth tracking.

## Local verification

- `unset BASH_ENV && cargo xtask ci pre-commit` — fmt, build, lint, deps, schemas, unit tests all green; integration tests pass except for the pre-existing `brush-interactive-tests::run_in_bg_then_fg` flake unrelated to this change (also fails on `main` on this dev machine; the local `.bashrc` references `mise`/`direnv`/`atuin`/`zoxide` hook functions that aren't present in CI).
- The original repro `brush -i <<< 'history -n; exit'` now exits clean instead of printing an error.

## Two commits, one story

1. `fix: implement history -n and history -r` — the actual implementation
2. `refactor: lift MAX_FILE_SIZE_FOR_HISTORY_IMPORT to module scope` — small follow-on dedupe of the file-size constant across the three import paths

Happy to squash if the maintainer prefers a single commit.

Filed as a draft for early input.

---

### DIFF
--- a/brush-core/src/history.rs
+++ b/brush-core/src/history.rs
@@ -19,6 +19,12 @@ pub struct History {
     items: rpds::VectorSync<ItemId>,
     id_map: rpds::HashTrieMapSync<ItemId, Item>,
     next_id: ItemId,
+    /// Tracks how many items were present in the history file the last time we read it.
+    /// Used by `read_new_from_file` (`history -n`) to skip already-loaded entries.
+    /// Serde-skipped: not persisted across serialization boundaries; resets to 0 on
+    /// deserialization (safe — subsequent `-n` reads everything new from the file).
+    #[cfg_attr(feature = "serde", serde(skip))]
+    last_file_read_count: usize,
 }
 
 impl History {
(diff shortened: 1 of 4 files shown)


brush-shell/tests/cases/compat/builtins/history.yaml
```diff
@@ -191,3 +191,198 @@ cases:
     stdin: |
       bash -c 'printf "\xff\xfe\xfd" >history-file.txt'
       HISTFILE=history-file.txt $0 -o history -c 'echo hi'
+
+  # Tests for history -r and history -n.
+  #
+  # These tests use skip_oracle + expected_stdout rather than comparing against bash's oracle,
```
reubeno, line 197: I'm not sure I follow why the oracle testing approach can't be used? The other tests in this file are able to compare history loading/manipulation against bash's behavior.

Are you suggesting that there's something unique about the options targeted by this PR?

brush-core/src/history.rs
```diff
@@ -19,6 +24,12 @@ pub struct History {
     items: rpds::VectorSync<ItemId>,
     id_map: rpds::HashTrieMapSync<ItemId, Item>,
     next_id: ItemId,
+    /// Tracks how many items were present in the history file the last time we read it.
+    /// Used by `read_new_from_file` (`history -n`) to skip already-loaded entries.
+    /// Serde-skipped: not persisted across serialization boundaries; resets to 0 on
+    /// deserialization (safe — subsequent `-n` reads everything new from the file).
+    #[cfg_attr(feature = "serde", serde(skip))]
+    last_file_read_count: usize,
```
reubeno, line 32: Is this captured independently of *which* file was last read? Do we have any new test cases that check behavior when subsequent invocations use different history file paths (or the history file has since been changed underneath us)?

brush-core/src/history.rs
```diff
@@ -19,6 +24,12 @@ pub struct History {
     items: rpds::VectorSync<ItemId>,
     id_map: rpds::HashTrieMapSync<ItemId, Item>,
     next_id: ItemId,
+    /// Tracks how many items were present in the history file the last time we read it.
+    /// Used by `read_new_from_file` (`history -n`) to skip already-loaded entries.
+    /// Serde-skipped: not persisted across serialization boundaries; resets to 0 on
```
reubeno, line 29: What's the thinking of not persisting it? Is your concern that the history file has been changed in the intervening time since serialization?

brush-core/src/history.rs
```diff
@@ -32,11 +43,106 @@ impl History {
...
+    /// If the file appears to have been truncated externally (i.e., it now contains fewer entries
+    /// than we last read), the baseline is reset so future appends remain visible.
+    ///
+    /// Note: this re-parses the full file on each call. Acceptable for typical history sizes;
+    /// this is a known cost, not an optimization target.
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - The path to the history file.
+    pub fn read_new_from_file(&mut self, path: impl AsRef<Path>) -> Result<(), error::Error> {
```
reubeno, line 63: This and the function below are quite similar; any way to share logic to ensure they don't drift from each other?

reubeno (commented):
Thanks for tackling these options!

I've left a few initial questions / comments.

PR #1138, opened 2026-05-05
fix: retain terminal control during interactive sessions

### LINKED ISSUE #1137: Brush as macOS login shell hangs with Ghostty
### Description

Setting the login shell to brush with `chsh`, rebooting, and then opening Ghostty does not show the prompt with no response. The built-in Terminal app works as expected.

Confirmed with Brush 0.4.0 (installed via Homebrew) and Ghostty 1.3.1 on macOS 26.4.1.

cataggar:
## Summary

- Keep the `TerminalControl` guard alive for the full `InteractiveShell` lifetime so foreground terminal ownership is not immediately restored before interactive input starts.
- Add a pty regression test that simulates login-shell invocation with a dash-prefixed `argv[0]` and verifies the prompt appears with `login_shell` enabled.

Created by GitHub Copilot.

Fixes #1137.

## Validation

- `cargo test --test brush-interactive-tests login_shell_via_argv0_shows_prompt`
- `cargo test --test brush-interactive-tests run_pipeline_interactively`
- `cargo test --test brush-compat-tests -- 'Interactive::Basic interactive test'`
- `cargo test --workspace --lib --no-fail-fast`
- `cargo xtask ci quick` progressed through fmt, build, and clippy; the unit-test step could not run in this local environment because `cargo-nextest` is not installed.

## Manual verification

This should be manually verified on macOS with Ghostty because CI pty tests cannot fully reproduce Ghostty's login-shell/session setup.

1. Build this branch, for example: `cargo build --bin brush`.
2. Ensure the built `brush` path is allowed as a login shell, for example by adding the absolute binary path to `/etc/shells` if needed.
3. Set it as the user's login shell: `chsh -s <absolute-path-to-brush>`.
4. Reboot or log out/in, then open Ghostty and confirm a prompt appears instead of hanging.
5. In Ghostty, run `shopt -q login_shell && echo login` and a simple command like `echo ok` to confirm the shell is interactive and marked as a login shell.
6. Open Apple Terminal with the same login-shell configuration and confirm it still shows a prompt.
7. Restore the previous login shell with `chsh -s <previous-shell>` after testing.

### DIFF
--- a/brush-interactive/src/interactive_shell.rs
+++ b/brush-interactive/src/interactive_shell.rs
@@ -57,6 +57,8 @@ pub struct InteractiveShell<'a, IB: InputBackend, SE: brush_core::ShellExtension
     input: &'a mut IB,
     /// Terminal integration utility, if any.
     terminal_integration: Option<crate::term_integration::TerminalIntegration>,
+    /// Terminal-control guard, held for the lifetime of the interactive shell.
+    _terminal_control: Option<brush_core::terminal::TerminalControl>,
     /// Options.
     options: InteractiveOptions,
 }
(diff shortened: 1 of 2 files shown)


omasanori:
I confirmed that this change fixes the problem on Ghostty and does not break the behavior on Terminal by executing the manual verification steps described. Thanks!

reubeno (approved):
Fix looks good! Also great to see a test case added.

PR #1144, opened 2026-05-06
fix(parser)!: track byte offsets instead of char counts in source locations

### LINKED ISSUE #1127: brush_parser: Word::loc.\*.index counts UTF-8 chars instead of bytes (off-by-one per multi-byte char)
## Summary

In `brush_parser` 0.4.0, `brush_parser::ast::Word::loc.start.index` and `.end.index` are reported as **character counts** rather than **byte offsets**. As soon as a script contains any multi-byte UTF-8 character, every subsequent token's reported `index` drifts by `byte_len(ch) - 1` per multi-byte char above it.

`loc.start.line` and `loc.start.column` *are* correct (1-based char counts), as is `Word::value`. So the bug is isolated to the byte-index fields.

## Minimal reproducer

```rust
// Cargo.toml: brush-parser = "=0.4"
use brush_parser::ast::*;

fn main() {
    // Two lines. Line 1 has one 2-byte UTF-8 char (é → 0xC3 0xA9).
    let src = "# café\nls\n";
    //                 ^^^ é is 2 bytes
    // Total bytes: '#', ' ', 'c', 'a', 'f', 0xC3, 0xA9, '\n', 'l', 's', '\n' = 11 bytes
    // The `l` of `ls` is at byte 8.

    let opts = brush_parser::ParserOptions::default();
    let mut p = brush_parser::Parser::new(src.as_bytes(), &opts);
    let prog = p.parse_program().unwrap();

    for cl in &prog.complete_commands {
        for it in &cl.0 {
            for cmd in &it.0.first.seq {
                if let Command::Simple(s) = cmd {
                    if let Some(w) = &s.word_or_name {
                        let loc = w.loc.as_ref().unwrap();
                        println!(
                            \"value={:?}  loc.line={} loc.column={} loc.start.index={} loc.end.index={}  bytes_at_index={:?}\",
                            w.value, loc.start.line, loc.start.column, loc.start.index, loc.end.index,
                            std::str::from_utf8(&src.as_bytes()[loc.start.index..loc.end.index]).unwrap()
                        );
                    }
                }
            }
        }
    }
}
```

### Expected

```
value=\"ls\"  loc.line=2 loc.column=1 loc.start.index=8 loc.end.index=10  bytes_at_index=\"ls\"
```

### Actual (brush-parser 0.4.0)

```
value=\"ls\"  loc.line=2 loc.column=1 loc.start.index=7 loc.end.index=9  bytes_at_index=\"\\nl\"
```

The `value`, `line`, and `column` are correct, but `start.index`/`end.index` are off by 1 — exactly the byte cost of the single `é` above. With 100 multi-byte chars above a token, the index is off by 100 (or 200, for 3-byte chars).

## Impact

Any downstream tool that uses `Word::loc.\\*.index` to slice the original source — for diagnostics, source-preserving rewrites, syntax highlighting, LSP-style ranges, etc. — will read the wrong bytes on any non-ASCII script. We hit this in [rusholve](https://github.com/qweered/rusholve) (a Rust port of [resholve](https://github.com/abathur/resholve)) when running it across nixpkgs: French shell scripts (`pdfmm`) had every command misreported (e.g. `mv` reported as `nc`, the literal `type` builtin reported as `ype`, and ghost commands like `sol` extracted from inside the word `absolu` in a comment).

## Workaround (in case it helps others)

Since `line` and `column` are correct (column is a 1-based char count within the line), the byte offset can be re-derived from the source:

```rust
fn byte_offset(source: &str, line: usize, col: usize) -> usize {
    let (mut cur_l, mut cur_c) = (1usize, 1usize);
    for (i, ch) in source.char_indices() {
        if cur_l == line && cur_c == col { return i; }
        if ch == '\n' { cur_l += 1; cur_c = 1; } else { cur_c += 1; }
    }
    source.len()
}
```

Combined with `Word::value.len()` (which is reliable) you can get an accurate `(start, end)` byte span. Also note: `loc.end.line/column` is *also* sometimes inconsistent for compound spans (we've seen `end < start` on multi-line constructs), so for `Word`s it's safest to use `start + value.len()`.

## Suggested fix

Track byte offsets directly in the tokenizer (rather than converting from char counts), and ensure `SourceSpan::end` is computed from the same coordinate as `start`.

## Versions

- `brush-parser = \"=0.4.0\"`
- Reproduced on rustc 1.94.1, Linux x86_64

lu-zero:
## Summary

- `SourcePosition.index` now tracks **byte offsets** rather than UTF-8 character counts, fixing off-by-one drift per multi-byte character that broke any downstream use of the index for source slicing (diagnostics, syntax highlighting, LSP, etc.)
- Two fixes in the tokenizer: `next_char()` increments by `ch.len_utf8()` instead of `1`, and the whitespace-skip branch does the same
- Updated doc comments on `SourcePosition`, `SourcePositionOffset`, and `SourceSpan::length()` to reflect byte semantics
- Added two tests verifying correct byte indices for tokens following multi-byte UTF-8 characters

Closes #1127

## Breaking change

`SourcePosition.index` previously documented as a "character index" now returns a byte offset. Any downstream consumer relying on character-count semantics will need to adjust.

### DIFF
--- a/brush-core/src/callstack.rs
+++ b/brush-core/src/callstack.rs
@@ -191,7 +191,7 @@ impl Frame {
         let mut new_start = if let Some(existing_start) = &self.source_info.start {
             if let Some(current) = pos {
                 Some(Arc::new(crate::SourcePosition {
-                    index: existing_start.index + current.index,
+                    offset: existing_start.offset + current.offset,
                     line: existing_start.line + (current.line - 1),
                     column: if current.line <= 1 {
                         existing_start.column + (current.column - 1)
@@ -214,7 +214,7 @@ impl Frame {
                 Some(Arc::new(pos))
             } else {
                 Some(Arc::new(crate::SourcePosition {
-                    index: 0,
+                    offset: 0,
                     line: self.current_line_offset + 1,
                     column: 1,
                 }))
--- a/brush-interactive/src/highlighting.rs
+++ b/brush-interactive/src/highlighting.rs
@@ -188,13 +188,13 @@ impl<'a, SE: brush_core::ShellExtensions> Highlighter<'a, SE> {
             for token in tokens {
                 match token {
                     brush_parser::Token::Operator(_op, token_location) => {
-                        let start = global_offset + byte_offset(token_location.start.index);
-                        let end = global_offset + byte_offset(token_location.end.index);
+                        let start = global_offset + byte_offset(token_location.start.offset);
+                        let end = global_offset + byte_offset(token_location.end.offset);
                         self.append_span(HighlightKind::Operator, start..end);
                     }
                     brush_parser::Token::Word(w, token_location) => {
-                        let start_byte = byte_offset(token_location.start.index);
-                        let end_byte = byte_offset(token_location.end.index);
+                        let start_byte = byte_offset(token_location.start.offset);
+                        let end_byte = byte_offset(token_location.end.offset);
 
                         // Parse the raw slice from `line`, not `w.as_str()`: the tokenizer may
                         // drop chars from `w` (e.g. `\<newline>` continuations), so offsets into
--- a/brush-parser/src/ast.rs
+++ b/brush-parser/src/ast.rs
@@ -2368,7 +2368,7 @@ echo there
             SourcePosition {
                 line: 1,
                 column: 1,
-                index: 0
+                offset: 0
             }
         );
         assert_eq!(
@@ -2376,7 +2376,7 @@ echo there
             SourcePosition {
                 line: 2,
                 column: 11,
-                index: 18
+                offset: 18
             }
         );
     }
@@ -2404,7 +2404,7 @@ my_func
             SourcePosition {
                 line: 1,
                 column: 1,
-                index: 0
+                offset: 0
             }
         );
         assert_eq!(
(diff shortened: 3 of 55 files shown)


lu-zero:
@reubeno not sure if it this alone warrants bumping to 0.5 but staying API-compatible would be more churn ...

reubeno:
@lu-zero What do you think about changing the field name to reflect the meaning change? That seems like something we ought to do so that all downstream consumers are forced to reevaluate whether their logic is still correct.

lu-zero:
if you prefer we can, but the new release is already breaking enough that...

lu-zero:
using `offset` is probably also more descriptive of what is it.

reubeno:
> using `offset` is probably also more descriptive of what is it.

Sure.

My main concern is avoiding a situation where the meaning of a field has changed but its name hasn't. Even if there are other breaking API changes, I'd want to lean on a compile break to at least give callers a chance to consider whether they need to change their logic.

ssilvius:
Ran into this too. Building something on brush-parser, and the AST spans count chars while word::parse counts bytes, so anything with a multi-byte char earlier in the line slices wrong.

Saw this has been sitting a while. I can run it against our log of ~47k real shell commands and report back, and rebase it if that helps.

PR #1146, opened 2026-05-06
fix(history): emit diagnostic when HISTFILE is unset for -a/-w

### LINKED ISSUE #1135: bug: `history -a/-w/-n/-r` silently no-op on unset HISTFILE; bash emits diagnostic

dashitongzhi:
## Summary

When `HISTFILE` is unset and no explicit file argument is provided, `history -a` and `history -w` previously silently succeeded with exit 0. Bash emits a diagnostic to stderr in the same situation.

This PR fixes the issue by emitting:
```
brush: history: HISTFILE: parameter null or not set
```
to stderr when the history file path resolves to `None`, matching bash's exact wording for compat-script friendliness.

## Changes

- `brush-builtins/src/history.rs`: Converted `if let Some` to `match` for both `-a` (append_session_to_file) and `-w` (write_session_to_file) code paths, adding a `None` arm that writes the diagnostic to stderr.

## Notes

- `-n` and `-r` already return "not yet implemented" errors, so they don't need the HISTFILE check yet.
- The fix preserves the existing exit code (0) to match bash behavior.

Fixes #1135

### DIFF
--- a/brush-builtins/src/history.rs
+++ b/brush-builtins/src/history.rs
@@ -50,6 +50,7 @@ pub(crate) struct HistoryCommand {
 
 struct HistoryConfig {
     default_history_file_path: Option<PathBuf>,
+    shell_name: String,
     time_format: Option<String>,
 }
 
@@ -63,6 +64,10 @@ impl builtins::Command for HistoryCommand {
         // Retrieve the shell's history config while we still can.
         let config = HistoryConfig {
             default_history_file_path: context.shell.history_file_path(),
+            shell_name: context
+                .shell
+                .current_shell_name()
+                .map_or_else(|| String::from("brush"), |name| name.to_string()),
             time_format: context.shell.history_time_format(),
         };
 
@@ -135,7 +140,8 @@ impl HistoryCommand {
                 None => {
                     writeln!(
                         stderr,
-                        "brush: history: HISTFILE: parameter null or not set"
+                        "{}: history: HISTFILE: parameter null or not set",
+                        config.shell_name
                     )?;
                 }
             }
@@ -167,7 +173,8 @@ impl HistoryCommand {
                 None => {
                     writeln!(
                         stderr,
-                        "brush: history: HISTFILE: parameter null or not set"
+                        "{}: history: HISTFILE: parameter null or not set",
+                        config.shell_name
                     )?;
                 }
             }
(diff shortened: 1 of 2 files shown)


reubeno:
@dashitongzhi Thanks for the contribution!

The change looks relatively straightforward. I've unblocked the PR checks to run and will take another look.

brush-builtins/src/history.rs
```diff
@@ -117,16 +117,24 @@ impl HistoryCommand {
...
+                        file_path,
+                        true,                         /* append? */
+                        true,                         /* unsaved items only */
+                        config.time_format.is_some(), /* write timestamps? */
+                    )?;
+                }
+                None => {
+                    writeln!(
+                        stderr,
+                        "brush: history: HISTFILE: parameter null or not set"
```
reubeno: As copilot's auto-comments called out, `brush-core` is embeddable in differently name shell apps, so we shouldn't hard-code `brush`.

If you look elsewhere in the code base you can see examples of what other builtins do. (Most of them probably omit the shell name, although the shell name can be dynamically queried via `.current_shell_name()` on the shell object.)

Could you please update this and the other error below accordingly?

reubeno (changes requested):
Thanks for the contribution, and for your patience. I left one comment that I'd like you to address; other than that it's looking good.

PR #1147, opened 2026-05-06
fix(builtins): handle quoted assignment arguments in declare

lu-zero:
When declare receives a single-quoted argument like 'arr=()', the parser cannot recognize it as an AssignmentWord (leading quote is not a valid variable name start). After word expansion strips the quotes, the builtin receives CommandArg::String("arr=(${X})") but never detected the = sign, treating the entire string as a variable name and failing validation.

Extend declaration_to_name_and_value to split on = for CommandArg::String args, handling both name=value and name[index]=value patterns. For scalar values the expanded string is used as-is (word expansion already handled quoting correctly). For array values with -a, add lift_string_array_assignment that reconstructs a declare command and runs it through the shell to let the parser and expander handle array parsing and parameter expansion at runtime.

This fixes the Gentoo rpm.eclass pattern:
  IFS=, declare -a 'types=()'

### DIFF
--- a/brush-builtins/src/declare.rs
+++ b/brush-builtins/src/declare.rs
@@ -151,7 +151,15 @@ impl builtins::Command for DeclareCommand {
                         result = ExecutionResult::general_error();
                     }
                 } else {
-                    if !self.process_declaration(&mut context, declaration, verb)? {
+                    let ok = if self.make_indexed_array.is_some()
+                        && Self::is_string_array_assignment(declaration)
+                    {
+                        self.lift_string_array_assignment(&mut context, declaration, verb)
+                            .await?
+                    } else {
+                        self.process_declaration(&mut context, declaration, verb)?
+                    };
+                    if !ok {
                         result = ExecutionResult::general_error();
                     }
                 }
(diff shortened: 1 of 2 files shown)


reubeno:
Thanks for putting this together; I hit this while looking at bash completion tests.

brush-builtins/src/declare.rs
```diff
@@ -339,6 +347,53 @@ impl DeclareCommand {
...
     }
 
+    /// Returns true if this declaration is a `CommandArg::String` containing
+    /// an `=` with an array-like value starting with `(`.
+    fn is_string_array_assignment(declaration: &brush_core::CommandArg) -> bool {
+        if let brush_core::CommandArg::String(s) = declaration {
+            if let Some((_name, value)) = s.split_once('=') {
+                return value.starts_with('(');
+            }
+        }
```
Elsie19: This can be collapsed:


```suggestion
        if let brush_core::CommandArg::String(s) = declaration
            && let Some((_, value)) = s.split_once('=')
        {
            return value.starts_with('(');
        }
```

Elsie19 (changes requested):

brush-builtins/src/declare.rs
lu-zero: sure, let me fold it in. Thank you :)

PR #1153, opened 2026-05-10
fix: Trailing path filters glob results

adil192:
This PR removes glob results if they are followed up with a literal.
E.g. the star in `*/file.txt` should only match directories that contain a `file.txt`.

Included is a positive test and regression tests to make sure nothing's removed that shouldn't be.

Also fixes the following tests:
- Glob with trailing slash matches directories
- Globstar disabled by default

I've added a failing test for tilde expansion that I didn't get around to fixing. If it's offtopic, I can remove it.

### DIFF
--- a/brush-core/src/patterns.rs
+++ b/brush-core/src/patterns.rs
@@ -261,13 +261,20 @@ impl Pattern {
                 matches!(piece, PatternPiece::Pattern(_))
                     && requires_expansion(piece.as_str(), self.enable_extended_globbing)
             }) {
-                for p in &mut paths_so_far {
+                paths_so_far.retain_mut(|p| {
                     let flattened = component
                         .iter()
                         .map(|piece| piece.as_str())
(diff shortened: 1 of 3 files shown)


reubeno:
Thanks for the contribution, @adil192!

I particularly appreciate the added tests. Added tests that track not-yet-fixed issues are always welcome.

I've unblocked the PR checks to run and we'll monitor their results.

adil192:
@reubeno Do I need to do anything to fix the bash-completion test suite, or is it supposed to fail?

reubeno:
> @reubeno Do I need to do anything to fix the bash-completion test suite, or is it supposed to fail?

Good question. We've not (yet!) hit 100% pass rate on the `bash-completion` tests. Our pass rate has been steadily on the rise as we improve compatibility, and is presently hovering at about what you see in this PR.

We watch for drops/increases to look for regressions/improvements.

PR #1155, opened 2026-05-11
docs: community Fedora installation instructions to README.md

Its-J:
I added this to the Terra repository, here's a link to the [PR](https://github.com/terrapkg/packages/pull/12197).

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,18 @@ nix run 'github:NixOS/nixpkgs/nixpkgs-unstable#brush' -- --version
 
 </details>
 
+<summary>🐧 <b>Installing on Fedora and Derivatives</b></summary>
+
+Fedora Linux users must first install the [Terra repository]([https://archlinux.org/packages/extra/x86_64/brush/](https://terrapkg.com/)).
+
+Then install `brush`:
+
+```bash
+dnf install brush-shell
+```
+
+</details>
+
 <details>
 <summary> 🔨 <b>Building from sources</b></summary>
 
--- a/README.md
+++ b/README.md
@@ -154,14 +154,15 @@ nix run 'github:NixOS/nixpkgs/nixpkgs-unstable#brush' -- --version
 
 </details>
 
+<details>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,6 +154,19 @@ nix run 'github:NixOS/nixpkgs/nixpkgs-unstable#brush' -- --version
...
 
+<details>
+<summary>🐧 <b>Installing on Fedora and Derivatives</b></summary>
+
+Fedora Linux users must first install the [Terra repository](https://terrapkg.com/).
+
+Then install `brush`:
+
+```bash
+dnf install brush
```
adil192: Should this be `brush-shell`?
`brush` gets autocorrected to `brush-shell`:
```console
ahann@dellpro:~$ sudo dnf install brush
Place your right index finger on the fingerprint reader
Updating and loading repositories:
Repositories loaded.
Package                             Arch      Version                              Repository              Size
Installing:
 brush-shell                        x86_64    0:0.4.0-1.fc44                       terra                7.3 MiB

Transaction Summary:
 Installing:         1 package

Total size of inbound packages is 2 MiB. Need to download 2 MiB.
After this operation, 7 MiB extra will be used (install 7 MiB, remove 0 B).
Is this ok [y/N]: n
Operation aborted by the user.
```

Its-J: It can be installed with either `brush` or `brush-shell`. The `Provides: brush` line in the [.spec file](https://github.com/terrapkg/packages/pull/12197/files) allows for this. Would you prefer if it was `dnf in brush-shell` in your README?

adil192: I'm not the owner of this repo, but I would personally prefer `brush-shell` since not all dnf commands will work with just `brush`.
E.g. `dnf repoquery brush` doesn't show anything, while `dnf repoquery brush-shell` does.

Its-J: Makes sense. Fixed!

reubeno:
Apologies for such a long delay! I've made some wording adjustments and will get this merged shortly.

PR #1156, opened 2026-05-13
feat: reduce allocations

Elsie19:
Some of the things I did:
* Initialize with `with_capacity` where applicable
* Delay or remove vector initialization
* Replace individual collection pushes to one `extend`
* Replace some function arguments with `IntoIterator` bounds for looser constraints
* Use `Cow` when appropriate
* Replace `Vec` in function arguments with `&[]`
* Replace `map_or_else` with `map_or` when default value is static

I also replaced these functions entirely:

```diff
- uppercase_first_char
- lowercase_first_char
+ pattern_to_first_char
```

and

```diff
- uppercase_pattern
- lowercase_pattern
+ pattern_to_string
```

Where the new functions take a closure over their respective argument and can transform it at the call site. You may want to pull out that specific commit and merge it outside of this PR, but I added it here anyways.

Feel free to pick and choose which ones to keep and leave! If you need I can merge likewise commits into one to make the merge cleaner.


lu-zero (approved):
I read it twice and it looks very nice.

brush-core/src/expansion.rs
```diff
@@ -1876,64 +1885,26 @@ impl<'a, SE: extensions::ShellExtensions> WordExpander<'a, SE> {
...
-        pattern: Option<&patterns::Pattern>,
-    ) -> Result<String, error::Error> {
+        transform: F,
+    ) -> Result<String, error::Error>
+    where
+        F: Fn(&str) -> String,
+    {
         if let Some(pattern) = pattern {
             if !pattern.is_empty() {
                 let regex = pattern.to_regex(false, false)?;
```
lu-zero, line 1896: Probably we could cache them in the future

reubeno:
Thanks for putting this all togther, @Elsie19!

I've read through everything other than `expansion.rs` so far -- and it's all looking pretty good. Your changes have left the code cleaner, more idiomatic, and definitely avoided some unneeded allocations :smile:.

There are 2 API surface changes flagged by the PR -- I think both of them are warranted, and they're relatively low impact, so I'm supportive of making them.

I'll spend some more time looking through `expansion.rs`, particularly given that the `expand_one_string` microbenchmark noted a possible perf change.

lu-zero:
It does crash with https://github.com/reubeno/brush/pull/1069 so once it lands I'd re-roll that branch and if we have time would be nice to land it as follow up.

reubeno:
> It does crash with #1069 so once it lands I'd re-roll that branch and if we have time would be nice to land it as follow up.

@lu-zero Any more info on what the crash is? Is there anything you'd recommend get adjusted in this PR before we merge? I just finished reviewing the latest changes (for this PR) and it's looking pretty good to me.

lu-zero:
Just I managed to typoing clash with crash... It works but on the portage-repo it seems to not speed up anything :/

reubeno (approved):

PR #1160, opened 2026-05-15
README.md: Link Windows(MSYS2) package

oech3:


README.md
```diff
@@ -115,6 +116,18 @@ pacman -S brush
 
 </details>
 
+<details>
+<summary>🐧 <b>Installing on MSYS2 or Windows</b></summary>
```
reubeno: I'm okay advertising how to install on `msys`, but I don't think I'd be ready to suggest that the *recommended* way to run `brush` on Windows is to use the `msys`/`mingw`-targeted installation.

I still would like us to pursue usability with the natively compiled version as well.

Could you adjust this, please, to specifically call out that this is for `msys` use (as opposed to generic Windows use)?

README.md
```diff
@@ -55,6 +55,7 @@
 $ cargo binstall brush-shell         # using cargo-binstall
 $ brew install brush                 # using Homebrew
 $ pacman -S brush                    # Arch Linux
+$ pacman -S mingw-w64-ucrt-x86_64-brush # or mingw-w64-clang-x86_64-brush or mingw-w64-clang-aarch64-brush # MSYS2
```
reubeno: The intention was to keep this list quite compact (with full details below); I'm inclined to leave this line out for now -- the details you've added below should still cover things.

reubeno (changes requested):

README.md
oech3: Removed about Windows.

README.md
oech3: removed

reubeno (approved):

README.md
oech3: https://github.com/reubeno/brush/discussions/1194

PR #1164, opened 2026-05-17
fix: syntax highlighting in light mode

adil192:
Fixes unreadable text in light mode (sue me!) by changing from `Color::White` to `Color::Default`:
image

Dark mode still works as before:
image


reubeno:
Thanks for the change! It seems reasonable.

I’ve unblocked the PR checks to run and will monitor them.

reubeno (approved):

Issue #1183, opened 2026-06-03
Add an exec/open interception hook to `ShellExtensions` so embedders can confine commands in-process


hartsock:
### Summary

`brush-core` is explicitly designed to be embedded in other Rust programs
(`Shell<SE: ShellExtensions>`, the builder, the `examples/custom-builtin.rs`
guide). A natural and increasingly common reason to embed a shell is to run
*untrusted* or *semi-trusted* scripts — for example, the action plans produced
by an LLM agent — while applying a host-defined policy about which programs may
run and which files may be touched.

Today an embedder cannot reliably enforce such a policy **in-process**. There is
no hook on the path from "the shell decided to run an external command / open a
file" to "the OS call happens." The only extension point on `ShellExtensions` is
`ErrorFormatter`, which is about presentation, not authority.

### The concrete gap

External command dispatch in `brush-core/src/commands.rs`
(`CommandExecutionInfo::execute`) has two branches:

```rust
// brush-core/src/commands.rs (around the resolution logic)
if !sys::fs::contains_path_separator(&self.command_name) {
    // PATH search + builtin table consulted here ...
    if let Some(path) = path {
        self.execute_via_external(&path)
    } else {
        Err(ErrorKind::CommandNotFound(self.command_name).into())
    }
} else {
    // Path-separator branch: PATH and the builtin table are BYPASSED.
    let command_name = PathBuf::from(self.command_name.clone());
    self.execute_via_external(command_name.as_path())
}
```

This is correct bash behavior — a command whose name contains a `/` (e.g.
`/bin/rm`, `./script`, `../x`) is run directly without a PATH search and without
considering builtins. But it means an embedder who tries to confine execution by
inspecting names/PATH cannot stop `/bin/rm`: it skips exactly the machinery they
might hook. Any name-based or PATH-based gate is trivially bypassed by spelling
the command with a path separator.

File opens have the same shape: redirections (`> file`, `< file`, `>> file`) and
`source`/`.` all funnel through `Shell::open_file` in
`brush-core/src/shell/fs.rs`, but there is no hook there either.

### Proposal

Add a new optional component trait to `ShellExtensions` — call it
`CommandInterceptor` — with two hooks that default to "allow," so existing
embedders and the default shell are byte-for-byte unchanged:

```rust
pub enum ExecDecision { Allow, Deny(String) }
pub enum OpenDecision { Allow, Deny(String) }

pub trait CommandInterceptor: Clone + Default + Send + Sync + 'static {
[...]

PR #1184, opened 2026-06-03
feat(core): add CommandInterceptor exec/open hook to ShellExtensions

hartsock:
Addresses #1183.

### What

Adds an optional `CommandInterceptor` component to `ShellExtensions` with two default-allow hooks — `before_exec(program, args)` and `before_open(request)` — letting an embedding host deny external command execution and file opens **in-process**. Default behavior is unchanged.

### Why

See #1183. Embedders cannot currently confine command execution in-process, and the path-separator dispatch branch (`/bin/rm`, `./x`) bypasses `PATH` and the builtin table, defeating any name- or PATH-based gate. This is the minimal seam that closes it.

### How

- New sub-trait `extensions::CommandInterceptor` (mirroring `ErrorFormatter`), collected as `ShellExtensions::CommandInterceptor`, with `DefaultCommandInterceptor` (allow-all). `ShellExtensionsImpl` gains a second type parameter defaulted to `DefaultCommandInterceptor`.
- `Shell<SE>` stores a `command_interceptor: SE::CommandInterceptor` instance (parallel to `error_formatter`), wired through `CreateOptions`/the builder, `new`, `Default`, and `Clone`. New accessor `Shell::command_interceptor()`.
- `before_exec` is invoked once, at the top of `commands::execute_external_command` — the single funnel for **both** dispatch branches. Denials return a new `ErrorKind::ExecDenied(program, reason)`, mapped to exit code 126.
- `before_open` is invoked in `Shell::open_file`, which covers redirections and `source`/`.`. Denials return an `io::Error(PermissionDenied)`, which the existing `RedirectionFailure` / `FailedSourcingFile` wrapping reports cleanly.

### ⚠️ Compatibility: this is **not** purely additive. Please read this part.

An earlier version of this description claimed the change was "purely additive". **That was wrong**, and the correction matters more than the rest of the PR:

* ✅ `type X = ShellExtensionsImpl<MyFormatter>` **keeps compiling** — the new parameter is defaulted.
* ❌ `impl ShellExtensions for MyExtensions { type ErrorFormatter = …; }` **stops compiling.** The trait gains a *required* associated type and Rust has no stable associated-type defaults, so every third-party direct impl fails with `E0046: missing CommandInterceptor in implementation`.

Verified both directions with a throwaway crate depending on `brush-core` by path: the direct-impl form compiles against `main` and fails on this branch; the type-alias form compiles on both. `ShellExtensions` is `pub` and unsealed, so direct impls are legal today. The "Analyze public API" job already reports this under **Removed items**.

**This is unavoidable for any design that adds a required associated type** — it isn't a fixable detail of this patch. See the API question at the bottom; I'll implement whichever shape you prefer.

### Second commit: `before_open` no longer parses `Debug` output

The first cut recovered write-intent with `format!("{options:?}")` on a `std::fs::OpenOptions` and string-matched `write: ` / `append: `. That made a read-vs-write authority decision depend on the `Debug` rendering of a std type — unstable by contract, different on Windows, free to change in any release.

It was also unnecessary: **brush already knows the answer explicitly at all four `Shell::open_file` call sites.** The redirection setup in `interp.rs` matches on `ast::IoFileRedirectKind`; `setup_redirect_output_and_error_to` always creates and writes; `source`/`.` and the history loader open read-only.

So this adds `extensions::OpenAccess { Read, Write, ReadWrite }` and `extensions::OpenRequest { path, access }`, and threads the access through `Shell::open_file` from each call site — in the redirection path derived from the redirect kind itself, not recovered from the options afterwards. `Shell::open_file` is `pub(crate)`, so the extra parameter costs nothing in public API. `OpenRequest` is `#[non_exhaustive]` so more detail can be added later without breaking implementors.

Classification is unchanged for every existing case (`Read`/`DuplicateInput` read; `Write`/`Append`/`Clobber`/`DuplicateOutput` write); `ReadAndWrite` (`<>`) is now reported as `ReadWrite` instead of collapsing to "write".

`before_open`'s rustdoc now also states the one coverage gap: platform-specific special files are resolved before the path is made absolute and are not shown to the hook. Today that is only `/dev/null` on Windows; the Unix special-file hook resolves nothing.

### Test plan

`brush-core/tests/command_interceptor_tests.rs`:

- denies a bare-name command (`rm`, PATH-resolved);
- **denies an absolute-path command (`/bin/rm`)** — the load-bearing case, proving the path-separator branch is hooked;
- allows a permitted command (`/usr/bin/true`);
- denies an output redirection writing outside an allowed directory while allowing one inside it;
- allows a read-only `source` open;
- **new:** `declared_access_matches_redirection_kind` pins the reported access for `<`, `>`, `>>`, `>|` and `<>` — the distinction the `Debug` parse could only approximate.

Validation at head `8456f09` against base `30a3bce`:

```
cargo xtask ci pre-commit
cargo test --test brush-compat-tests
```

- fmt ✅, build ✅, clippy `--workspace --all-features --all-targets` ✅, schema check ✅
- unit — **384 run, 384 passed, 1 skipped**
- integration (nextest `--workspace`) — **2703 run, 2703 passed, 40 skipped**
- compat — `2300 ran: 1844 succeeded, 0 failed, 456 known to fail, 39 skipped`, **identical to clean `main` `30a3bce`**, as a default-allow change should be.

One caveat reported honestly: `cargo deny --all-features check all` reports `advisories FAILED` — but it fails **identically on clean `main`** (yanked `chacha20`, plus a `deny.toml` ignore for `RUSTSEC-2026-0253` that no longer matches anything). Pre-existing, unrelated to this PR, and not something I've touched here.

### The API question — one decision, and I'll do the work

Since a required associated type on a public unsealed trait is breaking no matter how it's written, the choice isn't between four designs. It's one question:

> **Is `ShellExtensions` meant to be implemented directly by third parties, or only obtained through `ShellExtensionsImpl<…>`?**

**If only through `ShellExtensionsImpl`** — keep the associated type and **seal the trait in the same release**. This break then becomes the *last* one: every future component (the `PlaceholderBehavior` slot already in `extensions.rs` is advertised as a "stub for future extension") becomes purely additive, because implementors can only reach `ShellExtensions` via `ShellExtensionsImpl`, whose extra parameters are defaulted. Keeps static dispatch and zero default-path cost. **This is what I'd recommend**, and I'm happy to add the sealing to this PR.

**If it's meant to be directly implementable** — then a required associated type is off the table permanently, and the interceptor should be a runtime field instead: `Option<Arc<dyn CommandInterceptor>>` on `CreateOptions`/`Shell`, `#[builder(default)]`, `serde(skip)`. Fully source-compatible today (per `AGENTS.md`, new defaulted fields on a `Default` struct are fine), object-safe as written, and costs one `Option` branch per spawn/open — nothing next to `fork`/`exec`. The cost is a second extension mechanism sitting alongside `ShellExtensions`.

Either way the `before_open` argument shape above is unaffected, so that part is settled regardless.

Deliberately **out of scope** here, to keep this reviewable: a `before_command` hook (fires per command including builtins; the natural place to observe cancellation, which bounds a pure-builtin runaway like `while true; do :; done` that reaches neither hook) and a notion of a *terminating* error so a denial isn't swallowed as an exit status by an enclosing loop. Both exist in our downstream fork and can follow as separate PRs once the shape here is settled.

### Answers to my own earlier open questions

- Naming: happy to rename `CommandInterceptor` → `SecurityPolicy`/`Sandbox`, or `before_*` → `check_*`. No strong preference; say the word.
- `before_open` denial: still an `io::Error(PermissionDenied)` rather than a dedicated `ErrorKind::OpenDenied`, to fit the existing open-file error flow. Easy to change.
- `(program, args)` has been sufficient for the downstream policy so far.
- Sync has been sufficient; the enforcement points are synchronous.

---

> Per the project's AI Policy: this change was authored with AI assistance and reviewed and tested by a human contributor; both commits carry `Assisted-by:` trailers accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -433,6 +433,7 @@ dependencies = [
  "async-recursion",
  "async-trait",
  "bon",
+ "brush-builtins",
  "brush-parser",
  "cached",
  "cfg-if",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -416,6 +416,7 @@ dependencies = [
  "async-recursion",
  "async-trait",
  "bon",
+ "brush-builtins",
  "brush-parser",
  "cached",
  "cfg-if",
--- a/brush-core/Cargo.toml
+++ b/brush-core/Cargo.toml
@@ -85,5 +85,10 @@ uuid = { version = "1.23.1", features = ["js"] }
 
 [dev-dependencies]
 anyhow = "1.0.102"
+# Used only by integration tests that need a realistically-populated builtin
+# table (e.g. `echo`, `.`/`source`) to exercise the capability-interceptor
+# hooks. This is a dev-dependency cycle (brush-builtins depends on brush-core),
+# which Cargo permits for dev-dependencies.
+brush-builtins = { version = "^0.2.0", path = "../brush-builtins" }
 pretty_assertions = { version = "1.4.1", features = ["unstable"] }
 tempfile = "3.27.0"
--- a/brush-core/src/commands.rs
+++ b/brush-core/src/commands.rs
@@ -579,6 +579,23 @@ pub(crate) fn execute_external_command(
         })
         .collect::<Vec<_>>();
 
+    // Give the configured command interceptor (capability confinement) a chance
+    // to deny this external spawn. This site is the single funnel for *all*
+    // external commands, including the path-separator branch (e.g. `/bin/rm`,
+    // `./x`) that bypasses PATH and the builtin table, so a policy here cannot
+    // be circumvented by spelling the command differently.
+    {
+        use crate::extensions::CommandInterceptor as _;
+        let hook_args: Vec<String> = cmd_args.iter().map(|s| (*s).clone()).collect();
+        if let extensions::ExecDecision::Deny(reason) = context
+            .shell
+            .command_interceptor()
+            .before_exec(executable_path, hook_args.as_slice())
+        {
+            return Err(error::ErrorKind::ExecDenied(context.command_name.clone(), reason).into());
+        }
+    }
+
     // Before we lose ownership of the open files, figure out if stdin will be a terminal.
     let child_stdin_is_terminal = context
         .try_fd(openfiles::OpenFiles::STDIN_FD)
--- a/brush-core/src/error.rs
+++ b/brush-core/src/error.rs
@@ -318,6 +318,12 @@ pub enum ErrorKind {
     /// A glob pattern failed to match any files (failglob).
     #[error("no match: {0}")]
     NoMatch(String),
+
+    /// Execution of an external command was denied by the configured
+    /// command interceptor (capability confinement). The first field is the
+    /// program; the second is the reason supplied by the interceptor.
+    #[error("{0}: execution denied: {1}")]
+    ExecDenied(String, String),
 }
 
 /// Trait implementable by built-in commands to represent errors.
@@ -363,6 +369,7 @@ impl From<&ErrorKind> for results::ExecutionExitCode {
             ErrorKind::FunctionParseError(..) => Self::InvalidUsage,
             ErrorKind::TestCommandParseError(..) => Self::InvalidUsage,
             ErrorKind::FailedToExecuteCommand(..) => Self::CannotExecute,
+            ErrorKind::ExecDenied(..) => Self::CannotExecute,
             ErrorKind::FunctionNameShadowsSpecialBuiltin { .. } => Self::InvalidUsage,
             ErrorKind::IoError(io_err) => io_err.into(),
             ErrorKind::BuiltinError(inner, ..) => inner.as_exit_code(),
--- a/brush-core/src/extensions.rs
+++ b/brush-core/src/extensions.rs
@@ -1,5 +1,7 @@
 //! Definition of shell behavior traits and defaults.
 
+use std::path::Path;
+
 use crate::{Shell, error, extensions};
 
 /// Trait for static shell extensions. Collects all associated types needed to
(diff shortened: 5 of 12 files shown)


hartsock:
Committed to landing this one — I'll keep the branch rebased on `main` so it stays mergeable whenever you have a chance to review. Thanks!

hartsock:
Full CI is green on my fork for this branch — all 30 jobs pass across macOS, Windows, Linux (x86_64/aarch64/musl), wasm32, and the OS-target distros: https://github.com/hartsock/brush/actions/runs/27522097280

hartsock:
@reubeno just a note from me to say I am highly invested in this patch. I am working on an Object Capability security harness and I think _working inside_ a shell like Brush is a major breakthrough in future security work that will ONLY be possible with a project like this. Thanks for taking on such a BEAST and doing a good job managing the community!

reubeno:
> @reubeno just a note from me to say I am highly invested in this patch. I am working on an Object Capability security harness and I think _working inside_ a shell like Brush is a major breakthrough in future security work that will ONLY be possible with a project like this. Thanks for taking on such a BEAST and doing a good job managing the community!

Thanks for checking in on this @hartsock, and also for your patience! I and we appreciate your enthusiasm and contributions :smile:.

I'm also very supportive of getting these kinds of hooks added -- I think it's just a matter of making sure we can arrive at the right API surface. June was rough for me, but I should hopefully have more time in the next few weeks to engage with you here on more detailed review.

hartsock:
Found alongside #1242 and #1244 during practical testing of brush embedded as a shell library.

hartsock:
@reubeno — you said this came down to the right API surface, so I've reduced it to one question. Rebased onto `30a3bce`; head `8456f09`.

**I was wrong to call this "purely additive."** True for the alias form, not for direct impls. A throwaway crate depending on `brush-core` by path:

- `impl ShellExtensions for MyExtensions { type ErrorFormatter = …; }` — compiles on `main`, fails here with `E0046: missing CommandInterceptor`.
- `type X = ShellExtensionsImpl<DefaultErrorFormatter>` — compiles on both; the new parameter is defaulted.

`ShellExtensions` is `pub` and unsealed, and Rust has no stable associated-type defaults, so **any** design adding a required associated type breaks direct impls. "Analyze public API" flags it under Removed items.

**`before_open` no longer parses `Debug` output.** The first cut took write-intent from `format!("{options:?}")` — a read-vs-write authority decision keyed on an unstable, platform-varying rendering, and needlessly, since all four `Shell::open_file` call sites already know their intent. The second commit adds `OpenAccess`/`OpenRequest` and threads it from each; `open_file` is `pub(crate)`, so no public cost. `<>` now reports `ReadWrite` instead of collapsing to "write", and a new test pins `<`, `>`, `>>`, `>|`, `<>`.

**The question**, since a required associated type breaks direct impls either way:

> Is `ShellExtensions` meant to be implemented directly, or only obtained through `ShellExtensionsImpl<…>`?

- **Only via `ShellExtensionsImpl`** → keep the associated type and **seal the trait in the same release**. That makes this the last such break: implementors could then only reach the trait through `ShellExtensionsImpl`, whose extra parameters are defaulted, so future components — the `PlaceholderBehavior` slot included — land additively. **My recommendation**; I'll add the sealing here.
- **Directly implementable** → a required associated type is permanently out, and this becomes a runtime `Option<Arc<dyn CommandInterceptor>>` on `CreateOptions`/`Shell`: source-compatible today, object-safe as written, one `Option` branch per spawn/open, at the cost of a second extension mechanism beside `ShellExtensions`.

Happy to implement either; `before_open`'s shape stands regardless. A `before_command` hook — per-command, builtins included, where cancellation would be observed to bound a pure-builtin runaway — exists in our fork and waits for its own PR.

Validation at `8456f09`: `cargo xtask ci pre-commit` clean; unit **384/384**, integration **2703/2703**; compat `2300 ran: 1844 succeeded, 0 failed, 456 known to fail` — identical to clean `main`. (`cargo deny` reports `advisories FAILED`, but identically on clean `main`: yanked `chacha20`, plus a `deny.toml` ignore for `RUSTSEC-2026-0253` matching nothing. Unrelated to this PR — flagging in case it's news.)

hartsock:
Still working this.

hartsock:
Closing this. It's the wrong seam, and I should have known that in June.

In #1183 you asked me to look at #972 and say whether the statically registered filters
would serve our use case. I didn't answer, and kept developing a parallel mechanism here
instead. Having now read #972 and `filter-architecture.md`: `CmdExecFilter` covers more of
what we need than this PR does — including `pre_simple_cmd`, which sees builtins and
functions, and which I'd left out entirely.

I was also wrong on the shape. I argued here that the associated type needed a decision from
you because it breaks a third-party `impl ShellExtensions` with `E0046`, and I was moving
toward a runtime `Option<Arc<dyn …>>`. #972 had already taken that break, and #838 says the
point of `Shell<SE>` is precisely to avoid `dyn`. Sorry for the noise.

I've written up the use case and a review of #972 — what it already covers, and four gaps I
can evidence — on #1183. The one worth flagging here: **the `exec` builtin bypasses both
this PR's hook and #972's**. `brush-builtins/src/exec.rs` composes its own `Command` and
calls `CommandExt::exec`, so `exec /bin/rm` replaces the shell process without any hook
firing. I have a regression test for it that runs in a dedicated child process, and I'd like
to contribute that against whichever mechanism you land.

Not deleting the branch, in case the tests are useful to lift.

PR #1187, opened 2026-06-05
docs: document background pid limitation

mtanski:
Documents the current `$!` limitation for background commands, since brush does not yet reliably expose the PID of the most recent asynchronous command.

### DIFF
--- a/docs/reference/compatibility.md
+++ b/docs/reference/compatibility.md
@@ -139,6 +139,19 @@ job2 &
 wait -n  # Wait for whichever finishes first
 ```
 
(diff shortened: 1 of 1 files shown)


reubeno (approved):

reubeno:
That's a good call-out, thanks for the doc update.

Did you spot an existing GitHub issue on it as well? We should file one if not.

mtanski:
I did check GH issues and code docs but I did not find anything resembling in either one. I will file one.

reubeno (approved):

PR #1188, opened 2026-06-05
fix(core): resolve stdio dev paths through shell fds

mtanski:
Brush already routes /dev/fd/N through the shell execution fd map, but /dev/stdin, /dev/stdout, and /dev/stderr reopened the host process stdio. That breaks cases like sourcing /dev/stdin with a heredoc, where the builtin should consume the redirected stdin.

This resolves those stdio aliases through the same path as /dev/fd/N and adds a compatibility case for source /dev/stdin with a heredoc.

This behavior has been validated in our real environment, running real real scripts/shell pipelines.

### DIFF
--- a/brush-core/src/shell/fs.rs
+++ b/brush-core/src/shell/fs.rs
@@ -195,13 +195,10 @@ impl<SE: crate::extensions::ShellExtensions> crate::Shell<SE> {
 
         let path_to_open = self.absolute_path(path.as_ref());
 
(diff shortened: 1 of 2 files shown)


reubeno:
Hi @mtanski -- thanks for the contribution!

This is a well-authored, clean change and we appreciate the added test along with it 🙂

I've unblocked the PR checks to run and will monitor them.

PR #1189, opened 2026-06-05
feat(core): support cooperative SIGINT cancellation

mtanski:
Brush currently handles Ctrl-C differently depending on what is running. Bash sends SIGINT to the foreground process group for external commands and pipelines, and it also interrupts shell-native evaluation. If an `INT` trap is installed, Bash runs the trap and lets it decide control flow; otherwise interrupted shell-native work returns the usual SIGINT status.

This brings Brush closer to that model for both shell-native execution and foreground processes. `ExecutionParameters` can now carry a host-requested signal, and centralized execution checkpoints let loops, functions, builtins, compound commands, and traps observe cancellation without manually instrumenting every AST node. Foreground process waits can wake on the same signal state and forward the signal to the child process group, including pipelines.

The interactive shell wires Ctrl-C into the active execution parameters. In a PTY, the terminal driver already delivers SIGINT to the foreground process group, so the interactive path avoids double-forwarding and uses the cooperative signal path to make Brush's own shell code see the interrupt.

For embedders that drive a shell over an API, `ExecutionParameters::request_host_signal(signal_number)` is the hook for injecting a signal into both shell-native execution and foreground children. `ExecutionParameters::request_forced_signal(signal_number)` is available for timeout/kill-style cancellation that should bypass shell traps.

### DIFF
--- a/brush-builtins/src/kill.rs
+++ b/brush-builtins/src/kill.rs
@@ -119,6 +119,15 @@ impl builtins::Command for KillCommand {
                 }
             } else {
                 let pid = brush_core::int_utils::parse(pid_or_job_spec.as_str(), 10)?;
+                let is_self_signal =
+                    i32::try_from(std::process::id()).is_ok_and(|current_pid| current_pid == pid);
+
+                if is_self_signal && context.shell.traps().handles(trap_signal) {
+                    if let Ok(signal_number) = i32::try_from(trap_signal) {
+                        context.params.request_host_signal(signal_number);
+                        return Ok(ExecutionResult::success());
+                    }
+                }
 
                 // It's a pid.
                 sys::signal::kill_process(pid, trap_signal)?;
--- a/brush-core/src/commands.rs
+++ b/brush-core/src/commands.rs
@@ -623,6 +623,7 @@ pub(crate) fn execute_external_command(
         if let Some(pgid) = process_group_id {
             cmd.process_group(pgid);
         }
+        cmd.reset_job_control_signals();
     }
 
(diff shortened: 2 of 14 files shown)


reubeno:
Thanks for the contribution!

This is an area I've long wanted to improve and appreciate you digging in.

I'll need more time to look through this than your other 2 PRs. One of the main things I'll be looking at is how the design will work for alternate (embedding) hosts of brush-core. As you've noted, our preexisting signal handling is inconsistent and wasn't too thoughtful about ownership of process signal configuration/management in an embedded host -- I'm hoping this will help a bit, given your description.

mtanski:
I have some additional improvements for signal handling that I plan on submitting later that help a lot with both embedding (at least in our environment) and conformance to bash behavior, but I need to time to rework them to make them upstream friendly.

Hopefully this one is pretty straightforward as it works for interactive login shells (brush interactive) and embedded cases when you need to emulated it (convert an API call to signal). Let me know if you have any questions.

mtanski:
@reubeno I ported the signaling changes from our private branch into this PR. As you can see this [fixes](https://github.com/reubeno/brush/pull/1189/changes#diff-39421b1a23b24016a1c433309cc38224ae72a98d3f308794c58e45360a41d461) a bunch bash signalizing/TRAP compatibility. This also should not hurt embedding (based on our 8mo experience), and exposes the cancel API in such a way it helps embedding cases cancel pipelines in a flexible way.

I see I have caused a windows build break, I will TAL when I have some time again.

PR #1196, opened 2026-06-12
fix(reedline): use theme-adaptive input styles

AdeAttwood:
Summary:

Keep Reedline input, completion menus, and history hints aligned with the terminal theme instead of forcing hard-coded colors. Reedline's default example highlighter painted unhighlighted input white, and our menu and hint styles also pinned specific colors, which made the interactive UI look wrong when syntax highlighting was disabled.

Install a plain-text highlighter whenever colors are enabled but syntax highlighting is not, and switch neutral menu and hint styles to attribute-only formatting. Add tests that assert these styles leave the terminal's foreground and background colors untouched.

Test Plan:

1. Run 'cargo fmt --check'.
2. Run 'cargo test --package brush-interactive'.
3. Run 'cargo test --package brush-interactive --features reedline'.

I have also been using this for a bit. Here are the screenshots.

Before

Screenshot 2026-06-12 at 18 46 16

Screenshot 2026-06-12 at 18 46 39

After

Screenshot 2026-06-12 at 18 43 07

Screenshot 2026-06-12 at 18 42 35

### DIFF
--- a/brush-interactive/src/reedline/highlighter.rs
+++ b/brush-interactive/src/reedline/highlighter.rs
@@ -73,6 +73,16 @@ pub(crate) struct ReedlineHighlighter<SE: brush_core::ShellExtensions> {
     pub shell: refs::ShellRef<SE>,
 }
 
+pub(crate) struct PlainTextHighlighter;
+
+impl reedline::Highlighter for PlainTextHighlighter {
+    fn highlight(&self, line: &str, _cursor: usize) -> reedline::StyledText {
+        let mut styled = reedline::StyledText::new();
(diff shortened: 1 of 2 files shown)


reubeno:
This is a great change -- thanks for working on it!

_(Aside: seeing this, it reminds me that we should find *some* reasonable way to allow users to influence the colors that are still hard-coded (e.g., "use green as the color for directories displayed in completion menus"). Now that we have a way to configure via our TOML file, we could at least add some properties there.)_

reubeno (approved):

PR #1197, opened 2026-06-14
fix(expansion): join $*/${a[*]} via IFS first char in scalar contexts

lu-zero:
## Summary

In scalar (non-field-splitting) contexts such as an assignment RHS, `basic_expand_to_str` converted the expanded fields to a string with a hardcoded space join (the `From<Expansion> for String` impl), ignoring both the expansion's `concatenate` flag and `IFS`.

So `x=${arr[*]}` (or `x=$*`) under `IFS=:` produced `a b c` instead of bash's `a:b:c`.

```sh
arr=(a b c)
IFS=:
x=${arr[*]}
echo "[$x]"   # bash: [a:b:c]   brush (before): [a b c]
```

This also breaks the common split-then-rejoin idiom (e.g. Gentoo's `llvm-utils` eclass), where `local IFS=:; sp=( ${PATH} ); export PATH=${sp[*]}` mangled every `:` separator into a space.

## Fix

Join the expanded fields following bash rules:
- `$*` / `${a[*]}` (`concatenate=true`) join with the **first character of IFS**
- `$@` / `${a[@]}` join with a **space**

A single field — the common scalar case — is unaffected.

## Tests

Verified against the bash oracle with new cases in `compat/ifs.yaml` (array/positional `[*]` and `[@]` joins in scalar assignment, the export-assignment PATH idiom, and split-then-rejoin). Full compat suite: 0 failures.

### DIFF
--- a/brush-core/src/expansion.rs
+++ b/brush-core/src/expansion.rs
@@ -592,7 +592,22 @@ impl<'a, SE: extensions::ShellExtensions> WordExpander<'a, SE> {
 
     /// Apply tilde-expansion, parameter expansion, command substitution, and arithmetic expansion.
     pub async fn basic_expand_to_str(&mut self, word: &str) -> Result<String, error::Error> {
-        Ok(String::from(self.basic_expand(word).await?))
+        let expansion = self.basic_expand(word).await?;
+        // Join the expanded fields for scalar (non-split) contexts following
+        // bash rules: `$*`/`${a[*]}` (concatenate=true) join with the first
+        // character of IFS, while `$@`/`${a[@]}` join with a space. A single
+        // field (the common case) is unaffected. Without this, e.g.
+        // `x=${arr[*]}` under `IFS=:` would yield `a b c` instead of `a:b:c`.
+        let joiner = if expansion.concatenate {
+            self.shell.get_ifs_first_char()
+        } else {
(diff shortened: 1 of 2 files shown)


brush-core/src/expansion.rs
```diff
@@ -592,7 +592,22 @@ impl<'a, SE: extensions::ShellExtensions> WordExpander<'a, SE> {
 
     /// Apply tilde-expansion, parameter expansion, command substitution, and arithmetic expansion.
     pub async fn basic_expand_to_str(&mut self, word: &str) -> Result<String, error::Error> {
-        Ok(String::from(self.basic_expand(word).await?))
+        let expansion = self.basic_expand(word).await?;
+        // Join the expanded fields for scalar (non-split) contexts following
```
reubeno, line 596: Looks good! Do we need to adjust other code paths in this file too?

Note the `// TODO(IFS)` comments below, but not sure if the rules are different for regexes, patterns, etc.

Should we also remove the String::from support to avoid another code path from making the same mistake?

lu-zero: I did a very surgical change so it can land immediately, I can expand and do a full pass once this lands. (small steps)

reubeno (approved):

PR #1198, opened 2026-06-14
fix: build issues on some BSD-ish platforms + add CI checks

reubeno:


oech3:
Can we add `cargo check` only CI. Release build CI for all platform is overkilled and check sometimes avoids toolchain issue on minor targets.

reubeno:
> Can we add `cargo check` only CI. Release build CI for all platform is overkilled and check sometimes avoids toolchain issue on minor targets.

That's the fallback plan -- and also why this is in draft. If build is too hard to make work, we'll just check.

_Edited: took the fallback plan :smile:._

PR #1203, opened 2026-06-17
build(deps): upgrade cached 0.59 -> 2.0.2

reubeno:
## Summary

Upgrades the `cached` dependency from `0.59.0` to **`2.0.2`** (the latest published 2.x) in both `brush-core` and `brush-parser`, working through the 0.x → 1.x → 2.x breaking changes.

### Migration
- proc-macro path `cached::proc_macro::cached` → `cached::macros::cached`
- `size = N` → `max_size = N` on every `#[cached]`
- dropped `result = true` — `Result`-returning cached fns now skip caching `Err` by default (prior behavior preserved)
- `SizedCache` → `LruCache`; removed `with_size()` constructor → `LruCache::builder().max_size(64).build()`

### Clone reductions (bonus)
The default `#[cached]` key is built by **cloning** the args, so any function whose caller did `.to_owned()` just to pass owned args was cloning twice per call. Switched the parser cached fns (`arithmetic`/`word`/`tokenizer`) to borrowed args with explicit `key`/`convert`, moving the single needed clone into key-construction — **halves clones** on those hot paths (and on cache hits removes the redundant caller-side clone). Cascaded `&str` through the prompt and parsing chains so **no `#[allow]`s** are needed.

The prompt/parsing chains genuinely need their owned values (`.into_owned()` is borrow-checker-mandated against `&mut self`; `parse_string`'s `Into<String>` bound forces materialization), so those save no clones — documented in the commit rather than papered over.

### regex.rs
`build()` is now fallible, but `unwrap`/`expect`/`panic` are all denied lints. The thread-local LRU cache is now `Option<LruCache>` built via `.build().ok()`, **gracefully degrading to uncached regex compilation** if construction ever fails (never happens with a fixed `max_size`, but no panic).

## Testing
- `cargo build --workspace` ✅
- `cargo clippy -p brush-parser -p brush-core` ✅ (pedantic/nursery/cargo/perf all denied)
- `cargo test -p brush-parser -p brush-core` ✅ (all pass)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -739,15 +739,14 @@ checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
 
 [[package]]
 name = "cached"
-version = "0.59.0"
+version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "53b6f5d101f0f6322c8646a45b7c581a673e476329040d97565815c2461dd0c4"
+checksum = "dc0df7748fe2f601e376916ab19e7bfc2c74461b8abe3bce2ce20036ad8de38f"
 dependencies = [
  "ahash",
  "cached_proc_macro",
  "cached_proc_macro_types",
  "hashbrown 0.16.1",
- "once_cell",
  "parking_lot",
  "thiserror 2.0.19",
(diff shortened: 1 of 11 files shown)


reubeno:
@copilot please fix formatting to fix format check PR check failures.

PR #1221, opened 2026-06-26
fix(expansion): honor IFS only for expansion results, not literal text (#295)

dalpra-hcd:
Implements #295: Only portions of a word produced by expansions should be subject to IFS field splitting. Literal text must remain intact.

**Before:** `echo a:b:c` with IFS=: → 3 fields (incorrect)
**After:** `echo a:b:c` with IFS=: → 1 field (matches bash)

## Changes

Mark `WordPiece::Text` as `Unsplittable` in expansion.rs to prevent literal text from being field-split by IFS, while still allowing expanded text to be split correctly.

## Testing

- Added two test cases validating the behavior
- All `cargo xtask ci quick` checks pass

Assisted-by: GitHub Copilot

### DIFF
--- a/brush-core/src/expansion.rs
+++ b/brush-core/src/expansion.rs
@@ -909,7 +909,9 @@ impl<'a, SE: extensions::ShellExtensions> WordExpander<'a, SE> {
     ) -> Result<Expansion, error::Error> {
         let expansion: Expansion = match word_piece {
             brush_parser::word::WordPiece::Text(s) => {
-                Expansion::from(ExpansionPiece::Splittable(s))
+                // Literal text from the original word should not be split by IFS;
+                // only portions produced by expansions are split. Mark as unsplittable.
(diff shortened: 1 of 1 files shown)


reubeno (changes requested):
@dalpra-hcd -- thanks for diving in! I've run PR checks against this change, but it looks like a number of tests fail with the change.

I see that you've run `cargo xtask ci quick` -- that's a great starting point, and something that can be run very frequently throughout development. To run integration tests, you'll also want to run `cargo xtask ci pre-commit` -- or if you'd prefer, directly run `cargo test --workspace` or `cargo nextest --workspace`.

dalpra-hcd:
Thanks for your reply, I will try using those commands.

dalpra-hcd:
It seems this approach breaks other unit tests, thanks for the review, but I will close it for now.

PR #1227, opened 2026-06-30
fix(expansion): slice sparse indexed arrays by subscript

### LINKED ISSUE #1219: bug: array splitting expansion is incorrect for sparse arrays
### brush version

brush 0.4.0 (git:85f6283)

5.3

### Platform / OS

Linux

### Steps to reproduce

The array splitting expansion `${array[@]:<start>:<count>}` does not seem to be correct for sparse arrays. For sparse arrays, the `<start>` is still the actual index of the array element.

( Internally, fyi, as far as I know sparse arrays in Bash are stored as a linked list, and bash keeps a "lastref" pointer for fast "close by" navigation. )

### Actual behavior (brush)

```
$ brush -c 'a[1]=1; echo "${a[*]:1}"; declare -p a'

declare -a a=([1]="1")
```

### Expected behavior

```
$ bash -c 'a[1]=1; echo "${a[*]:0:1}"; declare -p a'
1
declare -a a=([1]="1")
```

### Additional context

_No response_

truffle-dev:
Substring expansion of an indexed array via `[@]`/`[*]` treated the offset as a position in the dense list of present elements. Bash interprets the offset as a subscript value: it selects elements whose subscript is at least the offset, and the length still counts elements positionally from there.

For the sparse array in #1219:

```bash
a=([1]=1 [5]=5 [9]=9)
echo "${a[*]:1}"   # brush: (empty)   bash: 1 5 9
echo "${a[*]:2}"   # brush: (empty)   bash: 5 9
echo "${a[*]:6}"   # brush: (empty)   bash: 9
echo "${a[*]:1:2}" # brush: (empty)   bash: 1 5
echo "${a[*]: -1}" # brush: (empty)   bash: 9
```

Element `1` sat at dense position 0, so a position-based offset of 1 skipped past it.

The fix translates the subscript-space offset onto a dense position before slicing. A non-negative offset maps to the first element whose subscript is at least the offset; a negative offset counts back from one past the largest subscript (matching bash). Scalars, positional parameters, and associative arrays keep their existing positional behavior, so contiguous arrays (`a=(x y z); ${a[*]:1}` -> `y z`) are unchanged.

Added a compat test (`Substring operator on sparse indexed array`) covering positive, negative, and length-bounded offsets against the bash oracle.

Fixes #1219.

### DIFF
--- a/brush-core/src/expansion.rs
+++ b/brush-core/src/expansion.rs
@@ -1361,24 +1361,53 @@ impl<'a, SE: extensions::ShellExtensions> WordExpander<'a, SE> {
                     );
                 }
 
+                // For an indexed array sliced via [@]/[*], bash interprets the offset as a
+                // subscript value rather than a position in the (possibly sparse) list of
+                // present elements. Capture the sorted subscripts so we can translate the
+                // offset below; positional parameters, scalars, and associative arrays keep
+                // their positional behavior.
+                let array_subscripts: Option<Vec<u64>> = if indirect {
+                    None
+                } else if let brush_parser::word::Parameter::NamedWithAllIndices { name, .. } =
+                    &parameter
+                {
+                    match self.shell.env().get(name) {
+                        Some((_, var)) => match var.value() {
+                            ShellValue::IndexedArray(array) => {
+                                Some(array.keys().copied().collect())
+                            }
+                            _ => None,
+                        },
+                        None => None,
+                    }
+                } else {
+                    None
+                };
+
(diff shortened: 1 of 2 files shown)


reubeno:
Thanks for getting the ball rolling with this, @truffle-dev! After review, I've spotted some possible issues and decided to merge some of your changes into #1283, and move forward with that PR instead of this one. We'll of course make sure to give you credit on the relevant commit.

Thanks for your contribution :smile:

PR #1228, opened 2026-07-05
fix(variables): keep dynamic well-known vars live across `declare -a` / array assign

lu-zero:
## Problem

Once a script `declare -a`s or plain-array-assigns a dynamic well-known variable such as `PIPESTATUS`, brush **never updates it again** on later pipelines in that shell — it stays frozen at whatever was last assigned. Real bash keeps the variable live and refreshes it on every pipeline.

Minimal repro:

```bash
declare -a PIPESTATUS=([0]="1")
true | true | true
echo "${PIPESTATUS[@]}"
```

- **bash**: `0 0 0`
- **brush** (current `origin/main`): `1` (frozen)

Same freeze happens with a plain array assignment (no `declare`):

```bash
PIPESTATUS=([0]="9")
true | false | true
echo "${PIPESTATUS[@]}"
```

- **bash**: `0 1 0`
- **brush**: `9`

## Root cause

`PIPESTATUS` (and other dynamic well-known variables: `RANDOM`, `SECONDS`, `FUNCNAME`, `BASH_LINENO`, `BASH_SOURCE`, `BASH_ALIASES`, …) are backed by a `ShellValue::Dynamic { getter, setter }` whose `getter` is consulted on every read and whose `setter` is a no-op (`|_| ()`) for all of them.

Two write paths in `brush-core/src/variables.rs` unconditionally overwrote `self.value` with a static array, discarding the `Dynamic` binding and permanently freezing the variable:

1. **`convert_to_indexed_array`** — the `_` arm hit for `Dynamic` values and replaced `self.value`. Triggered by `declare -a VAR=(...)` via `brush-builtins/src/declare.rs`.
2. **`assign`** (non-append Array-literal arm) — explicitly matched `ShellValue::Dynamic { .. }` in the overwrite set. Triggered by a plain `VAR=([0]="...")` assignment.

From that point on the getter closure was gone, so the variable stopped refreshing for the rest of the shell's life.

## Fix

- `convert_to_indexed_array`: add an explicit `ShellValue::Dynamic { .. } => Ok(())` arm so the binding survives.
- `assign`: drop `ShellValue::Dynamic { .. }` from the overwrite set; the existing `(ShellValue::Dynamic { .. }, _) => Ok(())` arm below now handles it correctly.

### `convert_to_associative_array` intentionally unchanged

`convert_to_associative_array` has the same overwrite pattern, but real bash is asymmetric here: `declare -A PIPESTATUS=([x]="9")` **does** permanently freeze it in bash itself (PIPESTATUS becomes a plain associative array that pipelines don't refresh). Pre-fix brush already matched bash's `-A` behavior, so changing it would diverge from bash compat.

## Verification

All four behavioral cases now match bash:

| op | bash | brush (after) |
|---|---|---|
| `declare -a` then pipe | `0 0 0` | `0 0 0` ✓ |
| `declare -A` then pipe | `9` | `9` ✓ |
| plain `=([0]="9")` then pipe | `0 1 0` | `0 1 0` ✓ |
| plain pipe | `0 0 0` | `0 0 0` ✓ |

Adds three compat test cases under `brush-shell/tests/cases/compat/well_known_vars.yaml` (`declare -a`, plain array assign, `declare -A`). `cargo fmt --check` clean.

This surfaced in the wild: a worker-env dump/restore mechanism restored a `declare -a PIPESTATUS=([0]="1")` line and a downstream `pipestatus || die` check fired incorrectly, misreporting a successful pipeline as failed.

### DIFF
--- a/brush-builtins/src/declare.rs
+++ b/brush-builtins/src/declare.rs
@@ -288,10 +288,18 @@ impl DeclareCommand {
             .get_mut_using_policy(name.as_str(), lookup)
         {
             if self.make_associative_array.is_some() {
-                var.convert_to_associative_array()?;
+                if initial_value.is_some() {
+                    var.convert_to_associative_array_for_reassignment()?;
+                } else {
+                    var.convert_to_associative_array()?;
+                }
             }
             if self.make_indexed_array.is_some() {
-                var.convert_to_indexed_array()?;
+                if initial_value.is_some() {
+                    var.convert_to_indexed_array_for_reassignment()?;
+                } else {
+                    var.convert_to_indexed_array()?;
+                }
             }
 
             self.apply_attributes_before_update(var)?;
--- a/brush-builtins/src/declare.rs
+++ b/brush-builtins/src/declare.rs
@@ -281,6 +281,20 @@ impl DeclareCommand {
             EnvironmentLookup::Anywhere
         };
 
+        // If the variable currently holds a dynamic value, resolve its current
+        // reading now, before taking a mutable borrow of the shell below. This
+        // lets a scalar dynamic (e.g. RANDOM) that ends up getting materialized
+        // by a shape conversion freeze at its actual current value instead of
+        // an empty string.
+        let resolved_dynamic_value = context
+            .shell
+            .env()
+            .get_using_policy(name.as_str(), lookup)
+            .and_then(|var| {
+                matches!(var.value(), brush_core::ShellValue::Dynamic { .. })
+                    .then(|| var.resolve_value(context.shell))
+            });
+
         // Look up the variable.
         if let Some(var) = context
             .shell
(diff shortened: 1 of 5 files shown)


lu-zero:
The **Source code checks (1.88.0, ubuntu)** failure is `cargo deny --all-features check all` flagging two newly-published advisories (`RUSTSEC-2026-0194`, `RUSTSEC-2026-0195`) in the advisory DB. `bans`, `licenses`, and `sources` all pass. This is unrelated to this change (a 2-line `variables.rs` edit + tests) and would fail on `origin/main` too — it is an upstream advisory-DB/infra issue affecting all open PRs.

reubeno:
@lu-zero Thanks for highlighting the errors from the new rustc and from `cargo deny`. I'm working on addressing them in #1240.

PR #1233, opened 2026-07-09
Use `IntoIterator` over `Vec`

Elsie19:
Both of these places only use `into_iter` by itself, not `Vec::into_iter`. It probably will result in basically the same machine code, but this way is technically better.


reubeno (approved):

reubeno:
I’m rebasing your branch now that we’ve resolved the clippy and deny issues that are unrelated.

change otherwise looks good!

(Love the branch name.)

PR #1237, opened 2026-07-11
fix(help): return non-zero exit status when no help topics match

### LINKED ISSUE #1236: Builtin `help x` should return non-zero if `x` don't have docs in `help`.
### brush version

brush 0.4.0 (git:96a26d0)

GNU bash, version 5.1.4(1)-release (x86_64-pc-linux-gnu)

### Platform / OS

WSL

### Steps to reproduce

run: 

~~~ bash
help xxx ; echo $?
~~~

### Actual behavior (brush)

out: 

~~~~
No help topics match 'xxx'
0
~~~~

### Expected behavior

out: 

~~~~
bash: help: no help topics match `xxx'.  Try `help help' or `man -k xxx' or `info xxx'.
1
~~~~

### Additional context

The out of `echo $?` is the point, the exit code is not `0` if no help topics match in bash, but is always `0` in brush.

zbl1998-sdjn:
## Problem

`help nonexistent` prints "No help topics match ..." but exits 0, whereas bash exits 1 when none of the requested topics match. (#1236)

## Root cause

`HelpCommand::execute` always returned `ExecutionResult::success()`, and `display_help_for_topic_pattern` didn't report whether it matched anything.

## Fix

`display_help_for_topic_pattern` now returns whether at least one topic matched, and `execute` returns `general_error()` (exit 1) only when none of the requested patterns matched — mirroring bash, which succeeds as long as at least one requested topic matches:

| stdin | bash 5.2.21 | brush before | brush after |
| --- | --- | --- | --- |
| `help cd` | 0 | 0 | 0 |
| `help nonexistent` | 1 | 0 | 1 |
| `help cd nonexistent` | 0 | 0 | 0 |
| `help nope1 nope2` | 1 | 0 | 1 |
| `help` (no args) | 0 | 0 | 0 |

The stderr message is unchanged; only the exit status is affected.

## Tests

Added two oracle cases to `builtins/help.yaml`: one for an unknown topic (expects exit 1) and one for a known+unknown mix (expects exit 0).

## Verification

`cargo build`, `cargo fmt --all -- --check`, and `cargo clippy --all-targets -- -D warnings` are clean. Also ran `brush -c 'help ...'` directly for each row above and confirmed the exit statuses match bash 5.2.21.

Closes #1236

---

*Assisted-by: Claude (Anthropic) — disclosed here and in the commit trailer per the project's AI policy. I've reviewed and understand the change and vouch for it.*

### DIFF
--- a/brush-builtins/src/help.rs
+++ b/brush-builtins/src/help.rs
@@ -31,13 +31,23 @@ impl builtins::Command for HelpCommand {
     ) -> Result<brush_core::ExecutionResult, Self::Error> {
         if self.topic_patterns.is_empty() {
             Self::display_general_help(&context)?;
-        } else {
-            for topic_pattern in &self.topic_patterns {
-                self.display_help_for_topic_pattern(&context, topic_pattern)?;
+            return Ok(ExecutionResult::success());
+        }
+
+        // Match bash: succeed if at least one requested topic pattern matched
+        // something; return a non-zero exit code only when none of them matched.
+        let mut any_matched = false;
+        for topic_pattern in &self.topic_patterns {
+            if self.display_help_for_topic_pattern(&context, topic_pattern)? {
+                any_matched = true;
             }
         }
 
-        Ok(ExecutionResult::success())
+        if any_matched {
+            Ok(ExecutionResult::success())
+        } else {
+            Ok(ExecutionResult::general_error())
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


reubeno:
Thanks for the contribution, @zbl1998-sdjn! 

The change looks good. Definitely appreciate the added tests.

I've unblocked the PR checks to run and will monitor their results...

reubeno (approved):

zbl1998-sdjn:
"Thanks for the quick review and the merge!"

PR #1244, opened 2026-07-20
feat(core): add opt-in kill-on-drop for externally spawned commands

hartsock:
Found during practical testing. We were experimenting with spawning shells and the outer process would terminate and leave behind the child shell.

`sys::process::spawn` never sets `kill_on_drop`, so a spawned child outlives the shell that started it. That is correct for a real shell — it is what makes job control, disowned jobs, and `nohup`-style usage work — and remains the default.

It is wrong for an embedded shell, where the host creates and destroys the shell as an object. There a surviving child keeps running unattended and holds a duplicate of the shell's stdout/stderr pipe, so a host draining that output never observes EOF.

Adds a `kill_external_commands_on_drop` creation option, defaulting to `false`, plumbed through `RuntimeOptions` to `sys::process::spawn` alongside the existing `external_cmd_leads_session`. The stub backend accepts and ignores it. Process-group behavior is untouched, so this signals the immediate child only.

Tests cover both directions and that the option defaults off. Mutation-checked: forcing `kill_on_drop(false)` fails the enabled case.

### DIFF
--- a/brush-core/src/commands.rs
+++ b/brush-core/src/commands.rs
@@ -633,7 +633,7 @@ pub(crate) fn execute_external_command(
             .join(" ")
     );
 
-    match sys::process::spawn(cmd) {
+    match sys::process::spawn(cmd, context.shell.options().kill_external_commands_on_drop) {
         Ok(child) => {
             // Retrieve the pid.
             #[expect(clippy::cast_possible_wrap)]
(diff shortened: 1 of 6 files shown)


hartsock:
Found alongside #1184 and #1242 during practical testing of brush embedded as a shell library.

reubeno (approved):

reubeno:
Thanks for the contribution! Agreed that this is valuable for embedded shell hosts.

PR #1246, opened 2026-07-23
feat(core): support building for iOS targets

Jackkakaya:
## What

Makes `brush-core` compile (and run) for iOS targets (`aarch64-apple-ios`, `aarch64-apple-ios-sim`). Three small, cfg-gated changes:

1. **`uzers` → non-iOS unix only, new `users_ios` module.** `uzers` fails to compile for iOS, and the iOS app sandbox has no user database to query anyway. The new module answers identity queries from `getuid()`/`getgid()`-family syscalls and the process environment (`$HOME`, `$USER`).
2. **`waitid` → `waitpid` fallback on iOS.** `nix` does not expose `waitid()` on iOS; this reuses the exact fallback path already in place for NetBSD/OpenBSD.
3. **`ioctl(TIOCSCTTY)` cfg widened from `target_os = "macos"` to `target_vendor = "apple"`.** The request argument is `c_ulong` on iOS just like macOS.

## Why

brush-core turns out to be a remarkably good fit for embedding a shell inside an iOS app, where `fork`/`exec` is forbidden: working dir and env are per-`Shell`-instance state rather than process globals, builtins do I/O through the shell's fd table (real pipes) rather than process stdio, and any command name can be routed to an in-process builtin. With just this patch, brush-core powers a working iOS terminal app.

## Testing

- `cargo check` + `cargo clippy` clean for `aarch64-apple-ios`, `aarch64-apple-ios-sim`, and the host (macOS).
- No behavior change on currently supported platforms: everything is cfg-gated to iOS except the ioctl cfg widening, which is value-identical on macOS.
- Ran a 102-case shell battery (language features, redirections/pipes, ~35 builtins, custom in-process commands) on an iOS 17 simulator inside a demo app embedding brush-core: 102/102 pass.

Happy to also add an iOS check target to CI in a follow-up if that's welcome.

https://claude.ai/code/session_011tLwxQ47SiGLGDDPXd8tzz

### DIFF
--- a/brush-core/src/sys/unix/users_ios.rs
+++ b/brush-core/src/sys/unix/users_ios.rs
@@ -1,0 +1,70 @@
+//! iOS: no user database is accessible from within the app sandbox; identity
+//! comes from uid/gid syscalls and the process environment.
+
+use crate::error;
+use std::path::PathBuf;
+
+pub(crate) const fn is_root() -> bool {
+    false
+}
+
+pub(crate) fn get_user_home_dir(username: &str) -> Option<PathBuf> {
+    // Only the current (sandboxed) user is resolvable.
+    if get_current_username().is_ok_and(|current| current == username) {
+        get_current_user_home_dir()
+    } else {
+        None
+    }
+}
+
+pub(crate) fn get_current_user_home_dir() -> Option<PathBuf> {
+    std::env::var_os("HOME").map(PathBuf::from)
+}
+
+pub(crate) const fn get_current_user_default_shell() -> Option<PathBuf> {
+    None
+}
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_current_uid() -> Result<u32, error::Error> {
+    // SAFETY: getuid is always successful and has no preconditions.
+    Ok(unsafe { libc::getuid() })
+}
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_current_gid() -> Result<u32, error::Error> {
+    // SAFETY: getgid is always successful and has no preconditions.
+    Ok(unsafe { libc::getgid() })
+}
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_effective_uid() -> Result<u32, error::Error> {
+    // SAFETY: geteuid is always successful and has no preconditions.
+    Ok(unsafe { libc::geteuid() })
+}
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_effective_gid() -> Result<u32, error::Error> {
+    // SAFETY: getegid is always successful and has no preconditions.
+    Ok(unsafe { libc::getegid() })
+}
+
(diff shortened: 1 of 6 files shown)


reubeno:
Thanks for the contribution, @Jackkakaya! It's great to hear that you were able to make use of `brush-core` on another platform :smile:

I'll need to look a bit more closely at `users_ios.rs`; the other targeted cfg changes look reasonable. I also agree that it would make sense to follow up with some CI-time checks to make sure that things don't regress. I'm open to that either being in this PR or being in a fast-follow subsequent PR.

brush-core/src/sys/unix/users_ios.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_current_username() -> Result<String, error::Error> {
+    Ok(std::env::var("USER").unwrap_or_else(|_| "mobile".to_owned()))
+}
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_user_group_ids() -> Result<Vec<u32>, error::Error> {
+    // SAFETY: getgid is always successful and has no preconditions.
+    Ok(vec![unsafe { libc::getgid() }])
```
reubeno, line 60: While these are straightforward, we prefer using the `nix` crate instead of direct unsafe `libc` calls. Could you please update the `libc` calls in this file to go through it instead?

brush-core/src/sys/unix/users_ios.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+pub(crate) fn get_all_users() -> Result<Vec<String>, error::Error> {
+    Ok(vec![get_current_username()?])
+}
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_all_groups() -> Result<Vec<String>, error::Error> {
+    Ok(vec!["mobile".to_owned()])
```
reubeno, line 69: Did you find a case where it's important to return at least one group instead of just returning an empty vector?

brush-core/src/sys/unix/users_ios.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_effective_gid() -> Result<u32, error::Error> {
+    // SAFETY: getegid is always successful and has no preconditions.
+    Ok(unsafe { libc::getegid() })
+}
+
+#[expect(clippy::unnecessary_wraps)]
+pub(crate) fn get_current_username() -> Result<String, error::Error> {
+    Ok(std::env::var("USER").unwrap_or_else(|_| "mobile".to_owned()))
```
reubeno, line 54: Is `USER` typically set in the iOS targets you've tested with? If so, I'm tempted to avoid a hard-coded fake fallback.

reubeno (changes requested):
Changes generally looking good -- just a few requests/questions.

PR #1252, opened 2026-07-27
fix(commands): resolve relative external command paths against shell cwd

### LINKED ISSUE #1251: Bug: [Windows] relative external command paths fail after `cd` because the executable path is not resolved against the shell working directory
### brush version

brush 0.4.0 (git:897b373)

_No response_

### Platform / OS

Windows

### 1. Build brush

From the brush repository root:

```bash
cargo build
cargo build --release
```

### 2. Start brush from the repository root

```bash
target/debug/brush.exe
```

At this point, the shell working directory and the actual process working directory are both the repository root.

### 3. Change directory inside brush

Change into any subdirectory of the directory where brush was started. For example:

```bash
cd target
```

The important condition is that the shell working directory is no longer the same as brush’s OS process working directory.

### 4. Try to execute a relative external command

Inside brush:

```bash
./release/brush.exe --version
```

### Actual behavior (brush)

brush reports:

```text
error: command not found: ./release/brush.exe
```

The same failure occurs with:

```bash
.\release\brush.exe --version
release/brush.exe --version
```

Absolute paths work:

```bash
C:/full/path/to/target/release/brush.exe --version
```

### Expected behavior

The command should execute successfully and print the brush version, for example:

```text
brush <version>
```

Equivalent behavior in other shells:

- POSIX shells resolve `./release/brush.exe` against the current shell working directory.
- Windows `cmd.exe` and PowerShell can execute relative paths like `.\release\brush.exe` from the current directory.

## Summary

On Windows, brush does not correctly execute external commands specified with a relative path when the shell’s current working directory differs from brush’s actual OS process working directory.

This most commonly happens after using `cd` to move into a subdirectory of the directory where brush was started. For example, if brush is started from the repository root, then the user changes into `target/` and tries to run:

```bash
./release/brush.exe --version
```

or:

```bash
./debug/brush.exe --version
```

the command fails with:

```text
error: command not found: ./release/brush.exe
```

even though the executable exists relative to the shell’s current working directory.

This is not specific to `brush.exe`, `release/`, or `debug/`. Any external command invoked through a relative path can fail once the shell working directory has diverged from brush’s OS process working directory.

The underlying spawn failure is Windows error code `3`:

```text
Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }
```

which corresponds to `ERROR_PATH_NOT_FOUND`.

Absolute paths work correctly:

```bash
C:/full/path/to/brush/target/release/brush.exe --version
```

Commands resolved through `PATH` also work, because brush resolves them to absolute paths before spawning.

The issue appears to be that brush passes the relative executable path directly to `std::process::Command`, while setting the child working directory via:

```rust
cmd.current_dir(context.shell.working_dir());
```

On Windows, this does not appear to be sufficient: the relative executable path is resolved against brush’s own OS process current directory, not against the child working directory specified by `Command::current_dir(...)`.

Since brush updates its internal shell working directory on `cd`, but does not necessarily update the actual OS process current directory, relative external command paths are resolved against the wrong directory.

---

## Reliable reproduction?

Yes. This reproduces reliably on my machine.

It appears to happen whenever:

1. brush’s internal shell working directory has been changed, for example by `cd`, and
2. the user executes an external command using a relative path containing a path separator.

Commands found through `PATH` are unaffected because brush resolves them to absolute paths before spawning.

---

DaZuo0122:
Fixes #1251.

On Windows, relative external command paths were resolved against the
brush process's OS current directory instead of the shell's working
directory. This caused commands such as:

```bash
cd target
./release/brush.exe --version
```

to fail with `ERROR_PATH_NOT_FOUND`, even though the executable existed
relative to the shell's current working directory.

## Fix

Resolve relative executable paths against `context.shell.working_dir()`
before composing the `std::process::Command`.

Absolute paths and `PATH`-resolved commands continue to work.

p.s. I have added `#[allow(clippy::too_many_lines)]` to `execute_external_command` to pass the `cargo xtask ci pre-commit` check.


reubeno:
@DaZuo0122 Thanks for the contribution (and for finding the issue)!

I'd like to find a way to resolve this without making any meaningful behavioral changes to non-Windows code paths. (We can also reuse the path normalization helper crate that we use elsewhere in the project -- that may help to simplify.)

Are you open to me collaborating with you on this and refactoring it a bit?

DaZuo0122:
@reubeno  Absolutely, and yes, I’m happy to collaborate on this!

I agree that the fix should avoid meaningful behavioral changes on non-Windows platforms. I can rework it so the change is Windows-specific, likely gated behind `#[cfg(windows)]` or placed in the platform-specific layer.

Feel free to let me know what direction you prefer and I’ll update it.

brush-core/src/commands.rs
```diff
@@ -584,6 +585,39 @@ pub(crate) fn execute_external_command(
         .try_fd(openfiles::OpenFiles::STDIN_FD)
         .is_some_and(|f| f.is_terminal());
 
+    // On Windows, normalize relative executable paths before spawning.
+    #[cfg(windows)]
+    let normalized_executable = if Path::new(executable_path).is_relative() {
+        let joined = context.shell.working_dir().join(executable_path);
+
+        let mut normalized = PathBuf::new();
```
reubeno, line 592: request(blocking): We already use the `normalize_path` crate elsewhere in this project. Could you look at using that instead of hand-normalizing?

brush-core/src/commands.rs
```diff
@@ -584,6 +585,39 @@ pub(crate) fn execute_external_command(
         .try_fd(openfiles::OpenFiles::STDIN_FD)
         .is_some_and(|f| f.is_terminal());
 
+    // On Windows, normalize relative executable paths before spawning.
+    #[cfg(windows)]
```
reubeno, line 588: request(blocking): While we have some still sprinkled around, our strong preference is for platform-specific code to be invoked via the `sys` module in the crate. For non-Windows platforms, it should ideally be a no-op/pass-through that doesn't impact the code path; for Windows, it can perform this additional manipulation.

Don't hesitate to holler or ask questions if you could use additional pointers/examples on how best to do this.

reubeno (changes requested):
Thanks for being willing to iterate on this! I've got 2 main comments here.

It would also be ideal if you can find a way to test any non-trivial logic that gets added to the Windows code path.

PR #1253, opened 2026-07-30
fix(parser): recognize a comment after an odd number of blanks

bresilla:
A comment inside `$( … )` is only recognized when the number of blanks before its `#` is even.
Otherwise the comment text gets tokenized as shell, and an unbalanced quote in it never closes.

Smallest repro I could get it down to is 7 bytes. bash accepts it:

```sh
$( #'
)
```

```
unterminated single quote at 1,5 (detected near line 3 col 1)
```

Zero blanks work. One doesn't. Two work, three doesn't, and so on. Tabs count the same as spaces.
Only `$( … )` is affected; the same comment at top level or inside backticks is fine.

One trap if you go poking at this: the comment needs an odd number of quotes to show the failure.
`# it's the shell's` has two apostrophes and they pair up harmlessly, so it parses. That cost me a
test I thought was passing.

### Cause

`consume_nested_construct` tokenizes with `include_space: true`. In that mode a blank appends to
the token when none is started, and delimits when one is, so the blanks alternate between the two.
After an odd number, a token is in progress. The `#` arm sits after the "token in progress" arm,
so it never runs and the `#` gets appended instead.

### Fix

The blanks `include_space` accumulates are there to reproduce the construct's original text. They
aren't a word, so a `#` after nothing but those blanks should still start a comment. That's one
clause on the existing condition plus a small helper.

I narrowed the guard rather than moving the `#` arm earlier, since `started_token()` is doing real
work there: `a#b` is one word.

### Testing

New test covers eight blank prefixes. `cargo test -p brush-parser`: 225 pass. clippy and fmt clean.

I ran `brush-compat-tests` on `main` and on the branch and got identical numbers both times (1730
pass, 65 fail, 379 known-fail, 28 skipped), so those 65 are pre-existing and none of them moved.

I hit this building a shell on `brush-parser`. Two files in modernish's test suite trip it, both
`$(`, newline, one tab, then a comment with an apostrophe in it.

### DIFF
--- a/brush-parser/src/tokenizer.rs
+++ b/brush-parser/src/tokenizer.rs
@@ -294,6 +294,15 @@ impl TokenParseState {
         !self.token_so_far.is_empty()
     }
 
+    /// Returns true if the token so far consists only of blanks.
+    ///
+    /// This can only happen when tokenizing with `include_space`, where blanks are accumulated
+    /// into the token so that the original text of a nested construct can be reproduced. Such
+    /// blanks are not a word, so a `#` following them still begins a comment.
+    pub fn only_blanks_so_far(&self) -> bool {
+        !self.token_so_far.is_empty() && self.token_so_far.chars().all(is_blank)
+    }
+
     pub fn append_char(&mut self, c: char) {
         self.token_so_far.push(c);
     }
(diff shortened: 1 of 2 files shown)


reubeno:
@bresilla  -- thanks for the contribution! The change looks righteous.

Would you please also add one or more YAML based compat test case that captures the end-to-end compatibility issue underlying this? There should be docs and examples that you can follow.

reubeno:
@bresilla -- thanks again for getting this started. I'd like to get this moving forward, so I've taken the liberty to add a few test cases and rebase against latest changes in `main`. Assuming PR checks pass, we'll get this merged.

PR #1259, opened 2026-08-05
fix(read): skip terminal settings for non-terminal input

Jpei1994:
## Problem

On Windows, `read` fails with `os error 87` (ERROR_INVALID_PARAMETER) whenever the input is a pipe, file, or redirect AND the host process's standard handles are a console. This is the common case for an agent shell hosted in a terminal (e.g. a coding agent running `printf 'x\n' | while read -r line; ...`), where the shell runs in-process in the TUI host.

## Root cause

`setup_terminal_settings` always calls `AutoModeGuard::new(file)` and applies settings. `AutoModeGuard::new`'s `from_term` keys off the process's standard handles (`GetStdHandle`) rather than the `file` argument — so in a terminal-hosted process it succeeds even when the read input is a pipe/file/redirect. `apply_settings` then calls `SetConsoleMode` on the non-console input, which fails with ERROR_INVALID_PARAMETER, and the `?` propagates the failure, failing the whole read.

Subprocesses never hit this because their stdin is a pipe, so `AutoModeGuard::new` fails and the `.ok()` degrades to no-op.

## Fix

Skip terminal-mode setup entirely for non-terminal input (`if !file.is_terminal() { return Ok(None); }`), matching bash's termios behavior (bash only applies terminal settings to terminal input).

## Tests

Two contract tests: pipe input and file input both return `Ok(None)` (no mode guard installed, no error).

Verified on Windows: after this change `printf 'ok\n' | while IFS= read -r x; do echo "$x"; done` works in the terminal-hosted agent shell.

### DIFF
--- a/brush-builtins/src/read.rs
+++ b/brush-builtins/src/read.rs
@@ -566,6 +566,17 @@ impl ReadCommand {
         &self,
         file: &brush_core::openfiles::OpenFile,
     ) -> Result<Option<brush_core::terminal::AutoModeGuard>, brush_core::Error> {
+        // Terminal settings only make sense for terminal input. `AutoModeGuard::new`
+        // keys off the process's standard handles, which may be a console even when
+        // the read input is a pipe/file/redirect (e.g. an agent shell hosted in a
+        // terminal): in that case `apply_settings` would hit `SetConsoleMode` on the
+        // non-console input and fail with ERROR_INVALID_PARAMETER (os error 87),
+        // failing the read entirely. Skip terminal setup for non-terminal input,
+        // matching bash's termios behavior.
+        if !file.is_terminal() {
+            return Ok(None);
+        }
+
         let mode = brush_core::terminal::AutoModeGuard::new(file.to_owned()).ok();
         if let Some(mode) = &mode {
             let config = brush_core::terminal::Settings::builder()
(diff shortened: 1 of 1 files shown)


brush-builtins/src/read.rs
```diff
@@ -691,6 +702,48 @@ mod tests {
...
+        assert!(result.is_ok(), "non-terminal input must not error");
+        assert!(
+            result.unwrap().is_none(),
+            "non-terminal input must not install a mode guard"
+        );
+    }
+
+    #[test]
+    fn test_setup_terminal_settings_skips_file_input() {
+        let mut tmp = std::env::temp_dir();
```
reubeno, line 729: Instead of manually creating a file in the temp dir and cleaning it up, could you please use [tempfile::tempfile](https://docs.rs/tempfile/latest/tempfile/fn.tempfile.html)? We use that crate elsewhere in testing to simplify auto-cleanup.

brush-builtins/src/read.rs
```diff
@@ -691,6 +702,48 @@ mod tests {
 
     use super::*;
 
+    // ==================== setup_terminal_settings tests ====================
```
reubeno, line 705: Thanks for adding these tests!

brush-builtins/src/read.rs
```diff
@@ -566,6 +566,17 @@ impl ReadCommand {
         &self,
         file: &brush_core::openfiles::OpenFile,
     ) -> Result<Option<brush_core::terminal::AutoModeGuard>, brush_core::Error> {
+        // Terminal settings only make sense for terminal input. `AutoModeGuard::new`
+        // keys off the process's standard handles, which may be a console even when
+        // the read input is a pipe/file/redirect (e.g. an agent shell hosted in a
+        // terminal): in that case `apply_settings` would hit `SetConsoleMode` on the
```
reubeno, line 572: Can you clarify where the call to `SetConsoleMode` happens -- or what the caller is? I don't see any direct calls to it in `brush`, nor do we have any Windows-specific console paths right now. (Perhaps I've overlooked something, though?)

brush-builtins/src/read.rs
```diff
@@ -566,6 +566,17 @@ impl ReadCommand {
         &self,
         file: &brush_core::openfiles::OpenFile,
     ) -> Result<Option<brush_core::terminal::AutoModeGuard>, brush_core::Error> {
+        // Terminal settings only make sense for terminal input. `AutoModeGuard::new`
```
reubeno, line 569: Meta-note: this comment is Windows-centric; could you simplify it and clarify which parts are platform specific?

reubeno (changes requested):
Thanks for putting this together! I've got a few questions/requests. My biggest question is what the actual call stack of the failing API looks like so I can better understand whether this is the best place for the check.

PR #1267, opened 2026-08-15
refactor: remove cached macro usages from parsers

lu-zero:
## Summary

Drop the `#[cached(...)]` attribute invocations in word, tokenizer, arithmetic, prompt, and shell parsing. Where a public function delegated to a private cached wrapper, the wrapper is inlined into the public function. The now-unused `cached` dependency is removed from `brush-parser`; `brush-core` keeps it because `brush-core/src/regex.rs` still uses `cached::LruCache` directly (not the macro).

## Motivation

This is an initial experiment to establish a baseline for benchmarking. It removes the `cached` crate macro usage so we can measure the cost/benefit of the caching before deciding how to replace or tune it.

## Changes

- `brush-parser/src/word.rs` — inline `cacheable_parse` into `parse`
- `brush-parser/src/tokenizer.rs` — drop the `uncached_tokenize_string` wrapper; `tokenize_str_with_options` calls `uncached_tokenize_str` directly
- `brush-parser/src/arithmetic.rs` — inline `cacheable_parse` into `parse`
- `brush-core/src/prompt.rs` — drop attribute on `parse_prompt`
- `brush-core/src/shell/parsing.rs` — drop attribute on `parse_string_impl`
- `brush-parser/Cargo.toml` + `Cargo.lock` — remove `cached` dependency

## Testing

- `cargo clippy --package brush-parser --package brush-core` passes
- `cargo test --package brush-parser --package brush-core` passes

Assisted-by: opencode:deepseek-v4-flash-free

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -593,7 +593,6 @@ dependencies = [
  "anyhow",
  "arbitrary",
  "bon",
- "cached",
  "criterion",
  "getrandom 0.4.3",
  "indenter",
--- a/brush-core/src/prompt.rs
+++ b/brush-core/src/prompt.rs
@@ -58,7 +58,6 @@ pub(crate) async fn expand_prompt(
     Ok(formatted_prompt)
 }
 
-#[cached::macros::cached(max_size = 64, key = "String", convert = r#"{ spec.to_owned() }"#)]
 fn parse_prompt(
     spec: &str,
 ) -> Result<Vec<brush_parser::prompt::PromptPiece>, brush_parser::WordParseError> {
--- a/brush-core/src/shell/parsing.rs
+++ b/brush-core/src/shell/parsing.rs
@@ -45,11 +45,6 @@ impl<SE: extensions::ShellExtensions> Shell<SE> {
     }
 }
 
-#[cached::macros::cached(
-    max_size = 64,
-    key = "(String, brush_parser::ParserOptions)",
-    convert = r#"{ (s.to_owned(), parser_options.to_owned()) }"#
-)]
 fn parse_string_impl(
     s: &str,
     parser_options: &brush_parser::ParserOptions,
--- a/brush-parser/Cargo.toml
+++ b/brush-parser/Cargo.toml
@@ -27,7 +27,6 @@ winnow-parser = ["dep:winnow"]
 [dependencies]
 arbitrary = { version = "1.4.2", optional = true, features = ["derive"] }
 bon = "3.9.1"
-cached = "2.0.2"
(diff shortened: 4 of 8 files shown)


lu-zero:
## Benchmark comparison: cached (main) vs uncached (this branch)

I ran the benchmarks locally on both `main` (with `cached` crate) and this branch (without). Besides the existing micro-benchmarks, I added two realistic mixed-command benchmarks (commit a1330ef7) and compared the medians.

Delta column = change in the **uncached** result relative to cached. Positive = uncached is slower; negative = uncached is faster.

### Existing micro-benchmarks (repeat the *same* command every iteration — cache's best case)

| Benchmark | Cached | Uncached | Delta (uncached vs cached) |
|---|---|---|---|
| `eval_arithmetic` | 150 ns | 981 ns | +6.5x (slower) |
| `expand_one_string` | 1.85 µs | 3.95 µs | +2.1x (slower) |
| `for_loop` | 31.6 µs | 61.2 µs | +1.9x (slower) |
| `run_echo_builtin_command` | 12.4 µs | 14.7 µs | +19% |
| `function_call` | 3.79 µs | 4.78 µs | +26% |
| `clone_shell_object` | 14.4 µs | 15.5 µs | +7.2% |
| `instantiate_shell` | 38.9 µs | 38.9 µs | ~0% |

### New mixed-command benchmarks (added in a1330ef7)

| Benchmark | Cached | Uncached | Delta (uncached vs cached) |
|---|---|---|---|
| `run_mixed_script` (same mixed script every iteration) | 617.7 µs | 747.8 µs | +21% (slower) |
| `run_mixed_commands` (128 distinct commands, exceeds 64-entry cache) | 1.439 ms | 1.388 ms | −3.5% (faster) |

### Interpretation

- When the same input is re-parsed (interactive repeat, tight loops), the cache pays off significantly (~2x-6x in the micro-benchmarks, ~20% for a repeated script).
- For a pool of distinct commands larger than the 64-entry LruCache, the cache mostly misses and its overhead (key `String` clones + hashing) makes the cached version *slower* (~3.5%). This is the realistic case for running varied real-world scripts.

Net takeaway: the cache is a win for repeated-identical-input workloads but roughly a wash-to-slight-loss for varied scripts. This gives us a baseline to decide whether to keep, replace (e.g. a simpler hashmap, larger capacity), or drop the caches.

Note: `brush-core` still depends on `cached` for the direct `LruCache` usage in `regex.rs`; only the macro-based caches were removed here.

Assisted-by: opencode:deepseek-v4-flash-free

PR #1271, opened 2026-08-16
fix(parser): decline out-of-range numeric values instead of unwrapping

ferentinai:
Two grammar rules match a run of ASCII digits and then unwrap the integer parse,
so a value too large for the target type panics the parser instead of failing to
match:

```console
$ echo 99999999999999999999>&1
thread 'main' panicked at brush-parser/src/parser/peg.rs:703:27:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: PosOverflow }

$ echo ~99999999999999999999
thread 'main' panicked at brush-parser/src/word.rs:913:115:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: PosOverflow }
```

The first is `io_number`, which parses into an `i32`. The second is the tilde
directory-stack index, which parses into a `usize`, in both the `~+N` and `~-N`
forms.

The digit-run guard is what makes these look safe: every character is
`is_ascii_digit`, so the parse cannot fail on shape. It can still fail on range,
and that is the case neither rule handles.

## The change

Both become fallible peg actions, so the rule declines and the input is
reconsidered by later alternatives instead of taking the process down.

The resulting parses match what bash prints:

| Input | bash | after this change |
|---|---|---|
| `~99999999999999999999` | prints literally | tilde expansion of a (non-existent) user name |
| `~-99999999999999999999` | prints literally | tilde expansion of a user name |
| `~+99999999999999999999` | prints literally | plain text, since `+` is not a valid user-name character |

## One deliberate divergence, worth your call

For the redirection case bash accepts the number at parse time and fails when the
redirection is performed:

```console
$ echo 99999999999999999999>&1
bash: file descriptor out of range: Bad file descriptor
```

Declining the rule instead means the digits parse as an ordinary word argument,
so `echo` prints them. That keeps the change local to the parser. Matching bash
properly would mean representing an out-of-range descriptor in the AST and
diagnosing it at redirection time, which means widening `IoFd` and touching the
executor.

I went with the local fix because the goal here was removing the panic, but I am
happy to do it the other way if you would prefer the AST stay faithful.

## Note on the crate's lint configuration

`brush-parser/Cargo.toml` sets `clippy::unwrap_used = "deny"`, which would have
caught both of these, but `lib.rs` disables it crate-wide:

```rust
// TODO(unwrap): remove or scope this allow attribute
#![allow(clippy::unwrap_used)]
```

Not part of this PR, but retiring that allow (or narrowing it to the specific
sites that need it) would stop this class recurring. There are other `unwrap`
calls behind it that I have not audited.

## Testing

- `cargo test -p brush-parser`: 228 passed, 0 failed. That is the existing 224
  plus 4 added here, covering both out-of-range forms and the in-range forms
  still being recognized.
- `cargo clippy -p brush-parser --all-targets`: clean.
- `cargo test -p brush-shell --test brush-compat-tests`: identical to a pristine
  checkout on the same machine.
- The tilde case was found by a structured fuzzer over shell metacharacters, and
  only after the `io_number` panic was fixed, because that one was firing first
  and masking it. 500,000 iterations report no panics with both fixed.

Assisted-by: Claude Opus 5 (Claude Code)

### DIFF
--- a/brush-parser/src/parser/peg.rs
+++ b/brush-parser/src/parser/peg.rs
@@ -707,9 +707,13 @@ peg::parser! {
             [Token::Word(w, num_loc) if w.chars().all(|c: char| c.is_ascii_digit())]
             &([Token::Operator(o, redir_loc) if
                     o.starts_with(['<', '>']) &&
-                    locations_are_contiguous(num_loc, redir_loc)]) {
+                    locations_are_contiguous(num_loc, redir_loc)]) {?
 
-                w.parse().unwrap()
+                // A run of ASCII digits still need not fit in an `IoFd`; e.g.
+                // `echo 99999999999999999999>&1` overflows. Decline the rule
+                // instead of unwrapping, so the token is reconsidered as an
+                // ordinary word rather than panicking the parser.
+                w.parse().map_err(|_| "io number out of range")
             }
 
         //
--- a/brush-parser/src/word.rs
+++ b/brush-parser/src/word.rs
@@ -1011,9 +1011,9 @@ peg::parser! {
             // N.B. A run of digits need not fit in a `usize`; decline the rule
             // when it does not, so the word falls through to being treated as a
             // (non-existent) user name instead of panicking the parser.
-            plus:("+"?) n:$(['0'..='9']*) &tilde_terminator() {? n.parse().map(|n| TildeExpr::NthDirFromTopOfDirStack { n, plus_used: plus.is_some() }).map_err(|_| "dir stack index out of range") } /
+            plus:("+"?) n:$(['0'..='9']*) &tilde_terminator() {? n.parse().map_or_else(|_| Err("dir stack index out of range"), |n| Ok(TildeExpr::NthDirFromTopOfDirStack { n, plus_used: plus.is_some() })) } /
             "-" &tilde_terminator() { TildeExpr::OldWorkingDir } /
-            "-" n:$(['0'..='9']*) &tilde_terminator() {? n.parse().map(|n| TildeExpr::NthDirFromBottomOfDirStack { n }).map_err(|_| "dir stack index out of range") } /
+            "-" n:$(['0'..='9']*) &tilde_terminator() {? n.parse().map_or_else(|_| Err("dir stack index out of range"), |n| Ok(TildeExpr::NthDirFromBottomOfDirStack { n })) } /
             user:$(portable_filename_char()*) &tilde_terminator() { TildeExpr::UserHome(user.to_owned()) }
 
         rule tilde_terminator() = ['/' | ':' | ';' | '}'] / ![_]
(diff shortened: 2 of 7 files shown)


reubeno:
Thanks for identifying the possible panic issues! I'll take a look at the changes.

brush-parser/src/word.rs
```diff
@@ -910,9 +910,12 @@ peg::parser! {
         rule tilde_expression() -> TildeExpr =
             &tilde_terminator() { TildeExpr::Home } /
             "+" &tilde_terminator() { TildeExpr::WorkingDir } /
-            plus:("+"?) n:$(['0'..='9']*) &tilde_terminator() { TildeExpr::NthDirFromTopOfDirStack { n: n.parse().unwrap(), plus_used: plus.is_some() } } /
+            // N.B. A run of digits need not fit in a `usize`; decline the rule
+            // when it does not, so the word falls through to being treated as a
+            // (non-existent) user name instead of panicking the parser.
+            plus:("+"?) n:$(['0'..='9']*) &tilde_terminator() {? n.parse().map(|n| TildeExpr::NthDirFromTopOfDirStack { n, plus_used: plus.is_some() }).map_err(|_| "dir stack index out of range") } /
```
reubeno: suggestion: use `.map_or_else()` here to slightly simplify the code?

brush-parser/src/parser/peg.rs
```diff
@@ -698,9 +698,13 @@ peg::parser! {
             [Token::Word(w, num_loc) if w.chars().all(|c: char| c.is_ascii_digit())]
             &([Token::Operator(o, redir_loc) if
                     o.starts_with(['<', '>']) &&
-                    locations_are_contiguous(num_loc, redir_loc)]) {
+                    locations_are_contiguous(num_loc, redir_loc)]) {?
 
-                w.parse().unwrap()
+                // A run of ASCII digits still need not fit in an `IoFd`; e.g.
+                // `echo 99999999999999999999>&1` overflows. Decline the rule
```
reubeno, line 713: Assuming I understand your explanation correctly, declining the rule would mean that the word shows up as a plain-old argument to the simple command.

Interestingly, that matches what `bash` does with something like:

```
echo hi 299999999999999999999999999999999999999999999999999999>/tmp/file.txt
```

...but doesn't match what it does with:

```
echo hi 299999999999999999999999999999999999999999999999999999>&20000000000000000000000000000000000000000000
```

For this case -- and others -- could you please also add YAML-based compatibility test cases to this change? (See other examples in the docs and repo for examples.)

For those cases where we now match `bash`'s behavior, the tests will prevent that behavior from regressing in the future. For the cases where we intentionally diverge (see your PR summary) we can mark the test as a known failure and a comment that explains why the difference is present and accepted (at least for now).

reubeno (changes requested):
Thanks for catching this! I'm supportive of the direction you took with the fixes -- I do think the behavior is much better than panicking :smile:.

I've left a few comments/requests, but the change is generally looking good.

reubeno:
FYI -- I've taken the liberty of pushing some changes to this PR branch to keep it moving along.

reubeno (approved):

PR #1274, opened 2026-08-17
fix(parser): accept [[ ... ]] as a brace-less function body

vivo75:
## Summary

bash's function-definition grammar allows the function body to be any
compound command, not just a brace group or subshell — including
`[[ ... ]]`, e.g. `is_thing() [[ -n $1 ]]`. brush's PEG grammar modeled
`[[ ... ]]` only as a top-level `Command::ExtendedTest` variant, disjoint
from `CompoundCommand`, so `function_body` (which only accepts
`compound_command`) rejected it outright with a syntax error — even for
the simplest case (`bash -c 'foo() [[ 1 == 1 ]]; foo'` succeeds in real
bash, fails in brush).

Real-world impact: I ran into this while experimenting with brush as an
embedded shell for a Gentoo Portage-adjacent project. Portage's
`bin/eapi.sh` defines ~60 predicate functions this exact way, e.g.:

```sh
___eapi_has_pkg_pretend() [[ ${1-${EAPI-0}} != [0-3] ]]
```

and that file is unconditionally sourced by `bin/isolated-functions.sh`,
so this single construct currently blocks brush from parsing essentially
any real ebuild/eclass pipeline from the get-go.

## Fix

- Add `CompoundCommand::ExtendedTest(ExtendedTestExprCommand)`.
- Teach `function_body` (only) to accept it, deliberately kept out of
  the shared `compound_command` rule to avoid a second, redundant path
  to parsing a bare `[[ ... ]]` as a command (which is already handled
  earlier in `command()` via `Command::ExtendedTest`).
- Handle the new `CompoundCommand::ExtendedTest` variant in the
  interpreter (`ast::CompoundCommand::execute`), reusing the existing
  `extendedtests::eval_extended_test_expr` logic already used for the
  top-level case.

I verified this against real, unmodified Portage source
(`bin/eapi.sh`, `bin/isolated-functions.sh`, `bin/phase-helpers.sh`):
all three now source cleanly under an embedded `brush_core::Shell`,
where previously `eapi.sh` failed to parse at all (line 7).

## Test plan

- [x] New snapshot tests: `parse_function_with_extended_test_body`,
      `parse_function_with_extended_test_body_and_redirect`
- [x] `cargo test -p brush-parser -p brush-core`: all passing, no
      regressions
- [x] `cargo fmt --check` / `cargo clippy --all-targets -- -D warnings`:
      clean
- [x] Manual repro against real `bin/eapi.sh` from the Gentoo Portage
      tree (previously: `syntax error at line 7 col 52`; now: sources
      without error)

### DIFF
--- a/brush-parser/src/ast.rs
+++ b/brush-parser/src/ast.rs
@@ -488,7 +488,7 @@ impl Display for CompoundCommand {
                 write!(f, "{coproc_clause_command}")
             }
             Self::ExtendedTest(extended_test_expr_command) => {
-                write!(f, "{extended_test_expr_command}")
+                write!(f, "[[ {extended_test_expr_command} ]]")
             }
         }
     }
--- a/brush-parser/src/ast.rs
+++ b/brush-parser/src/ast.rs
@@ -448,6 +448,8 @@ pub enum CompoundCommand {
     UntilClause(WhileOrUntilClauseCommand),
     /// A coprocess, which runs a command asynchronously in a subshell.
     Coprocess(CoprocessCommand),
+    /// An extended test command, evaluating an extended test expression.
+    ExtendedTest(ExtendedTestExprCommand),
 }
 
 impl Node for CompoundCommand {}
@@ -465,6 +467,7 @@ impl SourceLocation for CompoundCommand {
             Self::WhileClause(w) => w.location(),
             Self::UntilClause(u) => u.location(),
             Self::Coprocess(c) => c.location(),
+            Self::ExtendedTest(e) => e.location(),
         }
     }
 }
@@ -493,6 +496,9 @@ impl Display for CompoundCommand {
             }
             Self::Coprocess(coproc_clause_command) => {
                 write!(f, "{coproc_clause_command}")
+            }
+            Self::ExtendedTest(extended_test_expr_command) => {
+                write!(f, "{extended_test_expr_command}")
             }
         }
     }
--- a/brush-parser/src/ast.rs
+++ b/brush-parser/src/ast.rs
@@ -374,8 +374,6 @@ pub enum Command {
     Compound(CompoundCommand, Option<RedirectList>),
     /// A command whose side effect is to define a shell function.
     Function(FunctionDefinition),
-    /// A command that evaluates an extended test expression.
-    ExtendedTest(ExtendedTestExprCommand, Option<RedirectList>),
 }
 
 impl Node for Command {}
@@ -391,7 +389,6 @@ impl SourceLocation for Command {
                 }
             }
             Self::Function(f) => f.location(),
-            Self::ExtendedTest(e, _) => e.location(),
         }
     }
 }
@@ -408,13 +405,6 @@ impl Display for Command {
                 Ok(())
             }
             Self::Function(function_definition) => write!(f, "{function_definition}"),
-            Self::ExtendedTest(extended_test_expr, redirect_list) => {
-                write!(f, "[[ {extended_test_expr} ]]")?;
-                if let Some(redirect_list) = redirect_list {
-                    write!(f, "{redirect_list}")?;
-                }
-                Ok(())
-            }
         }
     }
 }
--- a/brush-parser/src/parser/peg.rs
+++ b/brush-parser/src/parser/peg.rs
@@ -454,8 +454,15 @@ peg::parser! {
         pub(crate) rule function_parens_and_body() -> ast::FunctionBody =
             specific_operator("(") specific_operator(")") linebreak() body:function_body() { body }
 
+        // N.B. A function body must be a compound command per POSIX grammar, but bash treats
+        // `[[ ... ]]` (itself a compound command, despite not being part of the shared
+        // `compound_command` rule above to avoid re-deriving `[[`-as-a-bare-command via two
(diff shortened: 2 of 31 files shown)


reubeno:
@vivo75 Thanks for identifying this issue and contributing a fix! I've unblocked the PR checks and will take a look at the changes.

vivo75:
thanks, please keep in mind this is claude code output, not mine and I'm
not a rust programmer!

Il giorno mar 18 ago 2026 alle ore 17:18 reuben olinsky <
***@***.***> ha scritto:

brush-parser/src/parser/peg.rs
```diff
@@ -454,8 +454,15 @@ peg::parser! {
         pub(crate) rule function_parens_and_body() -> ast::FunctionBody =
             specific_operator("(") specific_operator(")") linebreak() body:function_body() { body }
 
+        // N.B. A function body must be a compound command per POSIX grammar, but bash treats
```
reubeno: This makes it sound like we should make a deeper change and properly treat `[[ ... ]]` as a compound command -- i.e., *only* have it show up in the CompoundCommand part of the AST, have `compound_command()` handle it (and leave this rule alone), etc.

Looking at the bash reference manual, I see `[[ ... ]]` constructs are indeed categorized as compound commands, just like `(( ... ))` commands.

Could you look at making that ^^^ change and more fully treating these commands as compound?

brush-parser/src/parser/tests/functions.rs
```diff
@@ -97,3 +97,28 @@ fn parse_function_with_local_vars() -> Result<()> {
     });
     Ok(())
 }
+
+#[test]
+fn parse_function_with_extended_test_body() -> Result<()> {
```
reubeno, line 102: Thanks for adding these tests!

Will you please also add 1 or more YAML-based compat test cases to cover these as well? (There should be documentation in the project on them.)

These tests you've added are great for us verifying expected AST outputs of input; the YAML test cases are a great complement, comparing `brush`'s behavior against `bash` as an oracle.

reubeno (changes requested):

vivo75:
Pushed a follow-up commit addressing both review comments:

- `[[ ... ]]` is now modeled purely as `CompoundCommand::ExtendedTest`, parsed by the shared `compound_command()` rule (same pattern as `(( ... ))`'s `Arithmetic` variant). Removed the disjoint `Command::ExtendedTest` variant/parse path entirely; `function_body()` no longer needs a special-cased second alternative. As a side effect, `[[ ... ]] |&` now works like other compound commands instead of being explicitly rejected.
- Added YAML-based compat test cases in `functions.yaml` comparing brush against bash as an oracle, for a function with an extended-test body with and without a redirect.

Updated the 24 existing parser snapshot tests to reflect that top-level `[[ ... ]]` is now wrapped in `Compound(...)`, matching how `(( ... ))` is already represented. Full workspace build/clippy/fmt and the parser/core/compat test suites pass locally.

reubeno:
Thanks for the quick update! It looks cleaner now, and I appreciate the added tests.

I've unblocked the PR checks to run and will take a closer look at the changes later.

brush-parser/src/parser/peg.rs
```diff
@@ -454,6 +454,9 @@ peg::parser! {
         pub(crate) rule function_parens_and_body() -> ast::FunctionBody =
             specific_operator("(") specific_operator(")") linebreak() body:function_body() { body }
 
+        // N.B. A function body must be a compound command per POSIX grammar; bash additionally
+        // accepts `[[ ... ]]` here (e.g. `is_thing() [[ -n $1 ]]`), which is covered since
```
reubeno: issue(blocking): With this PR and `[[ ... ]]` properly being considered a compound command (and the latest updates you've made), I don't think this comment will be too helpful to keep around -- at least not the part after the semicolon.

brush-parser/src/ast.rs
```diff
@@ -494,6 +487,9 @@ impl Display for CompoundCommand {
             Self::Coprocess(coproc_clause_command) => {
                 write!(f, "{coproc_clause_command}")
             }
+            Self::ExtendedTest(extended_test_expr_command) => {
+                write!(f, "{extended_test_expr_command}")
```
reubeno: issue(blocking): Copilot is right on this one; the move ended up dropping the prefix and suffix on the serialized string here -- it should be:

```
write!(f, "[[ {extended_test_expr_command} ]]")
```

reubeno (changes requested):
Change is looking in quite good shape. Just one comment on the peg code comment, and one issue with the string formatting of it.

Thanks for the iteration on this!

brush-parser/src/ast.rs
vivo75: Hi @reubeno I'm on holiday and will be the whole next week and will be unable to interact with this pull request untill I return.
Sorry if the fix is easy feel free to apply it yourself, otherwise I (actually claude code) will manage it on return

reubeno (approved):

reubeno:
@lu-zero FYI on this change, particularly because it includes a change to the AST. I think this is probably where the extended test expressions should have been in the first place. If you're able to skim through the change, too, would be great to get your feedback.

lu-zero:
It seems fine to me.

reubeno (approved):

PR #1283, opened 2026-08-20
fix(expansion): read a negative substring length as an offset from the end

luantaraschi:
A negative length in `${var:offset:length}` says where the slice ends, counting back from the end of the value, rather than how long the slice is. The code added the value's length to it and then treated the result as a count, which is the same thing only when the offset is zero. That is why the existing cases in `params.yaml` pass: they all use offset 0.

On `var=abcdefgh`:

| | bash | before | after |
| --- | --- | --- | --- |
| `${var:2:-2}` | `cdef` | `cdefgh` | `cdef` |
| `${var: -4:-1}` | `efg` | `efgh` | `efg` |
| `${var:0:-1}` | `abcdefg` | `abcdefg` | `abcdefg` |

bash also treats a length that reaches back past the offset as an error rather than as an empty result. `${var:2:-7}` on the same value prints `-7: substring expression < 0`, where brush printed nothing and carried on. It now returns a `CheckedExpansionError` carrying the same wording, for anyone grepping for it.

That error is deliberately not fatal. An earlier version of this description said bash stops the script, and that was wrong. Measured on 5.2, bash prints the error and keeps going:

```
var=abcdefgh
echo "${var:2:-7}"          # prints -7: substring expression < 0
echo REACHED-AFTER          # still runs, shell exits 0
```

The exit status of 1 only appears when the bad expansion is the last command in the shell. `${var:?msg}` is the case that really does end the script, which is why that path calls `into_fatal()` and this one does not.

Checked against bash for every combination of a positive, negative and out-of-range offset with a negative length, plus the arithmetic forms `${var:1:-n}` and `${var:1:(-2)}`, and for non-negative lengths to confirm nothing there moved. One corner worth naming: an offset past the end of the value stays an empty result and not an error, `${var:9:-1}` on a value of 8 characters, while `${var:8:-1}` is the error. The tests pin both.

Scope: after the rebase this also covers the array forms, carrying the related changes from #1227. The rule there turned out simpler than I first described. All three array-shaped expansions reject any negative length, while a string only rejects one that reaches back past the offset:

```
str    ${var:2:-2}       -> cdef
arr    ${arr[@]:2:-2}    -> substring expression < 0
pos    ${@:2:-2}         -> substring expression < 0
assoc  ${assoc[@]:2:-2}  -> substring expression < 0
```

Sparse indexed arrays also count the offset by subscript rather than by element position. With `sparse=([0]=a [3]=d [7]=h)`, both `${sparse[@]:1:2}` and `${sparse[@]:3:2}` give `d h`.

### DIFF
--- a/brush-core/src/expansion.rs
+++ b/brush-core/src/expansion.rs
@@ -82,11 +82,89 @@ impl Default for ExpanderOptions {
 /// consumes both characters, leaving nothing.
 pub(crate) const DOUBLE_QUOTED_ESCAPE_CHARS: &[char] = &['\\', '$', '`', '"', '\n'];
 
+/// Describes what a parameter's expansion is a list of. Operators whose behavior depends
+/// on that shape -- `${#x}` and `${x:offset:length}` today -- consult this to decide
+/// whether they address characters or elements, via the methods below.
+#[derive(Clone, Debug)]
+enum ExpansionKind {
+    /// A single string; addressed by character.
+    String,
+    /// The elements of an indexed array, along with their subscripts in ascending order.
+    /// The subscripts matter because bash addresses these by subscript rather than by
+    /// position, which only makes a difference when the array is sparse.
+    IndexedArray(Vec<u64>),
+    /// Elements with no indices to address them by: an associative array, the positional
+    /// parameters, or a generated list of names.
+    ElementList,
+    /// A scalar that was written `${x[@]}` or `${x[*]}`. bash counts it as a single
+    /// element but slices it as a string.
+    ScalarAsArray,
+}
+
+impl ExpansionKind {
+    /// Returns whether `${x:offset:length}` rejects a negative length, which bash does for
+    /// every real array form.
+    const fn is_array(&self) -> bool {
+        matches!(self, Self::IndexedArray(_) | Self::ElementList)
+    }
+
+    /// Returns whether `${#x}` counts characters rather than elements.
+    const fn counts_chars(&self) -> bool {
+        matches!(self, Self::String)
+    }
+
+    /// Returns whether `${x:offset:length}` addresses characters rather than elements.
+    /// bash slices a scalar as a string even when it was written `${x[@]}`, so this is not
+    /// the same question as [`Self::counts_chars`].
+    const fn slices_chars(&self) -> bool {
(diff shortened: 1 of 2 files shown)


reubeno:
@luantaraschi  -- I've taken the liberty of rebasing against latest changes in `main` as well as adding a commit with some related changes from #1227 as well as some additional modifications. I believe it will more comprehensively correct substring/sub-array slicing.

luantaraschi:
Thanks for folding in #1227, that covers more than I was willing to guess at on my own.

One correction, and it is mine. The description says bash stops the script on this error. It does not. I took that from the exit status without checking what came after it. On bash 5.2:

```
$ cat two.sh
var=abcdefgh
echo "A: ${var:2:-7}"
echo "REACHED-AFTER-SUBSTRING"

$ bash two.sh
REACHED-AFTER-SUBSTRING
two.sh: line 2: -7: substring expression < 0
$ echo $?
0
```

The exit 1 I quoted only appears when the bad expansion happens to be the last command in the shell. The form that really is fatal behaves differently:

```
$ bash three.sh            # echo "${u:?is unset}" then echo REACHED-AFTER-QUESTION
three.sh: line 2: u: is unset
$ echo $?
1
```

So these are two different errors in bash, and treating them differently is right: the `:?` path calls `into_fatal()` and this one does not. Making the substring error fatal would move it away from bash rather than towards it.

The array rule also came out simpler than I claimed. I wrote that the condition differs across indexed arrays, associative arrays and positional parameters. Measured, it is the same for all three, and it is not the string rule:

```
str    ${var:2:-2}      -> cdef
arr    ${arr[@]:2:-2}   -> substring expression < 0
pos    ${@:2:-2}        -> substring expression < 0
assoc  ${assoc[@]:2:-2} -> substring expression < 0
```

Any negative length is rejected for the three array forms, while a string only rejects one that reaches back past the offset. That is what the branch now does, so it is the scope note in my description that is stale, not the code.

Same for the sparse offsets: bash counts subscripts, not element positions.

```
sparse=([0]=a [3]=d [7]=h)
${sparse[@]:1:2}  -> d h
${sparse[@]:3:2}  -> d h
```

Updating the description now so it stops misleading whoever reads it next.

PR #1284, opened 2026-08-20
fix(expansion): keep the leading zeros in a numeric brace sequence

luantaraschi:
A numeric brace sequence written with a leading zero counts with the zero kept: `{01..05}` is `01 02 03 04 05`. brush dropped it and produced `1 2 3 4 5`, because the bounds were parsed into `i64` and the text they were written as was gone by the time the sequence was generated.

```
$ echo {01..05}
bash:   01 02 03 04 05
before: 1 2 3 4 5
after:  01 02 03 04 05

$ echo {0009..11}
bash:   0009 0010 0011
before: 9 10 11
after:  0009 0010 0011
```

It is quiet when it goes wrong. `for i in {01..12}` over a directory of `01.log` through `12.log` just stops matching anything.

The rule, taken from bash by measurement rather than from the manual, which does not spell all of it out:

- A leading zero in either bound turns padding on. A lone `0` is not a leading zero, so `{0..10}` stays `0 1 2 ... 10`.
- The width is the longer of the two bounds as they were written, sign included: `{-01..01}` gives `-01 000 001`, and `{1..005}` gives `001` through `005`.
- A zero that follows a `+` does not turn padding on, though the text still counts toward the width once the other bound has asked for it. `{+01..3}` is `1 2 3` and `{+01..05}` is `001` through `005`.
- The increment has no say in either question. `{1..10..03}` is `1 4 7 10`.

Checked against bash for 53 sequences, covering both directions, both signs, a step, a prefix and a suffix, and the cases where padding must not happen.

**Public API change**, since the checklist asks for these to be called out: `BraceExpressionMember::NumberSequence` gains a `zero_padded_width: Option<usize>` field. Nothing else in the workspace constructs or matches that variant, but it is a breaking change for anyone outside it who does. The alternative is carrying the two bounds' source text on the variant instead, which is a wider change for the same information; happy to go that way if you would rather keep the decision out of the parser.

One thing came along for the ride. The new bound rule declines when a bound does not fit in an `i64` instead of unwrapping, so `echo {99999999999999999999..1}` now prints itself back the way bash does, where before it panicked the parser:

```
thread 'main' panicked at brush-parser/src/word.rs:760:38:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: PosOverflow }
```

Sign and digits go to the parser together, so the bottom of the range still expands, which it would not if the digits went in on their own:

```
$ echo {-9223372036854775808..-9223372036854775806}
bash:  -9223372036854775808 -9223372036854775807 -9223372036854775806
brush: -9223372036854775808 -9223372036854775807 -9223372036854775806
```

That is the same class of thing #1271 is fixing in `io_number`, in a different rule. The sequence increment still goes through the old `number()` rule and still unwraps; I left it alone rather than widening this beyond the bounds it needs.

### DIFF
--- a/brush-core/src/braceexpansion.rs
+++ b/brush-core/src/braceexpansion.rs
@@ -34,14 +34,23 @@ fn expand_brace_expr_member(bem: word::BraceExpressionMember) -> Box<dyn Iterato
             start,
             end,
             increment,
+            zero_padded_width,
         } => {
             let mut increment = increment.unsigned_abs() as usize;
             if increment == 0 {
                 increment = 1;
             }
 
+            // A bound written with a leading zero asks for every member to be padded
+            // out to the width of the longer bound, so `{01..10}` counts `01 02 ...`
+            // and not `1 2 ...`. The sign takes one of those columns.
+            let format = move |n: i64| match zero_padded_width {
+                Some(width) => std::format!("{n:0width$}"),
+                None => n.to_string(),
+            };
+
             if start <= end {
-                Box::new((start..=end).step_by(increment).map(|n| n.to_string()))
+                Box::new((start..=end).step_by(increment).map(format))
             } else {
                 // Iterate from start down to end by decrementing.
                 #[allow(clippy::cast_possible_wrap)]
@@ -51,7 +60,7 @@ fn expand_brace_expr_member(bem: word::BraceExpressionMember) -> Box<dyn Iterato
                         let next = n - increment;
                         (next >= end).then_some(next)
                     })
-                    .map(|n| n.to_string()),
+                    .map(format),
                 )
             }
         }
--- a/brush-parser/src/word.rs
+++ b/brush-parser/src/word.rs
@@ -504,6 +504,9 @@ pub enum BraceExpressionMember {
         end: i64,
         /// Increment value.
         increment: i64,
+        /// Width to zero pad each member out to, set when either bound was
+        /// written with a leading zero.
+        zero_padded_width: Option<usize>,
     },
     /// An inclusive character sequence.
     CharSequence {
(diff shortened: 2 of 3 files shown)


reubeno:
Thanks for the contribution, @luantaraschi! The change is looking pretty good. I've unblocked the PR checks and will monitor them.

reubeno:
@luantaraschi Looks like some clippy failures; could you take a look?

In case it's helpful `cargo xtask ci pre-commit` in this project will generally run the same set of checks run against PRs. (Requires `cargo-nextest`, `cargo-deny`, and clippy to be installed.)

luantaraschi:
Took a look, and it is not coming from this branch or from #1283. Both errors land in `brush-test-harness/src/execution.rs`, at lines 209 and 293, and neither PR touches that file. The same pair shows up on `main` at 18851e7, in run 32410152610.

The lint is `clippy::unused_async_trait_impl`, which is new in 1.98.0 and comes in through the pedantic group. The `#[expect(clippy::unused_async)]` already sitting on both functions does not cover it, since it is a separate lint. The workspace allows `unknown_lints`, so putting the new name next to the old one stays quiet on the older toolchains in the matrix. I checked that part on 1.97.1.

Thanks for the `cargo xtask ci pre-commit` pointer, I had not run into it. Happy to send the fix as its own PR if you want it off your plate.

reubeno:
> Took a look, and it is not coming from this branch or from #1283.
> ...
> Happy to send the fix as its own PR if you want it off your plate.

Thanks for triaging this -- that was very helpful. Definitely appreciate the offer, but I was able to address the issues and get something merged.

I'd expect that rebasing this PR against `main` should get things into better shape again!

luantaraschi:
Rebased, and the checks come back clean here: fmt, clippy and the compat suite with no failures. Thanks for sorting out the clippy side, that was quicker than my offer would have been. The other two are rebased as well.

luantaraschi:
One more push onto the same commit. The Copilot note above holds up: the bound rule parsed the digits without the sign, so the bottom of the range declined and printed itself back where bash expands it.

```
$ echo {-9223372036854775808..-9223372036854775806}
bash:   -9223372036854775808 -9223372036854775807 -9223372036854775806
before: {-9223372036854775808..-9223372036854775806}
after:  -9223372036854775808 -9223372036854775807 -9223372036854775806
```

Sign and digits go to the parser together now. The boundary is in the parser test and there is a compat case on either side of it. fmt and clippy clean, and the compat suite is 1849 passing with 0 failures.

reubeno (approved):

PR #1294, opened 2026-08-22
fix(builtins): avoid deadlock when a builtin reads a process substitution

cataggar:
## Overview

Fix a deadlock when an in-process builtin reads from a process substitution inside a command substitution.

This change was made by GitHub Copilot at my direction.

## Details

Commands such as the following previously never returned:

```bash
x=$(mapfile -t lines < <(printf "one\ntwo\n"); echo "${lines[*]}")
```

The process-substitution producer is spawned on the async runtime, while `mapfile` and `read` synchronously consume its descriptor. Blocking the runtime worker can prevent the producer from running and closing the write end.

The synchronous descriptor reads now run through an awaited private `spawn_blocking` path on Unix and Windows. This works on multi-thread runtimes, current-thread runtimes, and inside `LocalSet`. Non-threaded targets such as WASM retain inline behavior. No new `brush-core` public API is introduced.

Regression coverage includes Bash-oracle compatibility cases plus Rust integration tests for `read` and `mapfile` under `LocalSet` and current-thread runtimes. Current-thread deadlock checks run in externally timed subprocesses so the test harness can terminate a regression.

## Testing

- `cargo test --package brush-builtins --test runtime_read_tests`
- `cargo test --test brush-compat-tests -- mapfile`
- `cargo test --test brush-compat-tests -- "while read loop fed by process substitution"`
- `cargo test --test brush-compat-tests -- "Command substitution containing"`
- `cargo clippy --package brush-core --package brush-builtins --package brush-shell --all-features --all-targets`
- `cargo check --package brush-builtins --target wasm32-wasip2`
- `cargo xtask check fmt`
- `cargo xtask test integration --coverage` (2697 passed; 2 unrelated environment failures because `which` is unavailable; report generated)

### DIFF
--- a/brush-builtins/Cargo.toml
+++ b/brush-builtins/Cargo.toml
@@ -171,3 +171,4 @@ procfs = "0.18.0"
 [dev-dependencies]
 anyhow = "1.0.102"
 pretty_assertions = { version = "1.4.1", features = ["unstable"] }
+tokio = { version = "1.52.3", features = ["time"] }
--- a/brush-builtins/src/lib.rs
+++ b/brush-builtins/src/lib.rs
@@ -118,6 +118,23 @@ mod wait;
 
 mod builder;
 mod factory;
+
+async fn run_blocking_io<T>(
+    f: impl FnOnce() -> Result<T, brush_core::Error> + Send + 'static,
+) -> Result<T, brush_core::Error>
+where
+    T: Send + 'static,
+{
+    #[cfg(any(unix, windows))]
+    {
+        tokio::task::spawn_blocking(f).await?
+    }
+
+    #[cfg(not(any(unix, windows)))]
+    {
+        f()
+    }
+}
 mod unimp;
 
 pub use builder::ShellBuilderExt;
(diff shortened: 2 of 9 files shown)


cataggar:
Thanks for the review. I confirmed that `runtime_flavor() == MultiThread` does not guarantee `block_in_place` is permitted: it can panic inside a Tokio `LocalSet`, while the inline fallback can retain the deadlock on a current-thread runtime. I’m moving this PR back to draft and will replace the helper with awaited `spawn_blocking`, remove the unintended `brush-core` public API, and add runtime-context regression coverage before requesting review again.

cataggar:
Implemented the requested changes in a4c058e:

- replaced `block_in_place` with awaited `spawn_blocking` for synchronous `read`/`mapfile` descriptor reads;
- removed the unintended public `brush-core` helper;
- preserved inline behavior for WASM/non-threaded targets;
- added `LocalSet` and current-thread regression tests, with current-thread deadlocks guarded by an external subprocess timeout.

Coverage report: 76.66% overall, 91.09% for `read.rs`, 88.89% for `mapfile.rs`, and 85.00% for the private helper in `lib.rs`. The coverage run passed 2697/2699 tests; the two failures are unrelated compatibility cases caused by `which` not being installed in this environment. Targeted runtime, compatibility, clippy, formatting, and WASM checks pass.

Moving this back to ready for review.

PR #1295, opened 2026-08-22
fix(core): quote arguments passed to external commands on Windows

cataggar:
## Overview

Preserve shell-expanded arguments when launching external commands on Windows, including programs built on the MSYS2 or Cygwin runtime.

This change was made by GitHub Copilot at my direction.

## Details

Windows passes a child process one command-line string rather than an `argv` array. Rust's standard library quotes arguments when required for native Windows parsing, but otherwise leaves them unquoted. MSYS2 and Cygwin programs can then apply their own glob, brace, tilde, and quote processing to those arguments, corrupting values that the shell has already expanded.

Examples include:

```text
grep -oE '[0-9]{3}'   -> [0-9]3
grep -oE 'MAC\(x'     -> MAC(x
find . -name '*.txt'  -> every matching file in the directory
printf '%s' '~'       -> the home directory
```

This change:

- routes command argument setup through a platform-specific `sys::commands::set_args`;
- leaves Unix and stub targets on the existing `Command::args` behavior;
- adds a Windows implementation that explicitly quotes every argument using `CommandLineToArgvW` and Microsoft C runtime rules;
- correctly escapes embedded quotes and trailing backslashes;
- preserves the standard library path for `cmd.exe`, `.bat`, and `.cmd` files because they require different escaping and injection protections; and
- keeps command tracing based on the shell-expanded arguments rather than their Windows command-line encoding.

The Windows module includes focused tests for ordinary arguments, MSYS metacharacters, embedded quotes, trailing backslashes, `cmd.exe` detection, batch-file behavior, and an optional end-to-end check using Git for Windows' MSYS2 `printf.exe`.

## Testing

- `cargo test --package brush-core`
- `cargo check --package brush-core --all-features --all-targets --target x86_64-pc-windows-gnu`
- `cargo clippy --package brush-core --all-features --all-targets --target x86_64-pc-windows-gnu`
- `cargo xtask check fmt`

### DIFF
--- a/brush-core/src/sys/windows/commands.rs
+++ b/brush-core/src/sys/windows/commands.rs
@@ -1,0 +1,300 @@
+//! Command execution utilities.
+
+use std::{
+    ffi::{OsStr, OsString},
+    os::windows::{
+        ffi::{OsStrExt, OsStringExt},
+        process::CommandExt as _,
+    },
+    path::Path,
+};
+
+pub use crate::sys::stubs::commands::{
+    CommandExt, CommandFdInjectionExt, CommandFgControlExt, ExitStatusExt,
+};
+
+/// Sets the arguments to be passed to the given command.
+///
+/// Windows has no `argv` array; a child process receives a single command-line
+/// string and is responsible for parsing it itself. The standard library only
+/// wraps an argument in double quotes when it is empty or contains whitespace
+/// or a quote, which leaves every other argument exposed to a second round of
+/// interpretation in the child.
+///
+/// That second round is not hypothetical: programs built on the MSYS2/Cygwin
+/// runtime (for example, the tools shipped with Git for Windows) re-expand the
+/// arguments they receive from a native parent, applying glob, brace, and tilde
+/// expansion along with quote removal. Words that the shell has already
+/// finished expanding are silently corrupted as a result.
+///
+/// To keep the child's view of its arguments identical to the shell's, each
+/// argument is explicitly quoted using the rules understood by
+/// `CommandLineToArgvW` and the Microsoft C runtime. Native programs are
+/// unaffected by the added quotes, since they strip them while parsing.
+///
+/// # Arguments
+///
+/// * `cmd` - The command to set arguments on.
+/// * `args` - The arguments to pass to the command.
+pub fn set_args<S: AsRef<OsStr>>(cmd: &mut std::process::Command, args: &[S]) {
+    // Arguments destined for `cmd.exe` are deliberately left alone; see
+    // `uses_cmd_exe_parsing` for why.
+    if uses_cmd_exe_parsing(cmd.get_program()) {
+        cmd.args(args);
+        return;
+    }
+
+    for arg in args {
+        cmd.raw_arg(quote_arg(arg.as_ref()));
+    }
+}
+
+/// Returns true if the given program's arguments will be parsed by `cmd.exe`.
+///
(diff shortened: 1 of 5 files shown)


brush-core/src/sys/windows/commands.rs
```diff
@@ -0,0 +1,300 @@
...
+///
+/// Windows has no `argv` array; a child process receives a single command-line
+/// string and is responsible for parsing it itself. The standard library only
+/// wraps an argument in double quotes when it is empty or contains whitespace
+/// or a quote, which leaves every other argument exposed to a second round of
+/// interpretation in the child.
+///
+/// That second round is not hypothetical: programs built on the MSYS2/Cygwin
+/// runtime (for example, the tools shipped with Git for Windows) re-expand the
+/// arguments they receive from a native parent, applying glob, brace, and tilde
```
reubeno, line 26: This makes it sound like the issue is not actually a `brush`-specific issue, but a challenging assumption made by msys2/cygwin tools. Perhaps I'm missing something, but wouldn't this be a problem for *any* Rust application that might use std::process::Command to invoke a program on Windows that might behave like this?

Do you know if there are precedents for how any other such apps have handled this?

brush-core/src/sys/windows/commands.rs
```diff
@@ -0,0 +1,300 @@
...
+/// bypassing that with a raw command line would reintroduce a command injection
+/// hazard.
+///
+/// Neither `cmd.exe` nor a batch file is an MSYS2/Cygwin program, so no
+/// protection is lost by deferring to the standard library here.
+///
+/// # Arguments
+///
+/// * `program` - The program the command will execute.
+fn uses_cmd_exe_parsing(program: &OsStr) -> bool {
```
reubeno, line 68: This kind of callee-specific logic, to me, begs the question of how many other applications may similarly have challenges and what risk there is in encoding such specific logic in our general execution paths. I understand there are real limitations that this PR is trying to mitigate but I worry that this is a slippery slope toward accumulated heuristics for how to interact with specific (Windows) invoked programs, and that's something that I would expect a general-purpose shell avoids.

reubeno (commented):
Thanks for highlighting the issue that you're seeing in this scenario. May main concern is whether this is addressing the problem at the right layer of the stack, and whether this kind of logic could cause problems for other called programs.

Let me do a bit more research, think about it a bit more, and respond back with some other suggestions for consideration.

reubeno:
@cataggar -- some additional things I found or thought about while doing a bit more research:

* There's an unstable feature for enabling `force_quotes` behavior on commands in Rust. Looks like you found it as well, as I see you referenced it.

* There's a `noglob` option that can be set via env vars (`CYGWIN` or `MSYS`) when launching cygwin/msys processes -- and that looks like it inhibits the default (re)expansion. I think it's worth considering whether that would mitigate the problem. (I don't understand offhand whether there are negative ramifications of this or if it would leak past the directly invoked child.) I'm not sure how I'd feel about `brush` default setting this, but if it were set in a brush-specific rc file, I wonder if it would be sufficient to cover the case here.

* As best as I can tell, the core Python library doesn't handle this in its `subprocess` module -- but I do suppose callers could always work around that. I'm really curious to know if any other cross-platform runtimes specifically try to handle this in some way.

* If we convince ourselves we do need to support some kind of force-quoting behavior, I think it should be something that's configurable and not unconditionally hard-wired behavior.

PR #1300, opened 2026-08-24
experiment: replace clap with usage-rs (usage) for CLI/builtin parsing

lu-zero:
# Experiment: replace clap with usage-rs

Compares [usage-rs](https://crates.io/crates/usage-rs) (the Rust framework behind `usage-cli`) against clap 4.6 as brush's argument-parsing layer. Kept as a separate branch/worktree so both parsers can be built and measured from the same sources.

> **Scope**: experimental comparison branch, not a finished migration. Performance numbers are tracked in #1301.

## What was migrated

All shipped runtime crates now parse with `usage` (v6):

- `brush-core` — `builtins::Command` infrastructure, completion value enums, example
- `brush-builtins` — all ~45 builtin argument structs
- `brush-experimental-builtins`
- `brush-shell` — main CLI (`args.rs`, `entry.rs`, `brushctl.rs`, `events.rs`)

Deliberately **not** migrated (dev tooling, not part of the shipped binary): `xtask`, `brush-test-harness`.

## Notable design changes

- usage's derive generates *inherent* methods rather than trait impls, so `builtins::Command` is no longer bounded on `clap::Parser`. A small new `UsageParse` trait bridges parsed types; each builtin adds one line of glue via `brush_core::impl_usage_parse!(T)` (`brush-core/src/builtins.rs`).
- usage parse errors borrow from argv; they are rendered eagerly into an owned `builtins::ParseError` that preserves clap's exit-code contract (help/version → 0, failure → 2).
- Strictness parity where clap was strict: `unknown_flags = "error"` + `args_override_self = false`. Builtins whose operands may legitimately look like flags (`echo`, `test`, `printf`, `let`, `eval`, …) use the permissive mode instead.
- The existing `-/+` option trick still works unchanged (`+x` → `--+x`, long declared as `"+x"`); verified against usage's parser.
- `set -o/+o`: repeated occurrences accumulate like bash; bare `-o`/`+o` (list-all) is distinguished from an explicitly empty name via an unspellable sentinel attached during `SetCommand::new`'s existing argv rewrite pass.
- `printf` gained a `new()` override so a standalone `--` appearing *after* the format string stays an ordinary operand (bash semantics the parser can't express).
- Colors: help pages and diagnostics are styled on color-capable streams; usage honors `NO_COLOR` / `CLICOLOR_FORCE` / per-stream tty detection. Only clap's *custom* palettes are non-portable.
- `xtask gen man/markdown` now emit the CLI's usage KDL spec (renderable by `usage-cli`); completion scripts come from usage's `completions` feature.
- `scripts/parser-parity.sh <ref> <cand>`: differential harness running a corpus of parsing-sensitive command lines through two builds and diffing stdout + exit status (20 cases covering every drift class found during this experiment).

## Verification

- `cargo xtask ci quick` green (fmt, clippy, unit tests)
- Compat suite: **2306/2306 passing**; full integration run clean except pre-existing flaky `run_suspend_and_fg` (flakes identically on `main`; pty/job-control timing)
- Real-world scripts verified for three-way output parity (bash vs clap build vs usage build) before any timing
- `scripts/parser-parity.sh`: 20/20 vs a clap build

## Performance

Microbenchmarks (hyperfine, release builds):

| scenario | main (clap) | this branch | Δ |
|---|---|---|---|
| startup: `brush -c 'exit 0'` | 7.1 ms | 7.2 ms | ~noise |
| `brush --help` | 1.3 ms | 1.3 ms | ~noise |
| builtin-parse hot loop (`let` ×3000) | 69.2 ms | 58.3 ms | ~1.19× |
| stripped binary size | 6.49 MB | 6.71 MB | +3.4% |

Real-world scripts (details and reproduction in #1301):

| workload | parse density | clap | usage | Δ |
|---|---|---|---|---|
| `gnuconfig/config.guess` (1818 ln) | external-bound | 177.7 ms | 177.4 ms | 1.00× |
| repo `scripts/test-across-shells.sh` | moderate | 42.4 ms | 42.0 ms | 1.01× |
| `benchmarks/real-world/deploy-sim.sh` | mixed | 92.3 ms | 89.2 ms | 1.03× |
| sourcing `bash_completion` (3624 ln) | dense builtins | 47.7 ms | 42.0 ms | **1.13×** |
| `benchmarks/real-world/config-lint.sh 250` | getopts per entry | ~71–78 ms | ~47–54 ms | **~1.5×** |
| `config-lint.sh 750` | same, scaled | 163.9 ms | 98.5 ms | **~1.66×** |

**Reading:** end-to-end time on I/O- or external-command-bound scripts is indistinguishable between parsers; as builtin argument parsing becomes the bottleneck, usage shows a consistent win that grows with parse density (~1.1× → ~1.7×).

## Known deviations / follow-ups

- Help pages use usage's renderer/layout with its automatic palette; clap's custom style palettes (`Styles::styled()` with brush's yellow/green/magenta/cyan scheme) are not portable.
- `override_usage` (declare) dropped — usage only takes literal usage lines; colorized usage line lost.
- `ulimit`: dynamic "(supported)/(unsupported)" help suffixes became static text.
- `exit`/`fc` operands: hyphen-leading non-numeric tokens now error instead of being captured (negative numbers still work).
- Public API changes (breaking): `builtins::Command` bounds, removal of `parse_known`, `clap::Error` → `builtins::ParseError` in `Command::new`.

Assisted-by: ox-alpha:opencode/x-preview-f-free

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,6 +405,7 @@ dependencies = [
  "thiserror 2.0.20",
  "tokio",
  "tracing",
+ "usage-rs",
  "uucore",
 ]
 
@@ -444,6 +445,7 @@ dependencies = [
  "thiserror 2.0.20",
  "tokio",
  "tracing",
+ "usage-rs",
  "uuid",
  "uzers",
  "whoami",
@@ -545,6 +547,7 @@ dependencies = [
  "brush-core",
  "clap",
  "serde_json",
+ "usage-rs",
 ]
 
 [[package]]
@@ -651,6 +654,7 @@ dependencies = [
  "toml",
  "tracing",
  "tracing-subscriber",
+ "usage-rs",
  "uuid",
  "version-compare",
  "walkdir",
@@ -865,15 +869,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "clap-markdown"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2a2617956a06d4885b490697b5307ebb09fec10b088afc18c81762d848c2339"
-dependencies = [
- "clap",
-]
-
-[[package]]
 name = "clap_builder"
 version = "4.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -884,15 +879,6 @@ dependencies = [
  "clap_lex",
  "strsim",
  "terminal_size",
-]
-
-[[package]]
-name = "clap_complete"
-version = "4.6.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3be2ad0423bdbbb0e25bc89add796f3559706d4a95e1bc98e4d9662a957b6a19"
-dependencies = [
- "clap",
 ]
 
 [[package]]
@@ -912,16 +898,6 @@ name = "clap_lex"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c8d4a3bb8b1e0c1050499d1815f5ab16d04f0959b233085fb31653fbfc9d98f9"
-
-[[package]]
-name = "clap_mangen"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "211d617eaa4b735c96c9e0228fcbdb5120ef623f2b8cb67ffb84c3e02dbc28a4"
-dependencies = [
- "clap",
- "roff",
-]
 
 [[package]]
 name = "color-print"
@@ -3684,12 +3660,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "roff"
-version = "1.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "323c417e1d9665a65b263ec744ba09030cfb277e9daa0b018a4ab62e57bc8189"
-
-[[package]]
 name = "ron"
 version = "0.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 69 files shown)


lu-zero:
Follow-up commit `092840b9` fixes the worst deviation reported in the initial description: repeated `-o`/`+o` options on `set` now accumulate again, matching bash and `main`.

Implementation note for reviewers comparing parser expressiveness: usage models per-occurrence-optional values via `default_missing`; combining it with a `Vec<String>` field restores clap's `num_args(0..=1)` + repeatable behavior:

```rust
#[usage(short = 'o', default_missing = "", value_name = "OPT")]
enable: Vec<String>,   // "set -o" -> [""], "set -o a -o b" -> ["a", "b"]
```

Verified: `set -o nounset -o xtrace`, mixed `-o/+o`, attached forms (`-oa`), bare listing (`set -o`, `set +o` byte-identical to `main`), plus the full compat suite.

lu-zero:
Follow-up commit `a4be20eb` closes the remaining gap from `092840b9`: with `default_missing = ""`, a *bare* `-o` and an explicitly empty name (`set -o ""`) were indistinguishable — both listed options, where bash and `main` reject the empty name (rc 2).

`SetCommand::new` now detects bare occurrences during its existing `+x`-renaming pass (next word missing or flag-like) and attaches an unspellable sentinel (`BARE_OPTION = "\u{0}"`) as the occurrence's value; explicit values are untouched. `execute()` triggers list-all only on the sentinel, so:

| input | behavior now | matches |
|---|---|---|
| `set -o nounset -o xtrace` | both applied | bash + main ✔ |
| `set -o` / `set +o` | list all | byte-identical to main ✔ |
| `set -o ""` | rc=2 | main ✔ (was rc=0) |
| `set -o -x` / `set -eo` / `set -oa -ob` | identical to main | ✔ |

**Detection**: also added `scripts/parser-parity.sh` — a differential harness that runs a corpus of parsing-sensitive command lines through two brush builds and diffs stdout + exit status. It encodes every drift class this experiment surfaced (±o accumulation/bare/odd values, flag-like operands for echo/printf/test/let, post-format `--`). Currently 20/20 vs a clap build; extend `CORPUS` as new cases appear.

    scripts/parser-parity.sh /path/to/clap-brush /path/to/usage-brush

lu-zero:
## Real-world script benchmarks

Added `benchmarks/real-world/` with two deterministic workloads (`deploy-sim.sh`, `config-lint.sh`), all verified for **three-way output parity** (bash vs clap build vs usage build) before timing. Timed with hyperfine (release builds, warmup, 100–300 runs):

| workload | what it stresses | clap build | usage build | Δ |
|---|---|---|---|---|
| `gnuconfig/config.guess` (1818 ln) | classic configure-style; external-command bound | 177.7 ms | 177.4 ms | 1.00× |
| repo's own `scripts/test-across-shells.sh` | getopts-style option loop, arrays, heredocs | 42.4 ms | 42.0 ms | 1.01× |
| `deploy-sim.sh -vn -j8 -t …` | mixed: getopts + pushd/popd + declarations + tar | 92.3 ms | 89.2 ms | 1.03× |
| sourcing `bash_completion` (3624 ln) | dense builtin parsing, no external cmds | 47.7 ms | 42.0 ms | **1.13×** |
| `config-lint.sh 250` | fresh `getopts` pass per entry (~2.5k builtin parses) | ~71–78 ms | ~47–54 ms | **~1.5×** |
| `config-lint.sh 750` | same, scaled | 163.9 ms | 98.5 ms | **~1.66×** |

**Reading:** on scripts dominated by external commands or I/O (`config.guess`, deploy), the parsers are indistinguishable — end-to-end time is elsewhere. As the share of builtin argument parsing rises, a consistent, growing win for usage appears: **~1.1× → ~1.7× end-to-end**, at parity of behavior (also re-verified with `scripts/parser-parity.sh`: 20/20).

So the earlier "hundreds of times faster" marketing number doesn't translate to shell startup, but for parse-heavy real-world scripts the improvement is material and scales with parse density.

Issue #520, opened 2025-05-25
Shall we consider switching to bpaf?


lu-zero:
Recently I considered switching to bpaf in pkgcraft since we need to support some syntax that clap does not really accept. 
Since we might end up having a similar problem here, shall we consider it as well?

Some problems solved by some effort with clap:
- `--` special usage
- `+/-flag` to enable/disable it

reubeno:
There's a lot of reasons I like `clap` and I wonder how open they'd be to someone working through these limitations with them, but I just haven't had the time to dive into it.

So with that said, I think it's worthwhile to evaluate `bpaf` to see if it provides what we need out of the box. I don't have experience with it but I think we should first summarize all the oddities / syntactic forms we need to support and then evaluate how well it would let us handle them (and still have decent errors, usage info, etc.).

In short, am totally open to that. It would be interesting also to see if it's more efficient; we spend a small but noticeable amount of time in parsing when invoking a builtin. We've got flamegraphs we could compare if needed.

pacak:
> I don't have experience with it but I think we should first summarize all the oddities / syntactic forms we need to support 

Having such a list would be a great start. Then ping me.

lu-zero:
Exactly, I do not remember right now if there are more cases as of now.

reubeno:
* We had to put workarounds in place for implementations of `set`, `getopts`, and `echo` for handling `--`, as well as the top-level command line parsing for the shell. `set` also needed to specially handle `-` as an argument.

    Relevant PRs (in order) are: #147 #235 #354.

* The `+`/`-` pattern is one we had to work around for multiple builtins as well (e.g., `declare`, `set`).

* `dirs`/`pushd`/`popd` all support a +*N*/-*N* syntax (where *N* can be any integer); we haven't yet worked around those, but need to do so.

Those are the main cases I'm aware of.

pacak:
For `--` I'll have to implement something separately, most likely to try to consume it with a regular parser, if not - treat it in a special way. A small change.

For `-` it should be able to parse it as a positional item as is.

Things like standalone `+a` / `-a` - they are easy to parse info `A::Enabled` / `A::Disabled` with help of `any`, but for combinations like `+abc` to be parsed into a collection - this would require some changes on my side. By support I mean a first class support - that is you define parser that knows that parsers for `a` and `b` go with `+` and `-` and it deals with `+ab` by itself. Consuming it into a string `"+ab"` and parsing it by hand is possible but not very convenient.

[...]

PR #1302, opened 2026-08-24
WIP: port argument parsing from clap to bpaf

lu-zero:
Draft PR exploring a port of brush's argument parsing from clap to [bpaf](https://github.com/petevine/bpaf) (0.9), linked to #520.

## What this does

Replaces clap with bpaf in the runtime crates (`brush-core`, `brush-builtins`, `brush-experimental-builtins`, `brush-shell`, `brush-test-harness`), redesigning the builtin `Command` trait around bpaf parser combinators:

- `builtins::Command` now requires `fn parser() -> impl bpaf::Parser<Self> + 'static` plus small metadata hooks (`about()`, `synopsis()`) that feed the `help` builtin. Detailed help is rendered by triggering bpaf's own `--help` handling.
- A single bash-faithful splitter (`builtins::split_option_section`) decides where the option section ends for builtins with verbatim trailing operands (`echo`, `test`, `set`, `fc`, ...). Builtins declare their value-taking short options; the splitter honors `--`, first-operand termination, and attached values (`-d:`, `-G=--foo`).
- **Plus-style options** (`set +x`, `declare +a`) no longer use the old clap-era `+x → --+x` encoding: `+abc` groups are expanded the way the shell tokenizes them and matched with bpaf `literal("+x")` parsers. As a bonus, literal `--+o foo` input — which bash rejects and the old encoding silently accepted as `+o` — is now rejected like bash.
- `clap::ValueEnum` types become plain `FromStr` enums (completion actions/options, trace events, input backends).
- The shell CLI is a combinatoric parser with bash-compatible `-c`/`--` handling; script arguments are captured verbatim after the option section.
- Option aliases use bpaf's chained-name support (`long("rcfile").long("init-file")`, `short('l').short('L')`): first spellings visible in help, rest hidden aliases.

### Behavior notes

Several builtins now behave closer to bash than before:
- `echo -- -n z` prints `-n z`; `echo ----------------` echoes the dashes
- `kill -s TERM $$`, `kill -9 $$`, `kill -L` work again under the harness
- `pwd -L -P` / `pwd -P -L`: last flag wins
- `set --` clears positional parameters; `test --` / `[ -- ]` are true
- `printf -- -5` and `command -- ls` honor the terminator

## Validation

- compat suite: **1854 passed, 0 failed**, 456 known-failures, 29 skipped — identical to the pre-port baseline
- all unit tests, completion tests (real `/usr/share/bash-completion`), integration tests pass
- `cargo fmt --check` and `cargo clippy --workspace --all-targets -D warnings` clean
- benchmarked against clap and the usage-based parser: see [benchmark comment](https://github.com/reubeno/brush/pull/1302#issuecomment-5397474989) and `benchmarks/three-way.py`

## Nice-to-have bpaf API changes

1. **Public access to rendered help text** — detailed help currently has to be obtained by running `run_inner(["--help"])` and unwrapping a `ParseFailure`. An `OptionParser::render_help_string()` (monochrome + colored variants) would remove the round-trip.
2. **Position-aware / sequential parsing primitive** — switches scan *all* unconsumed items regardless of position ("before" = declaration order). Bash-style "stop option parsing at the first operand" (`echo hi -n` must keep `-n` literal) cannot be expressed combinatorially today; we pre-split argv ourselves. Something like a positional parser with `allow_hyphen_values`-plus-`trailing_var_arg` semantics would let us delete the splitter.
3. **Per-parser "last occurrence wins" for repeated flags** — clap's `overrides_with` pattern (`cd -P -L`) can't be expressed; alternatives conflict and independent switches lose ordering.
4. **Man-page / markdown export** — clap ecosystem had `clap_mangen`/`clap_markdown`. An official (or documented third-party) way to render bpaf parsers to roff/markdown would restore richer doc generation; xtask currently emits from monochrome help text.
5. **`ParseFailure: std::error::Error`** (or an `into_error()`) so `?`-style plumbing works in test code without manual matching.

*(An earlier version of this list claimed aliases were unsupported; that was wrong — chained `.long()`/`.short()` provide hidden aliases, now used throughout.)*

## Follow-ups (not in this draft)

- Port `xtask`'s own CLI and any remaining dev-tool CLIs to bpaf
- Colored help rendering (bpaf's colored renderer is crate-private)
- Revisit `ContentOptions::colorized` once colorized rendering is reachable


lu-zero:
## Benchmark: bash vs {clap, bpaf, usage} × {fat-LTO, thin-LTO}

Interleaved, pinned sampling (`benchmarks/three-way.py`, 15 samples, median ± MAD·1.4826), byte-for-byte output parity verified across all shells before timing. All builds from identical release-profile flags; thin-LTO via `CARGO_PROFILE_RELEASE_LTO=thin`.

### Wall time (ms, median)

| Workload | bash | clap-fat | clap-thin | bpaf-fat | bpaf-thin | usage-fat | usage-thin |
|---|---|---|---|---|---|---|---|
| startup (-c exit) | 1.72 | 1.73 | 1.74 | **1.70** | 1.76 | 1.72 | 1.71 |
| interp-loop (60k) | 139.5 | 273.5 | 275.3 | 285.0 | 286.3 | **220.2** | 221.7 |
| wordops (string/array) | 150.6 | 54.4 | 55.2 | 53.1 | 53.8 | **52.9** | 53.5 |
| config-lint-500 (getopts-dense) | 24.0 | 95.3 | 97.1 | 74.7 | 75.3 | **50.9** | 51.6 |
| deploy-sim | 15.9 | 21.2 | 21.9 | 21.0 | 21.1 | 19.8 | **19.6** |

### Takeaways

**Parser backends** (config-lint, the parse-densest workload):
- clap → bpaf: **bpaf is 1.28× faster** (95.3 → 74.7 ms)
- clap → usage: usage is 1.87× faster (95.3 → 50.9 ms)
- bpaf → usage: usage is 1.47× faster

The ranking (usage < bpaf < clap) is consistent and well outside the ±0.5% noise floor.

**LTO: fat ≈ thin everywhere.** Within each backend the two variants sit within 1–3% of each other on every workload, with fat slightly ahead more often than not. No evidence that LTO flavor is a meaningful lever here — and notably, an earlier apparent "+35% pure-interp regression" for bpaf did **not** reproduce under this harness (interp-loop: clap-fat 273 ms vs bpaf-fat 285 ms, ~+4%), indicating it was a measurement artifact of sequential test-vs-reference sampling on a multi-tenant machine rather than real codegen cost.

**Other observations:**
- `wordops`: all brush builds are ~2.8× faster than bash
- `interp-loop`: the usage fork is also faster outside of parsing (220 vs 273–286 ms) — worth investigating what else differs there
- Startup: all builds within ±3% of bash's process spawn floor
- Peak RSS: all brush builds within ~0.5 MiB of each other across workloads

Reproduce with:
```
python3 benchmarks/three-way.py -s bash=/usr/bin/bash \
  -s clap-fat=... -s bpaf-fat=... -s usage-fat=... [--core N] [--samples 15]
```

pacak:
> Nice-to-have bpaf API changes

>  Public access to rendered help text — detailed help currently has to be obtained by running run_inner(["--help"]) and unwrapping a ParseFailure. A OptionParser::render_help_string() (monochrome + colored variants) would remove the round-trip.

Problem is `OptionParser` can contain many nested subparsers, say commands "foo" and "bar" and you can't really access it without running the parser. A helper like that can exist in theory, but to focus on a specific command it will be running the parser anyway.

0.10 should expose a bit more primitives so you can implement some bits, but even then dealing with global parsers will require building the whole context.

>  Position-aware / sequential parsing primitive — switches scan all unconsumed items regardless of position ("before" = declaration order). Bash-style "stop option parsing at the first operand" (echo hi -n must keep -n literal) cannot be expressed combinatorially today; we pre-split argv ourselves. Something like a positional parser with allow_hyphen_values-plus-trailing_var_arg semantics would let us delete the splitter.

0.10 adds `leftovers()`. If user passes a bunch of known named items followed by a positional item (and parser doesn't parse positionals) - `leftovers()` will collect that positional and everything else up to the end.


> Per-parser "last occurrence wins" for repeated flags — clap's overrides_with pattern > (pwd -P -L, cd -L -P) can't be expressed; alternatives conflict and independent switches lose ordering. A .conflicts_with_self(LastWins) style modifier would help.

Hmm... Isn't it what `.last()` does?

https://docs.rs/bpaf/latest/bpaf/trait.Parser.html#method.last

> NamedArg aliases & hide — .alias("log-enable") and calling .hide() on an unconsumed NamedArg aren't available; aliases currently require duplicating parsers inside a construct![a, b] alternative and hiding after conversion.

I don't think I understand this :)


> Custom metavar on named arguments — metavar() exists on any-based parsers but not on NamedArg before .argument(); being able to set metavar/help/hide uniformly on both classes would reduce helper shims.

metavar is needed to indicate that something is an argument (an option-argument if you use Open group terminology). I'm not sure how ability to change that. Do you have any examples where this can be used?

> Man-page / markdown export — clap ecosystem had clap_mangen/clap_markdown. An official (or documented third-party) way to render bpaf parsers to roff/markdown would restore richer doc generation; xtask currently emits from monochrome help text.

Hmm... https://docs.rs/bpaf/latest/bpaf/doc/index.html, requires docgen feature or something like that. Changed a bit in 0.10, but it should still be able to produce roff/markdown.

> ParseFailure: std::error::Error (or an into_error()) so ?-style plumbing works in test code without manual matching.

Hmmm... Yeah, I can add that.

lu-zero:
> NamedArg aliases & hide — .alias("log-enable") and calling .hide() on an unconsumed NamedArg aren't available; aliases currently require duplicating parsers inside a construct![a, b] alternative and hiding after conversion.

clap allows to have aliases, and bpaf apparently not in a as straightforward way, but I can be possibly wrong.

pacak:
> clap allows to have aliases, and bpaf apparently not in a as straightforward way, but I can be possibly wrong.

Each named item can have two visible names (first short and first long) and as many hidden aliases as you want. `short('n').short('N').long("name").long("also-name")`. Getting them to show in help can be more tricky (but much easier in 0.10), I'm not offering an official way because it gets too wide and too ugly too fast.

pacak:
https://docs.rs/bpaf/latest/bpaf/fn.long.html

lu-zero:
Updated, thank you :)

PR #1306, opened 2026-08-26
feat(builtins): implement declare/local -I; honor += in declarations

lu-zero:
Implements `declare`/`local -I` (bash 5.0+): a local declaration inherits value and attributes from the nearest same-name variable in an enclosing scope instead of starting unset. Combined with `+=`, the assignment appends to the inherited value. With no same-name variable anywhere it falls through to ordinary creation.

While there: the parser already records `+=` on assignments, but `declare`/`local` ignored it and *replaced* — declarations now honor the append flag (the existing "Appending via declare" oracle case is un-marked as known_failure).

## Motivation

Real-world ebuild shell scripts use both:

- `go-env.eclass`: `local -I $(all-flag-vars)`
- `cargo.eclass`: `local -Ix CARGO_TARGET_*_RUSTFLAGS+=" ..."`

Gentoo ebuilds hitting these failed outright at declaration time (`local: not yet implemented: declare -I`), which is what surfaced this.

## Testing

- 4 new oracle cases in `variables.yaml` (inherit, append, combined `-Ix`, no-inherit fallthrough), verified byte-for-byte against real bash
- "Appending via declare" oracle case now passes
- Full suite green apart from one pre-existing flake (`run_suspend_and_fg`, fails on clean `main` here too)

Draft for review — happy to split the `+=` part into its own PR if preferred.

### DIFF
--- a/brush-builtins/src/declare.rs
+++ b/brush-builtins/src/declare.rs
@@ -5,7 +5,6 @@ use std::{io::Write, sync::LazyLock};
 use brush_core::{
     ErrorKind, ExecutionResult, builtins,
     env::{self, EnvironmentLookup, EnvironmentScope},
-    error,
     parser::ast,
     variables::{
         self, ArrayLiteral, ShellValue, ShellValueLiteral, ShellValueUnsetType, ShellVariable,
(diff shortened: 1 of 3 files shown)


reubeno (approved):
Great fix -- thanks also for fixing the missing `+=` handling!

Issue #1318, opened 2026-08-30
8s hang on pty


Shiftwell-Jon:
### brush version

brush 0.4.0 (cargo:0.4.0)

### bash version (for comparison)

GNU bash, version 3.2.57(1)-release (arm64-apple-darwin24)

### Platform / OS

macOS

### Steps to reproduce

Ran terminal command:  sw_vers; echo "---"; uname -a; echo "---"; brush --version; echo "---"; brew info brush 2>&1 | head -10; echo "---"; echo $TERM_PROGRAM; echo $TERM

## Bug Report: `brush` hangs indefinitely in any interactive session attached to a real controlling terminal (pty)

**Project:** brush-shell (`brush`)
**Version:** 0.4.0 (`brush 0.4.0 (cargo:0.4.0)`), installed via Homebrew (`homebrew-core`, bottled)

### Environment
- OS: macOS 15.8 (Build 24H20)
- Kernel: Darwin 24.6.0, arm64 (Apple Silicon, T8103)
- Install path: `brush` (Homebrew `brush` formula)
- Reproduced both as the account's login shell (`dscl` `UserShell`) and by invoking the binary directly

### Summary
`brush` hangs forever (never prints a prompt, never returns) whenever it is started as an interactive shell attached to a **real controlling terminal (pty)** — regardless of whether it's invoked as a login shell (`-l`), a plain interactive shell (`-i`), or with no flags at all. It only works correctly when stdin is a pipe (non-tty), which masked the bug during quick manual testing (`echo cmd | brush` works fine).

This makes `brush` completely unusable as an interactive/login shell today: setting it as the macOS login shell (`chsh -s /opt/homebrew/bin/brush`) causes every new Terminal window/tab to hang on open.

### Reproduction
Using macOS `script` to allocate a real pty (a plain pipe does *not* reproduce the bug):

```bash
# Hangs (killed by timeout, exit 124):
script -q /dev/null bash -c 'timeout 8 brush -l'      # login shell
script -q /dev/null bash -c 'timeout 8 brush -i'      # interactive shell
script -q /dev/null bash -c 'timeout 8 brush'         # plain invocation

# Works fine (no real tty attached):
( echo 'echo hi'; sleep 1 ) | timeout 5 brush -i      # prints "hi", exits 0
( echo 'echo hi'; sleep 1 ) | timeout 5 brush -l      # prints "hi", exits 0
brush -c 'echo hello'                                  # prints "hello", exits 0
```

[...]

PR #1322, opened 2026-08-31
test(shell): add pty startup tests over the default input backend

reubeno:
## Summary

Adds `brush-shell/tests/pty_startup_tests.rs` (gated to the same expectrl-supported unix targets as `interactive_tests.rs`): tests that verify the shell becomes responsive — prints a prompt and executes a first command — within a bounded timeout when attached to a real pty, across the invocation modes real terminal environments use:

- plain invocation, explicit `-i`, and explicit `-l` (the `--login` flag path, distinct from the argv0-based login detection already covered)
- a login shell with an empty `$HOME` and no `--norc`/`--noprofile` (startup must not depend on dotfiles existing)
- the shell spawned through an intermediate `sh -c` inside the pty, so it is not the pty session leader and must take terminal ownership from another process group (as under `script(1)`, `su`, multiplexers)

Unlike `interactive_tests.rs`, which exercises interactive behavior (job control, pipelines) via the simplified `basic` input backend, these tests leave the input backend at its **default** and focus on startup, where terminal-ownership setup (process groups, `tcsetpgrp`, `SIGTTIN`/`SIGTTOU`) happens. All tests enforce a bounded 8s expect timeout so a hang fails fast with a clear message, and the harness answers cursor-position (DSR) queries as a real terminal emulator would (the default backend requires a response at startup).

The test file is deliberately not declared as a `brush-*-tests` target so it stays out of the prebuilt test-binary artifact consumed by the os-tests CI jobs, which cannot run pty-based tests (documented in `Cargo.toml`).

## Background and findings

Motivated by #1318 ("8s hang on pty", reported against the 0.4.0 release on macOS):

- **The hang does not reproduce on current `main`** — all 5 tests pass on Linux and macOS CI.
- **Root cause identified**: #1318 matches #1137, fixed by #1138 (`TerminalControl` guard previously dropped immediately, handing the foreground back to the parent's process group before reading from the tty → SIGTTIN stop). #1138 merged three days *after* the `v0.4.0` tag, so the reporter's build predates the fix.
- **Verified empirically**: re-simulating the pre-#1138 guard drop on `main` makes `login_shell_as_grandchild_starts_promptly` hang for the full timeout — the non-session-leader topology is the only one that exercises the ownership transition, which is why direct-spawn pty tests never caught it.

## Test plan

- `cargo nextest run --test pty_startup_tests` — 5/5 pass on Linux; macOS CI confirmed running all 5 (test count 2750→2755, binaries 17→18 vs. `main`).
- `cargo clippy` / `cargo fmt` clean.

https://claude.ai/code/session_01AgPLaFEeNqLCEyQm21XYcj

### DIFF
--- a/brush-shell/Cargo.toml
+++ b/brush-shell/Cargo.toml
@@ -44,6 +44,12 @@ path = "tests/interactive_tests.rs"
 name = "brush-completion-tests"
 path = "tests/completion_tests.rs"
 
+# N.B. tests/pty_startup_tests.rs is deliberately left auto-discovered (no
+# `brush-*-tests` name): the CI `brush_*_tests-*` artifact glob would otherwise
+# ship it to the os-tests jobs, which run prebuilt test binaries outside a cargo
+# target dir and so can't run pty-based tests (see the interactive/version test
+# skips there).
+
 [[bench]]
 name = "shell"
 path = "benches/shell.rs"
--- a/brush-shell/tests/pty_startup_tests.rs
+++ b/brush-shell/tests/pty_startup_tests.rs
@@ -1,0 +1,192 @@
+//! Shell startup tests run over a real pseudo-terminal (pty).
+//!
+//! These tests verify that the shell becomes responsive — prints a prompt and executes
+//! a first command — within a bounded amount of time when attached to a controlling
+//! terminal, across the invocation modes real terminal environments use: plain
+//! invocation, explicit `-i`, explicit `-l`, presence/absence of rc and profile files,
+//! and different process topologies (the shell as the direct pty-attached child vs. a
+//! descendant of another pty-attached process).
+//!
+//! They complement `interactive_tests.rs`, which exercises interactive *behavior* (job
+//! control, suspension, pipelines) using the simplified `basic` input backend and
+//! argv0-based login detection. This file instead focuses on *startup*, uses the
+//! default input backend (the one real terminal sessions get), and exercises the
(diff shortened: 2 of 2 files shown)


reubeno:
**CI findings (run 33427419068):**

✅ **Confirmed the macOS job ran the new tests, and they all passed.** nextest doesn't print passing test names at the default status level, but the counts are conclusive: Test (macOS) went from **2750 tests / 17 binaries** on the last `main` run to **2755 tests / 18 binaries** on this PR — exactly the 5 new tests and 1 new test binary, all passing (Linux shows the same +5/+1: 2769/17 → 2774/18). So the #1318 hang does **not** reproduce on macOS CI runners under expectrl's pty either; a repro may require macOS Terminal/`script`-specific pty semantics not emulated here.

Two notes:
- The `OS target tests (openSUSE Tumbleweed/latest)` failure is unrelated: a compat-suite discrepancy in `mapfile -u invalid fd` vs. Tumbleweed's bash 5.3.15 oracle (same case passed on `main` ~2.5h earlier with the same bash version, so it looks flaky). The new test binary isn't even part of that job — see next point.
- The new test file is deliberately *not* declared as a `brush-*-tests` `[[test]]` target: the CI `brush_*_tests-*` artifact glob would otherwise ship it to the os-tests container jobs, which run prebuilt test binaries outside a cargo target dir and can't run pty-based tests (the same reason `*interactive*`/`*version*` binaries are skipped there). Added a Cargo.toml comment documenting this.

PR #1330, opened 2026-09-02
fix(interactive): handle Signal::HostCommand so bind -x keys don't kill the shell

### LINKED ISSUE #1328: bind -x keys terminate the shell on main (reedline 0.49 returns Signal::HostCommand)
## Summary
On current `main`, every key bound with `bind -x` terminates the interactive shell the moment it is pressed:

```
error: unexpected error occurred reading input
```

Not in any tagged release yet — 0.4.0 (reedline 0.47) is fine — but `main` has carried it since #1240 (2026-07-18), so it hits nightly / from-source users today and will ship in 0.5.0 unless fixed first. In practice it breaks atuin's `Ctrl-R` and anything else built on `bind -x`.

## Environment
- brush `main` @ 0a04d23d (also reproduced at 30a3bce), reedline 0.49.0
- Linux x86_64 and aarch64 (Void Linux, glibc), `--input-backend=reedline` (the default)
- Not reproducible with the 0.4.0 release binary (git 96a26d0, reedline 0.47.0)

## Steps to reproduce
```
$ brush --norc --noprofile -i
brush-0.5$ bind -x '"\C-t": echo hello'
brush-0.5$ <press Ctrl-T>
error: unexpected error occurred reading input
$
```
No atuin or other external program needed; the bound command can be anything. Reproduced by hand in a terminal, and under a scripted pty harness (with a responder for the cursor-position queries so the failure can't be mistaken for #1329).

## Root cause
reubeno/brush wires `bind -x` through `ReedlineEvent::ExecuteHostCommand` and recognises the result in `ReedlineInputBackend::read_line` via `Signal::Success(s)` plus its `# bind-command` marker (`edit_mode::is_reedline_host_command`).

reedline 0.48 changed `ExecuteHostCommand` to return a dedicated `Signal::HostCommand(String)` instead of `Signal::Success` (nushell/reedline#1049, merged 2026-04-17). `read_line` has no arm for `HostCommand`, so it hits `Ok(_) => Err(ShellError::UnexpectedInputFailure)` and the interactive loop exits.

The bump to reedline 0.49 landed in #1240; this has been the behavior of `main` since.

## Fix
One additional match arm: `Ok(reedline::Signal::HostCommand(s)) => Ok(ReadResult::BoundCommand(s))`. Everything downstream (`BoundCommand` execution, `READLINE_LINE`/`READLINE_POINT`) already works — with the arm in place atuin's full `Ctrl-R` → search → accept → splice → run flow works on both architectures. PR incoming with a pty regression test.

Blfrg:
Fixes #1328.

reedline 0.48 (nushell/reedline#1049) made `ExecuteHostCommand` return `Signal::HostCommand` instead of smuggling the command through `Signal::Success`. `ReedlineInputBackend::read_line` only recognised the `Success` + `# bind-command` marker form, so since the reedline 0.49 bump in #1240 every `bind -x` key press fell into the `Ok(_)` catch-all and exited the shell. Most visible symptom: atuin's `Ctrl-R` kills brush on any build from `main`.

This adds the `HostCommand` arm, mapping it to `ReadResult::BoundCommand` exactly as the marker path does. The marker path stays so the `edit_mode` format/parse round-trip is untouched.

**Test:** new `brush-shell/tests/reedline_interactive_tests.rs` drives the reedline backend over a pty, answering reedline's cursor-position queries itself (same approach as `pty_startup_tests.rs`; left auto-discovered for the same CI-artifact reason, note in `Cargo.toml` extended). It binds `Ctrl-T` with `bind -x`, presses it, and requires the bound command's output and a still-interactive shell.

- on `main`: fails with EOF (shell exited)
- with this change: passes

Also verified by hand on x86_64 and aarch64: atuin `Ctrl-R` → search → accept → `READLINE_LINE` splice → re-run works end to end.

### DIFF
--- a/brush-interactive/src/reedline/input_backend.rs
+++ b/brush-interactive/src/reedline/input_backend.rs
@@ -157,12 +157,21 @@ impl InputBackend for ReedlineInputBackend {
         if let Some(reedline) = &mut self.reedline {
             match reedline.read_line(&prompt) {
                 Ok(reedline::Signal::Success(s)) => {
+                    // reedline < 0.48 delivered `ExecuteHostCommand` payloads through
+                    // `Success` (hence the marker check); newer versions use the
+                    // dedicated `HostCommand` signal below. Keep both so the marker
+                    // round-trip stays correct regardless of which one we get.
                     if edit_mode::is_reedline_host_command(s.as_str()) {
                         Ok(ReadResult::BoundCommand(s))
                     } else {
                         Ok(ReadResult::Input(s))
                     }
                 }
+                // Since reedline 0.48 (nushell/reedline#1049), a key bound with
+                // `bind -x` comes back as `Signal::HostCommand` rather than
+                // `Signal::Success`. Without this arm it fell into the catch-all
+                // below and every bound key press terminated the interactive shell.
+                Ok(reedline::Signal::HostCommand(s)) => Ok(ReadResult::BoundCommand(s)),
                 Ok(reedline::Signal::CtrlC) => Ok(ReadResult::Interrupted),
                 Ok(reedline::Signal::CtrlD) => Ok(ReadResult::Eof),
                 Ok(reedline::Signal::ExternalBreak(_)) => Err(ShellError::UnexpectedInputFailure),
--- a/brush-shell/Cargo.toml
+++ b/brush-shell/Cargo.toml
@@ -44,11 +44,11 @@ path = "tests/interactive_tests.rs"
 name = "brush-completion-tests"
 path = "tests/completion_tests.rs"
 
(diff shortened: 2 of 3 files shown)


reubeno:
Thanks for the contribution, @Blfrg! I actually was working on fixing the same issue in parallel. (In my branch, I'm preferring ripping out the old workaround present, particularly given that we require a newer reedline in our Cargo.toml and it should be guaranteed to have the upstream host command support.)

I'll take a look at your changes more closely and figure out how best to proceed.

reubeno (approved):
I've looked through the code -- and I'd like to get it merged sooner than later. I'll likely follow up subsequently with my already-drafted, more aggressive cleanup of the old workaround for host commands.

PR #1350, opened 2026-09-06
wip: abstract builtin arg parsing; clap/bpaf/usage backends

lu-zero:
## Summary

Redo of the `wip/bpaf-engine` experiment on top of current `origin/main`, importable in reviewable pieces:

- Engine-neutral contract: `brush_core::args::{FromArgs, ArgsError}`; `builtins::Command: FromArgs` with a transitional clap blanket impl and `clap_content()` help helper.
- Per-builtin split: logic stays in `{builtin}.rs`, engine structs live in `{builtin}/{clap,bpaf,usage}.rs`, selected by `arg_impl!` with global priority usage > bpaf > clap (per-builtin mixing explicitly deferred).
- Shared runtimes: `args/bpaf_support.rs`, `args/usage_support.rs` (plus-cluster expansion, option/trailing split, `--` handling).
- Benchmark harness ported from `bpaf-port` (`benchmarks/three-way.py` + `benchmarks/real-world/*`, plus `--numa-node` pinning) with recorded results.

## Verification (all pinned to NUMA node 3 / cpu 96)

- Compat suite, all three backends: **1971 passed / 0 failed** (471 known-fail, 29 skipped).
- Abstraction zero-cost check (interleaved anchor-vs-abstract): pairwise medians 0.99-1.03x, RSS within 128 KiB.
- Three-backend shootout vs bash oracle, byte-identical outputs: `benchmarks/results/backends-c986f537/` (parse-dense `config-lint-500`: usage 62ms, bpaf 79ms, clap 107ms).
- `cargo fmt --check` and `cargo clippy` clean for all three engine configs.

## Notes for reviewers

- Conflict resolutions vs the original wip stack all favor current main: shared `lookup` framework (`command`, `type`), `declare -I`, `read` UTF-8 rework, `printf %q` rework.
- One genuine port gap fixed along the way: `command -v/-V` last-wins in both non-clap backends.
- `interp-loop` timing anomaly (usage inexplicably faster on a parse-free workload) needs a re-run before quoting.

Assisted-by: Muse Spark:muse-spark-1.3-contributor-free [opencode]


lu-zero:
## Azure Linux CI failures: environmental, not caused by this branch

The remaining failure is `OS target tests (Azure Linux/4.0)` with 2 failed cases. I investigated; both are environment-sensitive and outside this branch's semantic surface:

**1. `mapfile -u invalid fd` (oracle `exit: 0` vs brush `exit: 1`)**

The oracle is stock container bash — our code cannot influence its exit status. I reproduced the exact signature locally by holding fd 99 open in the parent:

```sh
exec 99>/tmp/probe
bash -c 'mapfile -u 99 arr; echo "oracle exit: $?"'   # oracle exit: 0
brush -c 'mapfile -u 99 arr; echo "brush exit: $?"'   # brush exit: 1
```

So on Azure the harness happened to hold fd 99 open (non-CLOEXEC, likely a leaked PTY master/pipe from a concurrent test) when spawning the oracle: bash inherits it (`exit: 0`), brush doesn't (`exit: 1`). Whether fd 99 is occupied at spawn time is timing-dependent — flaky by nature. This branch changes no fd/startup code (`brush-shell` delta is feature forwarding + one `get_content` impl), so `main` fails identically in that environment.

**2. `shopt -o interactive monitor default` (empty brush output)**

Passes locally, including 10/10 runs under a real PTY via `script(1)` (both shells print `monitor on`) and through the compat harness. The `shopt` logic is untouched except for the mechanical clap-module split (struct identical). Empty-stdout-with-exit-0 smells like a PTY/job-control acquisition race in that container, and these tests are timing-sensitive to the binary under test.

**Ask**: I don't have admin rights on this repo so I can't re-run the job — could someone please re-run the Azure Linux OS-target job? Both cases pass reliably outside that container. If it fails identically a second time I'll dig deeper.

Assisted-by: Muse Spark:muse-spark-1.3-contributor-free [opencode]

PR #1361, opened 2026-09-09
perf(*): extend over for loop

Elsie19:
A couple for loops can easily be condensed to `.extend`s, and a couple can potentially trigger resizes when they aren't needed, so I fixed those.

These are probably micro-optimizations so feel free to close if you don't want them; I just think they're more idiomatic.


reubeno (approved):

reubeno:
It's not clear if this wins us meaningful runtime performance, but I think the code is better with the change than without it. Happy to merge it.

Issue #1206, opened 2026-06-17
brush doesn't seem to implement kill with a numeric signal argument


jmquigs:
### brush version

brush 0.4.0 (git:d42bdd2)

### bash version (for comparison)

GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)

### Platform / OS

Linux

### Steps to reproduce

Under brush just run kill with any numeric argument for the signal on a pid you own, it will display an error.

```
~$ bash
~$ ps
    PID TTY          TIME CMD
  62862 pts/5    00:00:00 bash
  62889 pts/5    00:00:00 bash
  62901 pts/5    00:00:00 ps
~$ kill -9 62889
Killed
~$ brush
~$ ps
    PID TTY          TIME CMD
  62862 pts/5    00:00:00 bash
  62922 pts/5    00:00:00 brush
  62955 pts/5    00:00:00 ps
~$ kill -9 62922
kill: invalid signal name
```

### Actual behavior (brush)

kill: invalid signal name

FWIW it also didn't kill the process as I expected it would

### Expected behavior

Should translate the numeric argument name into the proper signal and send it

jmquigs:
Thanks for fixing this.  Note there is still an issue with kill -0 which is not parsed successfully:

```
~$ bash
~$ ps
    PID TTY          TIME CMD
  40467 pts/3    00:00:00 bash
 154866 pts/3    00:00:00 bash
 154878 pts/3    00:00:00 ps
~$ kill -0 40467
~$ echo $?
0
~$ kill -0 1234567
bash: kill: (1234567) - No such process
~$ echo $?
1
~$ brush
~$ ps
    PID TTY          TIME CMD
  40467 pts/3    00:00:00 bash
 154866 pts/3    00:00:00 bash
 154929 pts/3    00:00:00 brush
 154962 pts/3    00:00:00 ps
~$ kill -0 40467
error: kill: EXIT: invalid signal specification
~$ echo $?
1
~$ kill -0 1234567
error: kill: EXIT: invalid signal specification
~$ echo $?
1
```

kill -0 acts like an existing check to see if the pid actually exists.  It seems trivial, but LLMs do generate bash code that uses it.
One spec describing the behavior is here (https://pubs.opengroup.org/onlinepubs/9699919799/functions/kill.html).  That is the kill system call, but I believe the internal bash implementation as well as /usr/bin/kill are both expected to handle that case.

reubeno:
[...]

PR #1367, opened 2026-09-10
Implement kill -0

pepa65:
Implement the `kill -0 PID` (and `kill -n 0 PID`) command, to mirror bash's behavior.
This would close Issue #1206.
Assisted-by: ChatGPT


brush-builtins/src/kill.rs
```diff
@@ -74,7 +80,9 @@ impl builtins::Command for KillCommand {
             if let Some(possible_sigspec) = arg.strip_prefix("-") {
                 // See if this is -sigspec syntax. The sigspec may be a signal name
                 // (e.g., -TERM) or a signal number (e.g., -9).
-                if let Ok(parsed_trap_signal) = possible_sigspec.parse::<TrapSignal>() {
+                if possible_sigspec == "0" {
```
reubeno: issue(blocking): We probably shouldn't just be doing a direct string comparison against `"0"`. From some quick testing, it looks like any string that would parse to the integer `0` should fall into this case, e.g.:

> Image

brush-core/src/sys/unix/signal.rs
```diff
@@ -35,6 +35,19 @@ pub fn kill_process(
...
 
+/// Checks whether a specific process exists and can be signaled.
+///
+/// This performs the `kill(2)` signal-zero check without actually sending
+/// a signal.
+pub fn check_process(
+    pid: sys::process::ProcessId,
+) -> Result<(), error::Error> {
+    nix::sys::signal::kill(nix::unistd::Pid::from_raw(pid), None)
+        .map_err(|_errno| error::ErrorKind::FailedToSendSignal)?;
```
reubeno: question(blocking): Since `kill -0` is typically used to check for the existence of a process, can we differentiate the return values and faithfully bubble them up? In other words, if the underlying call fails because the process couldn't be found -- ideally we'd be able to report via the `kill` builtin that the process doesn't exist. This is in contrast with, for example, a permissions issue.

reubeno (changes requested):
Thanks for the contribution, @pepa65! This looks like a great start.

I've left a couple of comments in addition to the ones that copilot left (which I'd agree with). In particular, it would be ideal to add some YAML-based test cases to our compatibility test suite; that's the best way to demonstrate that the functionality as implemented now matches `bash`.

There should be plenty of examples of tests like that, and tests for `kill` in particular. Holler if you could use some pointers.

brush-core/src/sys/unix/signal.rs
pepa65: To do the same as bash and as kill in procps, where only 0 (signalable) and 1 (insufficient permissions or non-existing process) are returned, brush should do the same in order to not break compatibility for existing scripts.

brush-builtins/src/kill.rs
pepa65: Agreed.

pepa65:
What do I do with such a test result:
```
* Test case: [kill -0 (numeric, non-existent)]... 
    Oracle comparison:
      status matches (exit status: 1) ✔️
      stdout matches ✔️
      stderr DIFFERS:
          ------ Oracle <> Test: stderr ---------------------------------
          - /usr/bin/bash: line 1: kill: (99999999) - No such process
          + error: kill: failed to send signal to process
            
          ---------------------------------------------------------------
      temp dir matches ✔️
    FAILED.
```

Firstly, the output differs, but that seems to be common for brush vs. bash..?

But I guess here is differentiation, even if the return code is always 1 for non-signalable processes: the error text is different for the non-existent and lacking-permissions cases. This is what should be implemented, with the error texts corresponding to the bash output in both cases.

brush-shell/tests/cases/compat/builtins/kill.yaml
```diff
@@ -22,6 +22,10 @@ cases:
     stdin: |
       kill -n 9 $$
 
+  - name: "kill -n 0"
+    stdin: |
+      kill -n $((5-5)) $$
```
reubeno, line 27: suggestion: It's probably worth adding a line of execution after the `kill` invocation -- maybe an echo? -- just to make sure that the `kill` doesn't actually terminate the shell. That would be a useful additional validation.

brush-shell/tests/cases/compat/builtins/kill.yaml
```diff
@@ -30,6 +34,14 @@ cases:
     stdin: |
       kill -9 $$
 
+  - name: "kill -0 (numeric)"
+    stdin: |
+      kill -0000 $$
```
reubeno, line 40: (Same as above)

brush-shell/tests/cases/compat/builtins/kill.yaml
```diff
@@ -30,6 +34,14 @@ cases:
...
+
+  - name: "kill -0 (numeric, non-existent)"
+    stdin: |
+      kill -0 99999999
```
reubeno, line 46: (Same as above)

brush-builtins/src/kill.rs
```diff
@@ -53,37 +55,46 @@ impl builtins::Command for KillCommand {
...
-                    trap_signal = parsed_trap_signal;
-                } else {
-                    writeln!(
-                        context.stderr(),
-                        "{}: {}: invalid signal specification",
-                        context.command_name,
-                        possible_sigspec
-                    )?;
-                    return Ok(ExecutionResult::general_error());
+                if let Ok(signal_number) = possible_sigspec.parse::<i32>() {
```
reubeno: suggestion:  This can be slightly simplified. Instead of the `if let Ok(...) ...` followed by the comparison against 0, we can instead do something like:

```rust
if let Ok(0) = possible_sigspec.parse::<i32>() {
    signal_zero = true;
} else if let Ok(parsed_trap_signal) = possible_sigspec.parse::<TrapSignal>() {
    ...
} else {
    ...
}
```

reubeno (changes requested):
Thanks for iterating on the changes and for adding the tests; it's looking in pretty good shape.

I've left a few additional comments/requests -- much smaller this time.

brush-shell/tests/cases/compat/builtins/kill.yaml
pepa65: To put an echo after the `kill -0 99999999` does not help, but I did it after all the other ones.

pepa65:
Thank you for your encouraging and helpful comments. I think brush is great, I just bumped into it last week, and it's basically ready to replace bash in most cases already..! (I am building it as a musl static binary, and that works great, and isn't too big.)

pepa65:
Sorry for wasting everybody's time with the untested push...

pepa65:
Not sure what to do about this (below). The oracle included the pid that doesn't exist in the error, should I mirror this? If you do a normal `kill 99999999` the error is "failed to send signal to process", but if you do `kill -0 99999999` the error is "no such process".

```
* Test case: [kill -0 (numeric, non-existent)]...                                                                              
    Oracle comparison:                                                                                                         
      status matches (exit status: 1) ✔️                                                                                        
      stdout matches ✔️                                                                                                         
      stderr DIFFERS:                                                                                                          
          ------ Oracle <> Test: stderr ---------------------------------                                                      
          - /usr/bin/bash: line 1: kill: (99999999) - No such process                                                          
          + error: kill: no such process                                                                                       
                                                                                                                               
          ---------------------------------------------------------------                                                      
      temp dir matches ✔️                                                                                                       
    FAILED.   
```

reubeno:
> Not sure what to do about this (below). The oracle included the pid that doesn't exist in the error, should I mirror this? If you do a normal kill 99999999 the error is "failed to send signal to process", but if you do kill -0 99999999 the error is "no such process".

@pepa65 -- good question. We probably could be doing something better about this in the project, but for now, we enable the `ignore_stderr` option on these kinds of negative tests where the error messages diverge. If you search nearby you should see some examples of that.

With that open set (just on the test that fails in that way), then exit code, stdout, file output, etc. are all verified -- just `stderr` is ignored.

pepa65:
Okay, fixed that. On my test locally I also got:
```
* Test case: [Pathname expansion: extglob disabled]... 
    Oracle comparison:                                         
      status mismatch: exit status: 2 from oracle vs. exit status: 126 from test                                               
      stdout matches ✔️                                                                                                        
      stderr ignored                                                                                                           
      temp dir matches ✔️                                      
    FAILED.       
```

(But I didn't 'fix' that.)

reubeno:
> Okay, fixed that. On my test locally I also got:
> 
> ```
> * Test case: [Pathname expansion: extglob disabled]... 
>     Oracle comparison:                                         
>       status mismatch: exit status: 2 from oracle vs. exit status: 126 from test                                               
>       stdout matches ✔️                                                                                                        
>       stderr ignored                                                                                                           
>       temp dir matches ✔️                                      
>     FAILED.       
> ```
> 
> (But I didn't 'fix' that.)

That does sound unrelated. Would you mind sharing more info about which OS you're running on, which version of bash you have installed, etc.?

(We won't hold up merging this PR because of that, though.)

reubeno (approved):

reubeno:
Merged! Thanks for the contribution, @pepa65!

PR #1390, opened 2026-09-20
refactor(builtins)!: engine-neutral builtin contracts; move clap out of brush-core

reubeno:
## Summary

Removes the argument-parsing engine from `brush-core`'s builtin contract. A builtin now implements `FromArgs`, `HelpContent`, and `Command::execute`; the shell passes arguments through exactly as expanded and does no option handling on a builtin's behalf. clap moves to a new `brush-builtin-utils` crate, whose `clap_builtin!` macro implements both traits from an existing derive, and a `verbatim_builtin!` adapter covers builtins that parse nothing.

## Breaking changes worth discussing

- `Command: FromArgs + HelpContent`; `new`, `takes_plus_options`, and `get_content` are gone.
- `SimpleCommand`, `DeclarationCommand`, and the `simple_builtin`/`decl_builtin`/`raw_arg_builtin` constructors are gone. Everything registers with `builtin::<T, _>()`; declaration handling is `FromArgs::TAKES_DECLARATIONS` on the type, so registration can no longer disagree with the implementation.
- `Registration` fields are private. `ContentType` is replaced by `synopsis`, `description`, and `detailed_help` methods.
- **Silent:** execute functions no longer receive the invoked name as `args[0]`. Bare-function builtins must use `context.command_name` and stop skipping the first argument.
- `parse_known` / `try_parse_known` leave core.
- `CompleteAction` / `CompleteOption` drop `clap::ValueEnum` for strum `Display` / `FromStr` / `VariantNames`, and become `Copy`.

Full list with detection greps and replacements: `brush-core/skills/migrate-0.5-to-0.6/SKILL.md`; walkthrough: `docs/migrations/brush-core/0.5-to-0.6.md`.

## Behavior changes

- `declare`, `local`, `readonly`, `export`, `typeset` stop reading options at the first operand, as bash does.
- `export` rejects operands that aren't valid identifiers (status 1); also fixes `export 'ea[0]=1'`.
- `help -s` / `help -d` print one framed line per builtin.

## Other changes

- `ShellBuilder::command::<T>(name)` shorthand; `SE` inferred at registration sites.
- Macro misuse is a compile error naming the accepted forms, pinned by trybuild goldens.
- `brush-core` gets a hand-written `custom-builtin` example with no parser; the clap one becomes `clap-builtin` in `brush-builtin-utils`.
- Establishes the migration-guide pattern: an agent guide shipped inside the crate (Symposium picks it up via `depends-on`) plus a human walkthrough, and a `document-breaking-changes` skill for future breaking changes.
- `+x` option handling is now inferred from clap metadata instead of hard-coded per builtin; `set` loses ~50 lines of bespoke parsing.

## Testing

`cargo xtask ci full` passes: unit, hooks, schemas, and the compat suite (3063 tests). New compat cases cover the `export` and `help -s` changes; an engine-neutral integration test in `brush-core` proves the contracts compile without clap.


reubeno:
@lu-zero FYI, I took your improved version of this work and layered on some additional changes based on my review. This is what resulted... much larger than I'd like but it's looking more like what I'd expect.

I'm going to review it again and try to get it merged before the weekend is over.

.claude/skills/document-breaking-changes
```diff
@@ -0,0 +1 @@
+../../.agents/skills/document-breaking-changes
```
reubeno: Why a symlink here? We should be clear on whether we're going with .agents or .claude, and which agents support which.

.agents/skills/document-breaking-changes/SKILL.md
```diff
@@ -0,0 +1,104 @@
+---
+name: document-breaking-changes
+description: How to document a breaking API or behaviour change to a published brush crate: the migration guide pair (crate-shipped agent guide plus human walkthrough), the commit footer, rustdoc notes, and tests. Use whenever a change removes, renames, or re-types public API in a library-tier crate, changes a convention that compiles unchanged, or changes observable shell behaviour that embedders or scripts could depend on.
```
reubeno: We should respell behaviour as behavior for consistency with the rest of the project.

.agents/skills/document-breaking-changes/SKILL.md
```diff
@@ -0,0 +1,104 @@
...
+
+- **MECHANICAL**: a rename, removal, or re-typing that the compiler catches and
+  that has a single replacement. Most changes.
+- **CONDITIONAL**: has a replacement that depends on how the caller used it.
+  Give the cases as a table.
+- **SILENT**: compiles unchanged but behaves differently. The most dangerous
+  kind: a changed argument convention, a changed default, a field that means
+  something else. Call these out separately and give a verification step a
+  reader can run.
+- **BEHAVIOUR**: no API change, but scripts or embedders can observe a
```
reubeno: Ditto. We should search and replace.

.config/nextest.toml
```diff
@@ -4,5 +4,12 @@ status-level = "slow"
 slow-timeout = { period = "5s", terminate-after = 3, grace-period = "2s" }
 fail-fast = false
 
+# trybuild compiles its compile-fail cases in a separate target directory, so a
```
reubeno, line 7: Is this something that can live closer to the definition of the tests in question?

brush-builtin-utils/src/args.rs
```diff
@@ -0,0 +1,90 @@
+//! Engine-neutral helpers for interpreting a builtin's arguments.
+//!
+//! The shell hands a builtin its arguments exactly as expanded. These helpers
+//! implement the conventions bash's own builtins follow, so any argument-parsing
+//! engine can apply them consistently.
+
+use brush_core::CommandArg;
+
+/// Flattens arguments into plain words. Assignments render as `name=value`.
+pub fn into_words(args: Vec<CommandArg>) -> Vec<String> {
```
reubeno: Does this really need to be pub, or just pub(crate)?

brush-builtin-utils/src/args.rs
```diff
@@ -0,0 +1,90 @@
...
+/// bash's declaration builtins (`declare`, `export`, `local`, ...) read them.
+///
+/// Options are the leading string arguments that begin with `-` or `+` and are
+/// longer than that one character. The first argument that is not an option
+/// ends them, so `declare a=1 -x` has no options and two operands. A `--` ends
+/// them too, and is dropped.
+///
+/// Assumes no option takes its value as a separate argument, which holds for
+/// every declaration builtin.
+pub fn split_leading_options(mut args: Vec<CommandArg>) -> (Vec<String>, Vec<CommandArg>) {
```
reubeno: Ditto.

brush-core/src/completion.rs
reubeno: Is there some cleaner way using strum or some nearby crate for having a single definition that gives us a flag set, without need for all the custom functions below?

brush-core/src/builtins.rs
```diff
@@ -1,467 +1,254 @@
...
+            "{name}: {}\n    {}\n",
+            Self::synopsis(name),
+            Self::description(name)
+        ))
+    }
 }
 
 /// Options for retrieving built-in command content.
 #[derive(Default)]
+#[non_exhaustive]
```
reubeno, line 115: Why is this non exhaustive? What do we have in mind?

lu-zero:
Stacked follow-up: #1404 builds on this PR to parse every builtin, and the `brush` command line, with usage-rs. It contains this PR's commits rebased onto current `main` (two small conflicts, with #1398's `-c` rework and in `docs/README.md`, described there), in case that rebase is useful here.

